PR #5, opened 2022-04-22
Add workflow for tox tests

plars:


plars:
Sorry for the weird merge here. I'm pretty sure that what happened is that when I pushed the previous PR to fix what was causing the workflow to fail to run, I think I branched from this rather than main. Since that one landed before this one, it created some confusion for git, especially after I added another commit on this to fix the pylint conflict with some sentry_sdk weirdness.
However, the workflow does run now, and passes everything, including pylint

kissiel:
Looks good, +1

PR #6, opened 2022-04-22
Move testfligner.py to a module to fix setup issues

plars:
This should fix the setup.py issues we see in the tests. Not sure why my previous checkout didn't expose this sooner, but now that I'm playing with a fresh checkout from github, it's failing for me locally too


mz2 (approved):
LGTM

PR #10, opened 2022-06-28
Add HACKING.md

plars:
This adds the instructions for setting up the multipass environment, along with a HACKING.md

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -29,10 +29,10 @@ Next run the following command (this will take a while):
     $ cat devel/testflinger.yaml |multipass launch --name testflinger -c4 -m8GB -d32GB --timeout 600 --cloud-init -
 ```
 
-After some time, it may still timeout due to the amount of time it takes to
-retrieve all the pieces and set them up.  If this happens, give it a few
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -0,0 +1,58 @@
...
+
+Next run the following command (this will take a while):
+```
+    $ cat devel/testflinger.yaml |multipass launch --name testflinger -c4 -m8GB -d32GB --timeout 600 --cloud-init -
+```
+
+After some time, it may still timeout due to the amount of time it takes to
+retrieve all the pieces and set them up.  If this happens, give it a few
+minutes before trying to run a test job, but you should still be able to
+connect.
```
jocave, line 35: Due to the complexity of the environment being setup, this command may timeout rather than complete. If this happens the environment should still reach the final state successfully, but you will need to wait a few minutes before trying to connect.

jocave (changes requested):
Found one of the paragraphs a little tricky to reason and made an alternative suggestion...

jocave (approved):
Nice, seems to make sense to me. Let's land it and we can get more feedback from people new to the project.

PR #11, opened 2022-06-30
Move api under testflinger dir and avoid circular dependencies another way

plars:


plars:
When writing the new charm for testflinger, I ran into a bit of trouble with this. Moving the api dir under the testflinger dir also created some issues with a potential for circular imports, but thanks to some flask magic, it looks like there's a good way to handle this with current_app

kissiel:
Overall, looks great! I think it makes things even simpler.
I'm not a contributor so I cannot use reaction. so +1
### +1 👍

codersquid (approved):
lgtm

PR #13, opened 2022-07-11
Add a default handler for 404

plars:


jocave (approved):
Today I read about werkzeug. Changes look good to me. +1

PR #14, opened 2022-07-12
Add a job cancellation functionality

nadzyah:


testflinger/api/v1.py
```diff
@@ -345,3 +376,37 @@ def job_position_get(job_id):
...
+            result_file, "r", encoding="utf-8", errors="ignore"
+        ) as results:
+            data = json.load(results)
+    except json.decoder.JSONDecodeError:
+        # If for any reason it's empty or has bad data
+        data = {"job_state": "bad_data"}
+    if data["job_state"] in ["complete", "cancelled"]:
+        return "The job is already completed or cancelled", 400
+    # Remove job from the queue if it hasn't been started
+    if data["job_state"] == "waiting":
```
plars, line 401: Ah, there's actually another state in which it would be valid to remove it - resubmitted. It's a little less common but can definitely happen.  Also, even if a job is running, we can cancel it! We won't be able to remove it from the redis list, but it's safe to call LREM on it even if it won't exist there (if that's easier).

I think it might be better to simply fail if the state is cancelled or complete. Any other state should be ok to try and cancel the job.

plars (changes requested):
Good work! I like the approach, just a few comments that I think we need to address before merging.

plars (changes requested):
I think this addresses all the concerns I had, just one minor thing noted below

plars (approved):
Looks good, thanks!

PR #16, opened 2022-07-26
Replace os.path with pathlib

nadzyah:


testflinger/__init__.py
```diff
@@ -64,8 +63,8 @@ def create_flask_app():
     # Otherwise load it from testflinger.conf in the testflinger dir
     config_file = os.environ.get("TESTFLINGER_CONFIG", "testflinger.conf")
     tf_app.config.from_pyfile(config_file, silent=True)
-    if not os.path.exists(tf_app.config["DATA_PATH"]):
-        os.makedirs(tf_app.config["DATA_PATH"])
+    if not tf_app.config["DATA_PATH"].exists():
+        Path.mkdir(tf_app.config["DATA_PATH"])
```
plars, line 67: let's go ahead and remove the conditional and just do Path.mkdir(tf_app.config["DATA_PATH"], exist_ok=True) since I don't think we need to worry about (< python3.5) at this point

testflinger/__init__.py
```diff
@@ -43,9 +44,7 @@ class DefaultConfig:
 
     REDIS_HOST = "localhost"
     REDIS_PORT = "6379"
-    DATA_PATH = os.path.join(
-        os.path.dirname(os.path.abspath(__file__)), "data"
-    )
+    DATA_PATH = Path(__file__).resolve().parent / "data"
```
plars, line 47: Did you consider using PurePath here? Any advantage to this over using something like this?
DATA_PATH = PurePath(__file__).parent / "data"

plars (changes requested):
A few minor things below, but there's a much bigger problem here I think. You have a lot of places that do something like:
`job_file = current_app.config.get("DATA_PATH") / (job_id + ".json")`
...and that will work great as long as you are getting the config from the DefaultConfig object or by setting it manually to a PathLike object the way you did in the test case. However, in production there's another option to load the config from a file. If we do that, then those paths specified in the config file will be strings, not PathLike objects.

So I think we need a better way to test that at the very least, but also, we would need to make from_pyfile smarter about which items to treat as a string vs a path.

testflinger/__init__.py
nadzyah: As I understand, if we use PurePath, we need to specify full path, because [PurePath doesn't access the filesystem](https://docs.python.org/3/library/pathlib.html#pure-paths). But with 	
Path.resolve() we can get the absolute path ([see the first row of the table here](https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module)). That's why I have decided to use Path.
Maybe I'm not right, please, tell me if I miss something

nadzyah:
> A few minor things below, but there's a much bigger problem here I think. You have a lot of places that do something like: `job_file = current_app.config.get("DATA_PATH") / (job_id + ".json")` ...and that will work great as long as you are getting the config from the DefaultConfig object or by setting it manually to a PathLike object the way you did in the test case. However, in production there's another option to load the config from a file. If we do that, then those paths specified in the config file will be strings, not PathLike objects.
> 
> So I think we need a better way to test that at the very least, but also, we would need to make from_pyfile smarter about which items to treat as a string vs a path.

@plars I added a handler for the situation when DATA_PATH variable is the string after importing it from a file.

testflinger/__init__.py
plars: resolve() does need to access the real filesystem to check symlinks, but it will only raise an exception (which is not handled here) if you set strict=True.  I'm just wondering if we need to resolve symlinks, or if having DATA_PATH potentially contain a symlink would create any problem with the way we use it. iirc, we are only reading/writing some data files with that dir as a base. We also currently use os.path.abspath, which also doesn't really inspect the filesystem.

testflinger/__init__.py
```diff
@@ -64,8 +63,9 @@ def create_flask_app():
     # Otherwise load it from testflinger.conf in the testflinger dir
     config_file = os.environ.get("TESTFLINGER_CONFIG", "testflinger.conf")
     tf_app.config.from_pyfile(config_file, silent=True)
-    if not os.path.exists(tf_app.config["DATA_PATH"]):
-        os.makedirs(tf_app.config["DATA_PATH"])
+    if isinstance(tf_app.config["DATA_PATH"], str):
+        tf_app.config["DATA_PATH"] = Path(tf_app.config["DATA_PATH"])
```
plars, line 67: Since we first load from the Default config object, we know this will be either a path or a string. We can Path() it either way, the only issue would be if it were None or something like that, which shouldn't happen. So I think you could even just skip the conditional here.

plars (commented):
one extremely minor change, otherwise looks good, thanks!

plars (approved):
+1

PR #18, opened 2022-08-26
Use mongodb instead of redis for storing data

plars:
Submitting this as a draft for now, but everything should pretty much work. It could probably use a bit more cleanup, but it's useful for testing right now, and should be completely compatible with the existing agent and cli

### DIFF
--- a/testflinger/__init__.py
+++ b/testflinger/__init__.py
@@ -19,13 +19,11 @@ This sets up the Testflinger web application
 
 import logging
 import os
-from dataclasses import dataclass
-from pathlib import Path, PurePath
 
-import redis
 from flask import Flask
 from flask.logging import create_logger
 from werkzeug.exceptions import NotFound
+from flask_pymongo import PyMongo
 
 from testflinger.api import v1
 
@@ -34,18 +32,6 @@ try:
     from sentry_sdk.integrations.flask import FlaskIntegration
 except ImportError:
     pass
-
-
-@dataclass(frozen=True)
-class DefaultConfig:
-    """
-    Default config object for Testflinger
-    """
-
-    REDIS_HOST = "localhost"
-    REDIS_PORT = "6379"
-    DATA_PATH = PurePath(__file__).parent / "data"
-    PROPAGATE_EXCEPTIONS = True
 
 
 def create_flask_app():
@@ -58,17 +44,16 @@ def create_flask_app():
         stream_handler.setLevel(logging.INFO)
         tf_log.addHandler(stream_handler)
 
-    tf_app.config.from_object(DefaultConfig)
+    tf_app.config["PROPAGATE_EXCEPTIONS"] = True
     # Additional config can be specified with env var TESTFLINGER_CONFIG
     # Otherwise load it from testflinger.conf in the testflinger dir
     config_file = os.environ.get("TESTFLINGER_CONFIG", "testflinger.conf")
     tf_app.config.from_pyfile(config_file, silent=True)
-    tf_app.config["DATA_PATH"] = Path(tf_app.config["DATA_PATH"])
-    Path.mkdir(tf_app.config["DATA_PATH"], exist_ok=True)
 
-    tf_app.redis = redis.StrictRedis(
-        host=tf_app.config["REDIS_HOST"], port=tf_app.config["REDIS_PORT"]
-    )
+    # Finally, override config data with env vars
+    tf_app.config.from_prefixed_env("MONGO_")
+
+    setup_mongodb(tf_app)
 
     sentry_dsn = tf_app.config.get("SENTRY_DSN")
     if sentry_dsn and "sentry_sdk" in globals():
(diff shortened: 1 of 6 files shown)


testflinger/__init__.py
```diff
@@ -153,4 +138,47 @@ def unhandled_exception(exc):
...
+        uri=application.config["MONGO_URI"],
+        uuidRepresentation="standard",
+    )
+    application.db = mongo_client.db
+    # PWL application.db = mongo_client
+
+    # Initialize collections and indices in case they don't exist already
+    # Automatically expire jobs after 7 days if nothing runs them
+    application.db.jobs.create_index(
+        "created_at", expireAfterSeconds=7 * 24 * 60 * 60
```
nadzyah, line 171: I guess `7 * 24 * 60 * 60` should be defined as a constant at the beginning of the file, because one may try to change it. So it's better to change it in one place instead of each line where this number is defined.

nadzyah (changes requested):

testflinger/__init__.py
```diff
@@ -153,4 +138,47 @@ def unhandled_exception(exc):
...
+    if not application.config.get("MONGO_URI"):
+        application.config[
+            "MONGO_URI"
+        ] = f"mongodb://{mongo_user}:{mongo_pass}@mongo:27017/{mongo_db}"
+
+    mongo_client = PyMongo(
+        application,
+        uri=application.config["MONGO_URI"],
+        uuidRepresentation="standard",
+    )
```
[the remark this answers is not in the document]
plars: We haven't needed to do this so far, because the connections are just between instances on the backend where the thing we're deploying on (whether it be docker/k8s/openstack/whatever should control access to only allow connections from testflinger->mongodb, so there shouldn't be a need to mitigate risk for something sitting in the middle of that. For the connections to the API server itself, we definitely should (and do) use ssl now. Is there something I'm not considering here for why we should consider adding this anyway?  If it doesn't become something that just complicates the configuration for a feature we don't use, then I'm happy to add it.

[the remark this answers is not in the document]
nadzyah: Sorry if this request is ignorant. I’m not aware of the topology of the network, where the database server and the API server are located. If they are located on the same host, we don’t need SSL/TLS. If they are located on the separate hosts, I just suggest to use TLS for communications with DB according to the Zero Trust Architecture. However, it’s up to you

testflinger/__init__.py
plars: updated this in https://github.com/plars/testflinger/commit/7a498752dabb85511d44f4a73db0bd12bdd8d36a

testflinger/__init__.py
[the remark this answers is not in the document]
plars: It's not ignorant at all! and I agree it's a good idea. I just wonder if it's a priority on an app that is completely internal and has no auth mechanism - thus you could get at the same data by going through the front door rather than by getting in the middle of a transaction between two backend services that are restricted to only talking to each other.  Just because I don't think it's a high priority today though, doesn't mean it won't ever be. I think we should add it to the backlog as something to consider in the future.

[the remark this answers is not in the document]
nadzyah: Okay, it's a good idea to add authentication mechanisms in the future.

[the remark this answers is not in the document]
plars: I've added a card for this to the backlog

testflinger/api/v1.py
```diff
@@ -378,25 +400,16 @@ def cancel_job(job_id):
...
-        data = {"job_state": "bad_data"}
-    if data["job_state"] in ["complete", "cancelled"]:
-        return "The job is already completed or cancelled", 400
-    job_file = current_app.config.get("DATA_PATH") / (job_id + ".json")
-    with open(job_file, "r", encoding="utf-8", errors="ignore") as jobfile:
-        job_data = json.load(jobfile)
-        output_key = "tf_queue_{}".format(job_data["job_queue"])
-    remove_job(output_key, job_id)
+    # remove one job from jobs where the job_id matches
+    # current_app.db.jobs.delete_one({"job_id": job_id})
```
nadzyah, line 404: Why is this line of code commented?

tests/test_v1.py
```diff
@@ -18,240 +18,275 @@
...
+    job_id = output.json.get("job_id")
     job_data["job_id"] = job_id
     output = app.post(
         "/v1/job", data=json.dumps(job_data), content_type="application/json"
     )
     result_url = "/v1/result/{}".format(job_id)
-    updated_data = json.loads(app.get(result_url).data.decode())
-    assert "resubmitted" == updated_data.get("job_state")
+    updated_data = app.get(result_url).json
+    assert "waiting" == updated_data.get("job_state")
```
nadzyah, line 106: Why do we assert it with `waiting`, not `resubmitted`? I guess it doesn't correspond to the function's title

testflinger/api/v1.py
plars: fixed

tests/test_v1.py
plars: this was a small, intentional change. We used to treat resubmitted as a different state from waiting. In retrospect, there's no real reason to do that. In both cases, the job is still in "waiting" state until something picks it up. So this test doesn't really check for a different job state now, but it does exercise both paths of job submission where there's a job_id already provided, and the one where there's no job_id provided.

nadzyah: Got it, thanks

nadzyah (approved):
lgtm

PR #19, opened 2022-09-06
Standard db env vars

plars:
The mongodb container I've been using for testing uses slightly different env vars from what we use. That's ok, but using the same ones means we can load a .env file with values that work for both rather than having to duplicate them. This also updates the testflinger.conf.example which can also be used, to reflect more current config items.


codersquid (approved):
lgtm

PR #20, opened 2022-09-08
Add the files needed for a local dev container

plars:
I've been using this locally. Give it a try and make sure it doesn't just work on my machine. It makes a really nice pattern for iterating on it locally without having to rebuild the container every time, and could also be used as the foundation for building containers that you might actually want to deploy.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,8 @@ ENV LC_ALL C.UTF-8
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US.UTF-8
 ENV PATH /srv/testflinger:$PATH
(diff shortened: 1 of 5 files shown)


nadzyah:
As I can see, the testflinger DB, db-username and db-password are not created when I up the containers. Does a user suppose to create it themselves? I expect the default DB, username and password to be created when I build the docker container. It can be done with the default values or values from the testflinger.conf file.

Maybe I miss something, but this is how I expect it to work.

nadzyah (approved):
I've checked and it works for me, +1

nadzyah (approved):
awesome, thanks for this change

PR #22, opened 2022-09-26
Allow for passing MONGODB_URI environment variable to docker image

mthaddon:
Allow for passing MONGODB_URI environment variable to the docker image.

This will override over variables, and can be used to deploy this application using https://charmhub.io/gunicorn-k8s (in the edge channel) which now supports a relation to mongodb-k8s.

### DIFF
--- a/testflinger.env
+++ b/testflinger.env
@@ -7,4 +7,4 @@
 #MONGODB_DATABASE=testflinger_db
 #MONGODB_ROOT_PASSWORD=testflinger
 # If you specify this variable the others are unneeded/ignored.
-#MONOGDB_URI=mongodb://mongo:27017/testflinger_db
+#MONGODB_URI=mongodb://mongo:27017/testflinger_db
--- a/testflinger.env
+++ b/testflinger.env
@@ -6,3 +6,5 @@
 #MONGODB_PASSWORD=testflinger
 #MONGODB_DATABASE=testflinger_db
 #MONGODB_ROOT_PASSWORD=testflinger
+# If you specify this variable the others are unneeded/ignored.
+#MONOGDB_URI=mongodb://mongo:27017/testflinger_db
--- a/testflinger/__init__.py
+++ b/testflinger/__init__.py
@@ -157,17 +157,21 @@ def setup_mongodb(application):
     mongo_pass = os.environ.get("MONGODB_PASSWORD")
     mongo_db = os.environ.get("MONGODB_DATABASE")
     mongo_host = os.environ.get("MONGODB_HOST", "mongo")
+    mongo_uri = os.environ.get("MONGODB_URI")
     if not application.config.get("MONGO_URI") and not (
-        mongo_user and mongo_pass and mongo_db
+        mongo_user and mongo_pass and mongo_db and mongo_uri
     ):
         # We are probably running unit tests
         return
 
     if not application.config.get("MONGO_URI"):
-        application.config["MONGO_URI"] = (
-            f"mongodb://{mongo_user}:{mongo_pass}@"
-            f"{mongo_host}:27017/{mongo_db}"
-        )
+        if mongo_uri:
+            application_config["MONGO_URI"] = mongo_uri
+        else:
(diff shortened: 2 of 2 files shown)


testflinger.env
```diff
@@ -6,3 +6,5 @@
 #MONGODB_PASSWORD=testflinger
 #MONGODB_DATABASE=testflinger_db
 #MONGODB_ROOT_PASSWORD=testflinger
+# If you specify this variable the others are unneeded/ignored.
+#MONOGDB_URI=mongodb://mongo:27017/testflinger_db
```
nadzyah, line 10: There's a typo. It should be "MON**GO**DB", not "MON**OG**DB"

testflinger/__init__.py
```diff
@@ -157,17 +157,21 @@ def setup_mongodb(application):
     mongo_pass = os.environ.get("MONGODB_PASSWORD")
     mongo_db = os.environ.get("MONGODB_DATABASE")
     mongo_host = os.environ.get("MONGODB_HOST", "mongo")
+    mongo_uri = os.environ.get("MONGODB_URI")
     if not application.config.get("MONGO_URI") and not (
```
nadzyah, line 161: Why is it called “MONGO_URI”, if we use “MONGO**DB**_URI”? Is it reasonable to change the value name in config?

nadzyah (changes requested):

testflinger.env
mthaddon: Ah, thx for catching, will update.

testflinger/__init__.py
mthaddon: I agree it's inconsistent. I'm not sure of the impact of changing that in application config, though

nadzyah: I guess we need @plars to take a look at this

plars: The difference is intentional here in order to use the same naming convensions as the modules and containers where the variables get used. MONGO_URI is the name of the config key that PyMongo will try to use from flask, so I thought it best to keep that as the same name for the environment variable.  This is done for similar reasons with MONGODB_{PASSWORD, DATABASE, ...} due to the expected vars in the mongodb container we are using.

If it makes you feel better, you will probably only ever use either MONGO_URI -or- the other MONGODB_* vars but not both at the same time.

nadzyah: Should MONGO_URI name also be used in these changes instead of MONGODB_URI? Or it looks fine as is?

plars: No, we don't need to manually check os.environ for that, because it will be loaded into application.config from [this ilne](https://github.com/canonical/testflinger/blob/a62184dd91a693019dd6e63749e8cd3622a740ee/testflinger/__init__.py#L59)

nadzyah (approved):
Lgtm

plars (changes requested):
We definitely need to sort some things out if we are going to merge this. This is giving new meaning to how hard "naming things" can be. It was already a bit more complicated than I think it needed to be but now it's getting worse. If we are setting the URI, then pymongo-flask needs the setting to be called "MONGO_URI", which is why I left it alone, so that the name would be consistent.  The other env vars use "MONGODB_*" because the docker container we were using for mongodb uses that naming convention.
Do we now have a new requirement for the uri to be called "MONGODB_URI" from one of the k8s operators?
Would it be better if we just don't allow setting the URI and restrict it to only allowing config through the other vars?

mthaddon:
> Do we now have a new requirement for the uri to be called "MONGODB_URI" from one of the k8s operators? Would it be better if we just don't allow setting the URI and restrict it to only allowing config through the other vars?

We don't, I was just following the convention of the other environment variables to use a `MONGODB_` prefix. This change is to allow us to pass in the full URI as an environment variable, but it can be named whatever is most appropriate - just let me know.

plars (approved):
Let's go ahead and just merge this. I think it's not really such a big deal to use the default MONGO_URI name for the var just because that's what flask_mongo want, since we are passing it anyway

PR #25, opened 2022-09-26
Add basic support for prometheus metrics

plars:


testflinger/views.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (C) 2016-2022 Canonical
```
nadzyah, line 1: I guess it should contain only 2022

testflinger/api/v1.py
```diff
@@ -26,6 +26,11 @@
 import pkg_resources
 from flask import current_app, jsonify, request, send_file
 from werkzeug.exceptions import BadRequest
+from prometheus_client import Counter
+
+
+jobs_metric = Counter("jobs", "Number of jobs", ["queue"])
+reservations_metric = Counter("reservations", "Number of reservations")
```
nadzyah, line 33: I'm not sure about that, but should there be something like `registry.register` after the metrics definition?

jocave (approved):
Looks reasonable to me for initial metrics

testflinger/api/v1.py
plars: I think this is only needed if you are registering a custom collector

Issue #26, opened 2022-09-27
Update multipass environment to work with the latest changes and mongodb


nadzyah:
The [testflinger.yaml](https://github.com/canonical/testflinger/blob/main/devel/testflinger.yaml) contains outdated information about Redis database, which we don’t use anymore. It needs to be updated, since we want to be able to run testflinger instance with MongoDB in multipass

plars:
Where are you seeing this? I believe the example .conf file was fixed in commit da37d30aab9734844432d93e3a93ad818e8aae7e

nadzyah:
> Where are you seeing this? I believe the example .conf file was fixed in commit [da37d30](https://github.com/canonical/testflinger/commit/da37d30aab9734844432d93e3a93ad818e8aae7e)

@plars check lines [69-70](https://github.com/canonical/testflinger/blob/main/devel/testflinger.yaml#L69-L70) and [125](https://github.com/canonical/testflinger/blob/main/devel/testflinger.yaml#L125)

plars:
Ah, the multipass stuff, I thought you were just talking about the example config file. Yes we need to update a LOT of stuff with that but I'm waiting on some other things to settle down first, since there seems to be a lot of debate over env vars and things like that. Let's change the title of this a bit so it's more clear.

nadzyah:
> Ah, the multipass stuff, I thought you were just talking about the example config file. Yes we need to update a LOT of stuff with that but I'm waiting on some other things to settle down first, since there seems to be a lot of debate over env vars and things like that. Let's change the title of this a bit so it's more clear.

Thank you!

PR #31, opened 2022-10-23
Update multipass dev env with all the latest changes and support for mongodb

plars:
Unfortunately we can't update to jammy yet in this environment because mongodb doesn't officially support it and I'm using their deb packages. But this gets everything working again with the current git changes.


nadzyah (approved):
Thanks, +1. Doesn't it close #26? If it does, I guess you should mention this to close the issue automatically after merging this changes ;)

PR #34, opened 2022-11-07
Improve mongo db importing and mocking

plars:
This improves the way we handle the db object, avoids needing to use current_app, and allows easier test db object creation


codersquid (approved):
lgtm though you might want to get an approval from someone who has more experience.

src/__init__.py
```diff
@@ -174,31 +175,27 @@ def setup_mongodb(application):
...
         uuidRepresentation="standard",
         serverSelectionTimeoutMS=2000,
         socketTimeoutMS=10000,
     )
-    application.db = mongo_client.db
 
     # Initialize collections and indices in case they don't exist already
     # Automatically expire jobs after 7 days if nothing runs them
-    application.db.jobs.create_index(
+    mongo.db.jobs.create_index(
```
nadzyah, line 188: Is it okay that you don't remove the `db` field here and further?

tests/test_v1.py
```diff
@@ -23,23 +23,38 @@
...
+
+
+class MongoClientMock(mongomock.MongoClient):
+    """Mock MongoClient and allow GridFS"""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        enable_gridfs_integration()
+
+    def start_session(self, *args, **kwargs):
```
nadzyah, line 37: Isn't it better to disable pylint for specific lines instead of using dirty hacks?

nadzyah (changes requested):
Thanks for your changes! Just minor things I guess we need to  improve :)

src/__init__.py
plars: Yes, "mongo" here is a PyMongo object, which contains the db object.

tests/test_v1.py
plars: I don't really have a strong opinion on whether it's dirtier to ignore the pylint warning or do this, however I think it's quite common that #noqa lines last forever and sometimes without a clear reason. Pylint was complaining that start_session() was an abstract method and that I wasn't implementing it. Looking at MongoClient, I don't *think* that seems to be true from what I'm seeing. However, I also see no harm in implementing the stub to make it happy. I don't call it directly, but perhaps something else might.

nadzyah: No problem, if you have a reason to keep it, let's try it!

nadzyah (approved):
+1

PR #35, opened 2022-11-08
Improve handling of configuration

plars:
With the recent config var changes, we moved away from using the flask app.config values. Which is fine, because we didn't really need it to be stored in the flask app config, but it also means that loading values from a python .conf file wouldn't work now.

I don't think using the flask config.from_pyfile is all that useful anyway since:
1. we can load the values we need from the environment
2. even if you want to use that config file, you can easily source it before running the flask app

This documents that recommendation in the README.rst, and makes some test changes to allow testing the error path so that the user gets an obvious error if the server isn't configured.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -40,9 +40,11 @@ except ImportError:
     pass
 
 
-def create_flask_app():
+def create_flask_app(config=None):
     """Create the flask app"""
     tf_app = Flask(__name__)
+    if config:
+        tf_app.config.from_object(config)
     tf_log = create_logger(tf_app)
 
     if not tf_app.debug:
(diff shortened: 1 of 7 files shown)


src/__init__.py
```diff
@@ -160,24 +156,23 @@ def setup_mongodb(application):
     mongo_user = os.environ.get("MONGODB_USERNAME")
     mongo_pass = os.environ.get("MONGODB_PASSWORD")
     mongo_db = os.environ.get("MONGODB_DATABASE")
-    mongo_host = os.environ.get("MONGODB_HOST", "mongo")
+    mongo_host = os.environ.get("MONGODB_HOST")
```
codersquid, line 159: Why did you get rid of the default?

testflinger.conf.example
```diff
@@ -1,8 +1,8 @@
...
-# MONGODB_USERNAME=testflinger
-# MONGODB_PASSWORD=testflinger
-# MONGODB_DATABASE=testflinger_db
-# MONGODB_HOST=mongo
-# MONGO_URI = "mongodb://mongo:27017/testflinger_db"
+# MONGODB_USERNAME="testflinger"
+# MONGODB_PASSWORD="testflinger"
+# MONGODB_DATABASE="testflinger_db"
+# MONGODB_HOST="mongo"
+# MONGODB_URI = "mongodb://mongo:27017/testflinger_db"
```
nadzyah, line 8: Just a question about formatting. Why does the last statement have "=" wrapped with the spaces, but the others don't?

src/__init__.py
plars: because it's sort of a dumb default and I'd rather they configure it. I was using it temporarily to test in docker, but it's best to insist it's configured either through the URI or the other env vars, or else exit with an error

testflinger.conf.example
plars: fixed, thanks!

nadzyah: Thanks for fixing this!

nadzyah (approved):
+1, thanks for these changes!

PR #36, opened 2022-11-10
Only consider "waiting" jobs in the queue when getting position

plars:
When querying for queue position, we are asking "how many jobs are waiting in the queue to run that came before my job". So we should restrict the query to only consider jobs in the "waiting" state.

### DIFF
--- a/tests/test_v1.py
+++ b/tests/test_v1.py
@@ -289,7 +289,7 @@ def test_job_position(mongo_app):
     """Ensure initial job state is set to 'waiting'"""
     app, _ = mongo_app
     job_data = {"job_queue": "test"}
-    # Place a 3 jobs on the queue
+    # Place 3 jobs on the queue
     job_id = []
     for pos in range(3):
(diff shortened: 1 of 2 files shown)


tests/test_v1.py
```diff
@@ -289,17 +289,32 @@ def test_job_position(mongo_app):
     """Ensure initial job state is set to 'waiting'"""
     app, _ = mongo_app
     job_data = {"job_queue": "test"}
-    # Place a job on the queue
+    # Place a 3 jobs on the queue
```
nadzyah, line 292: Just a very minor fix; I guess we don’t need “a” before “3 jobs” there

nadzyah (approved):
Looks reasonable. I’ve suggested an extremely minor fix to ensure you that I’ve looked at your changes attentively

tests/test_v1.py
plars: Thanks! I'm failing at english today :)

PR #41, opened 2023-01-25
Add api to allow posting data about the agents

plars:
Implements the piece of the dashboard spec that allows us to gather data about the agents from the agents themselves, such as the queues they are listening on, the state, etc.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -127,6 +127,13 @@ def create_flask_app(config=None):
         "/v1/agents/images", "images_post", v1.images_post, methods=["POST"]
     )
 
+    tf_app.add_url_rule(
+        "/v1/agents/data/<agent_name>",
+        "agents_post",
+        v1.agents_post,
+        methods=["POST"],
+    )
+
     @tf_app.errorhandler(NotFound)
     def handle_404(exc):
         tf_log.error("[404] Not found: %s", request.url)
--- a/src/api/v1.py
+++ b/src/api/v1.py
@@ -343,6 +343,36 @@ def images_post():
     return "OK"
 
 
+def agents_post(agent_name):
+    """Post information about the agent to the server
+
+    The json sent to this endpoint may contain data such as the following:
+    {
+        "state": string, # State the device is in
+        "queues": array[string], # Queues the device is listening on
+        "location": string, # Location of the device
+        "job_id": string, # Job ID the device is running, if any
+        "log": array[string], # push and keep only the last 100 lines
+    }
+    """
+
+    data = request.get_json()
+    # We should receive json, if so it would now be a dict
+    if not isinstance(data, dict):
+        return "Invalid data\n", 400
+    data["name"] = agent_name
+    data["updated_at"] = datetime.utcnow()
+    # extract log from data so we can push it instead of setting it
+    log = data.pop("log", [])
+
+    mongo.db.agents.update_one(
+        {"name": agent_name},
+        {"$set": data, "$push": {"log": {"$each": log, "$slice": -100}}},
+        upsert=True,
+    )
+    return "OK"
+
+
 def check_valid_uuid(job_id):
     """Check that the specified job_id is a valid UUID only
 
--- a/tests/test_v1.py
+++ b/tests/test_v1.py
@@ -408,3 +408,54 @@ def test_cancel_job_good(mongo_app):
     assert "OK" == output.text
     job = mongo.jobs.find_one({"job_id": job_id})
     assert job["result_data"]["job_state"] == "cancelled"
+
+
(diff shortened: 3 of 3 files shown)


src/__init__.py
```diff
@@ -142,8 +148,6 @@ def unhandled_exception(exc):
         tf_log.exception("Unhandled Exception: %s", (exc))
         return "Unhandled Exception: {}\n".format(exc), 500
 
-    tf_app.register_blueprint(views)
```
nadzyah, line 145: I’m wondering why this line was removed

src/api/v1.py
```diff
@@ -343,6 +343,33 @@ def images_post():
...
+        "state": string, # State the device is in
+        "queues": array[string], # Queues the device is listening on
+        "location": string, # Location of the device
+        "job_id": string, # Job ID the device is running, if any
+        "log": array[string], # push and keep only the last 100 lines
+    }
+    """
+
+    data = request.get_json()
+    data["name"] = agent_name
```
nadzyah, line 360: Shouldn't we add a verification that the posted data is correct and if it’s not, return the 400  status code?

tests/test_v1.py
```diff
@@ -408,3 +408,39 @@ def test_cancel_job_good(mongo_app):
     assert "OK" == output.text
     job = mongo.jobs.find_one({"job_id": job_id})
     assert job["result_data"]["job_state"] == "cancelled"
+
+
+def test_agents_post(mongo_app):
```
nadzyah, line 413: I guess it would be good to have “failing test(s)” as well

nadzyah (commented):
It looks great, just minor questions

src/__init__.py
plars: Good catch, thanks! It was unintentional due to git madness when I was rearranging some things

src/api/v1.py
plars: In this case, we don't always know what we're going to get. The agent is just self-reporting some data about it so that we can reflect that from the server side. Sometimes it could be telling us information like the queues it listens on, other times it could be updating the latest logs. Or is there another situation you are concerned about?

plars: hmm, yeah actually I guess if we sent it something other than json, it could wind up getting a string or something back from get_json(). Let's just check and ensure it's a dict before adding things to it, good idea!

tests/test_v1.py
plars: done.

hum4n0id:
Looks good to me.

hum4n0id (approved):

src/api/v1.py
nadzyah: I mean, what if the JSON data won't have "name" key? In this case, you'll get a KeyError and a client will get 500 status code. However, it's not a server mistake, but the client's one. So, I guess we should check is the key is present in the data, and if it's not, we should respond with 400 code.

tests/test_v1.py
nadzyah: Do I miss it? Cannot find the changes. I mean, we test the case when the data is valid and we get 200 code. Shouldn't we also have a test when the data is not valid?

src/api/v1.py
nadzyah: I mean, what if the JSON data won't have "name" key? In this case, you'll get a KeyError and a client will get 500 status code. However, it's not a server mistake, but the client's one. So, I guess we should check is the key is present in the data, and if it's not, we should respond with 400 code.

nadzyah (changes requested):

src/api/v1.py
nadzyah: Could you please add the check that the “name” and “updated_at” keys exist in the data?

tests/test_v1.py
nadzyah: Thanks!

src/api/v1.py
plars: They should not normally exist in the data that we get from get_json(). If they do, then they will be replaced. The idea here is that we will add those keys to the dict before pushing the record to the db, along with the other data sent by the agent.

nadzyah: Okay, thanks!

nadzyah (approved):
You’ve answered all my questions, thanks!

src/api/v1.py
nadzyah: _the missing comment:_ I mean, what if the JSON data won't have "name" key? In this case, you'll get a KeyError and a client will get 500 status code. However, it's not a server mistake, but the client's one. So, I guess we should check is the key is present in the data, and if it's not, we should respond with 400 code.

PR #42, opened 2023-02-07
Starting place for the frontend

plars:
This does not include the detail pages for now. There are links to them, and it is known/expected that they will 404 for now. The general pages for agent/queues/jobs exist though


nadzyah (approved):
After seeing a working demo and looking at the code, I have nothing to complain here

PR #44, opened 2023-03-06
Add a job detail view

plars:
First pass at a job details page. Here's a sample of what it might look like. I added some extra lines (which really aren't useful) in the test_cmds section so you can see the added scrollbar. Since this isn't on a server where anything is actually running the job and adding provision/test output, you don't see anything in those fields, but they would be populated if it were actually running

image

### DIFF
--- a/src/static/assets/css/testflinger.css
+++ b/src/static/assets/css/testflinger.css
@@ -11,6 +11,14 @@ main {
     width: 95%;
 }
 
+table {
+    text-align: left;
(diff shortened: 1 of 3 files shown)


nadzyah (approved):
I haven't noticed anything weird, +1

jocave (approved):
Same for me. The information displayed is what I would expect to see.

Design in not a great concern at this point.

PR #45, opened 2023-03-08
Add queue details page

plars:
queue details view - including a list of jobs in the queue. Interesting thing to note from this screenshot is that this is NOT an advertised queue, but you can still at least see the name and list of jobs in the queue
image

### DIFF
--- a/src/views.py
+++ b/src/views.py
@@ -96,3 +96,24 @@ def queues():
         ]
     )
     return render_template("queues.html", queues=queue_data)
+
+
+@views.route("/queues/<queue_id>")
+def queue_detail(queue_id):
+    """Queue detailed view"""
+    queue_data = mongo.db.queues.find_one({"name": queue_id})
+    if not queue_data:
+        # If it's not an advertised queue, create some dummy data
+        queue_data = {"name": queue_id, "description": "No description"}
+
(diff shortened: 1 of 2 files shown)


src/views.py
```diff
@@ -96,3 +96,23 @@ def queues():
...
+    """Queue detailed view"""
+    queue_data = mongo.db.queues.find_one({"name": queue_id})
+    if not queue_data:
+        # If it's not an advertised queue, create some dummy data
+        queue_data = {"name": queue_id, "description": "No description"}
+
+    job_data = mongo.db.jobs.find(
+        {
+            "job_data.job_queue": queue_id,
+            "result_data.job_state": {"$nin": ["complete", "cancelled"]},
```
jocave, line 112: Could you explain what this is doing? Sorry I can't work that out.

Also could the commented out code below be removed?

jocave (changes requested):
Request for info below please

src/views.py
plars: That finds all the active jobs in this queue so that we can display it in the view. I added a comment to that effect just now, so see if that's better. I also removed that line that was commented out. Thanks!

jocave (approved):

PR #46, opened 2023-04-16
Add agent detail view

plars:
Agent details view for CERTTF-80
Here's a screenshot from some testing I did locally
image

### DIFF
--- a/src/templates/agent_detail.html
+++ b/src/templates/agent_detail.html
@@ -1,0 +1,18 @@
+{% extends 'base.html' %}
+{% set active_page = 'agents' %}
+{% block content %}
+<div class="center">
+    Agent Name: {{ agent.name }}<br>
+    State: {{ agent.state }}<br>
+    Location: {{ agent.location }}<br>
+    Last Updated: {{ agent.updated_at }}<br>
+    <br>
+    <br>
+    Agent Log:
+    <div class="center scrollable">
+        <span style="white-space: pre-line">
+            {{ agent.log|join('\n') }}
+        </span>
+    </div>
+</div>
+{% endblock %}
--- a/src/views.py
+++ b/src/views.py
@@ -37,6 +37,13 @@ def agents():
     return render_template("agents.html", agents=agent_info)
 
 
+@views.route("/agents/<agent_id>")
+def agent_detail(agent_id):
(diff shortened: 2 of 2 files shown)


hum4n0id (approved):
lgtm

src/templates/agent_detail.html
```diff
@@ -0,0 +1,18 @@
+{% extends 'base.html' %}
+{% set active_page = 'agents' %}
+{% block content %}
+<div class="center">
+    Agent Name: {{ agent.name }}<br>
```
nadzyah, line 5: Is it a good practice in frontend to use this many `<br>` elements? Can we replace it using `<p>` or row block for each property?

plars: By "row block" I assume you mean putting it in a table?
I experimented with both. The table approach felt pretty heavy, but perhaps with some additional styling it could work. I think the `<p>` would need quite a bit of css help as well, as it looks pretty weird natively:
image

We've used `<br>` in the other templates as well, so while I agree that we have a lot of work to make this prettier and I make no claims of being a good gui designer, I think it might make more sense to go in later and take a consistent approach to updating all of them at once.

nadzyah: No, I don’t mean tables. You can you containers and rows inside them (like `<div class=“row”`. You can include the vanilla css file (by using its link), so you don’t have to write any css file for it yourself

nadzyah: Here is the doc on how to apply it (the Hotlink section)
https://vanillaframework.io/docs
Actually, it’s just one line of code

plars: Yeah, I definitely like the idea of moving to vanilla, but I think it's going to require a little more than just importing it and using this one element. There seem to be some side effects:
image

We do have a jira issue for looking at converting to vanilla though, and adopting a more ubuntu-like theme. Or also, there's a chance still that we might replace the UI completely with something flutter based, so it may not matter eventually if that happens.

nadzyah: Okay, you’re right, and we can keep it this way for now but rewrite it in future when we have a specific task for improving frontend design

nadzyah (approved):

PR #48, opened 2023-04-28
Fix test for bad job difference when using Flask 2.3.x

plars:
Small change needed for tests to pass in more recent versions of flask


hum4n0id (approved):
lgtm

PR #49, opened 2023-04-28
Handle potential for special chars in db passwords

plars:
If you have chars that can't be encoded into the mongo URI, it will get errors, this fixes that.

### DIFF
--- a/src/__init__.py
+++ b/src/__init__.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2016-2022 Canonical
+# Copyright (C) 2016-2023 Canonical
 #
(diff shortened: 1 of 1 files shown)


plars:
I definitely agree @nadzyah ! Unfortunately, in this case, it's pymongo which gave me the error, which isn't something I can really reach into with a unit test. With some restructuring of all this, some thing else would be responsible for gathering the credentials in their proper form, and we could test what comes out of that. I'd like to do a lot of that sort of cleanup, but it also takes a considerable amount of time and though. And in this case, it's straightforward enough that I think we're ok for now. But someday, yes... I would definitely like to rework a lot of things :)

PR #50, opened 2023-04-28
Redirect / to /agents for now

plars:
Once this gets deployed, it's not useful to have people just get a version string when they go to the index page. Let's redirect it to /agents unless we later decide on a better landing page for the index.

### DIFF
--- a/src/views.py
+++ b/src/views.py
@@ -17,11 +17,17 @@
 Additional views not associated with the API
 """
 
-from flask import Blueprint, render_template
+from flask import Blueprint, render_template, redirect, url_for
 from prometheus_client import generate_latest
 from src.database import mongo
 
 views = Blueprint("testflinger", __name__)
+
+
+@views.route("/")
+def home():
+    """Home view"""
+    return redirect(url_for("testflinger.agents"))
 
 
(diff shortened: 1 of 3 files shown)


src/views.py
```diff
@@ -17,13 +17,19 @@
...
-from flask import Blueprint, render_template
+from flask import Blueprint, render_template, redirect, url_for
 from prometheus_client import generate_latest
 from src.database import mongo
 
 views = Blueprint("testflinger", __name__)
 
 
+@views.route("/")
+def home():
```
hum4n0id, line 28: This is more robust; LGTM.

hum4n0id (approved):

tests/test_v1.py
```diff
@@ -20,17 +20,9 @@
 import json
 
 from io import BytesIO
-import src
 from src.api import v1
 
 
-def test_home(mongo_app):
```
nadzyah, line 27: Shouldn't we test here that a user gets redirected?

plars: Good idea, unit test added

nadzyah (approved):
Great job!

PR #51, opened 2023-05-23
Add settings for the auth source db to use for mongodb

plars:
So I *think* this is what we need for getting things working with the mongodb version used by mongodb-k8s charm. I've tried testing this by building the container and deploying it in juju, but so far I haven't had any luck getting juju to re-pull the oci image. So for now, I'm testing it by connecting to the container, making the changes, and manually starting testflinger.

Without this, it will try to authenticate against the testflinger_db rather than the admin db. If you use the default of admin db, you don't have to specify the authentication source database, but I added the env var in case you use something like the bitnami mongodb image, so that it will be more flexible.

I also added documentation in the README about the new option

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -69,6 +69,8 @@ Currently supported configuration values are:
 
 - **MONGODB_DATABASE**: Name of the MongoDB database to use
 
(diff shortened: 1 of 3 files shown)


hum4n0id (commented):
Looks good

PR #52, opened 2023-05-25
Add basic pieces for an initial charm

plars:
This is currently pushed to charmhub and can be installed from there. It works, in so much as you can relate a mongodb charm to it and it can connect to the database. However, I haven't added ingress support yet. That's still coming.


charm/metadata.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+containers:
+  testflinger:
+    resource: testflinger-image
+
+resources:
+  testflinger-image:
+    type: oci-image
+    description: OCI image for the 'testflinger' container
+    upstream-source: plars/testflinger:latest
```
nadzyah, line 16: Is there any canonical testflinger image? Or are we okay with using yours?

charm/src/charm.py
```diff
@@ -0,0 +1,186 @@
...
+        self.framework.observe(
+            self.on.testflinger_pebble_ready, self._on_testflinger_pebble_ready
+        )
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
+
+    @property
+    def version(self) -> str:
+        """Report the current version of the app"""
+        # TODO: get the version somehow and return it
+        return "Version ?"
```
nadzyah, line 56: What if we get it from a docker image version? Is it possible?

charm/src/charm.py
```diff
@@ -0,0 +1,186 @@
...
+        initial = dict(self._stored.reldata["mongodb"])
+        self._stored.reldata["mongodb"].update(
+            self.mongodb.fetch_relation_data()[event.relation.id]
+        )
+        if initial != self._stored.reldata["mongodb"]:
+            self._update_layer_and_restart(None)
+
+    def _update_layer_and_restart(self, event) -> None:
+        """Define and start layer for testflinger using Pebble"""
+
```
nadzyah, line 83: I guess this blank line is redundant. Have you run black on these files?

charm/src/charm.py
```diff
@@ -0,0 +1,186 @@
...
+        # add workload version in juju status
+        self.unit.set_workload_version(self.version)
+        self.unit.status = ops.ActiveStatus()
+
+    def _on_mongodb_client_relation_removed(
+        self, event: ops.framework.EventBase
+    ) -> None:
+        """Event is fired when relation with mongodb is broken."""
+        self.unit.status = ops.WaitingStatus("Waiting for database relation")
+        raise SystemExit(0)
```
nadzyah, line 114: `sys.exit(0)` is a preferable way to raise this exception. See https://docs.python.org/3/library/exceptions.html#SystemExit

charm/src/charm.py
```diff
@@ -0,0 +1,186 @@
...
+    @property
+    def app_environment(self) -> dict:
+        """Get dict of env data for the mongodb credentials"""
+        db_data = self.fetch_mongodb_relation_data()
+        env = {
+            "MONGODB_HOST": db_data.get("db_host", None),
+            "MONGODB_PORT": db_data.get("db_port", None),
+            "MONGODB_USERNAME": db_data.get("db_username", None),
+            "MONGODB_PASSWORD": db_data.get("db_password", None),
+            "MONGODB_DATABASE": db_data.get("db_database", None),
```
nadzyah, line 155: It’s None by default, so you don't need to specify it explicitly

nadzyah (changes requested):
There are just minor things I’ve noticed in the code. However, in general, it looks excellent (as usual☺️)

charm/metadata.yaml
plars: Not at this point, but it probably won't matter as there doesn't seem to be a way that I've found to force it to pull unless you specify a new tag every time. It always wants to use a cached copy. However, this doesn't seem to be the way to get the image anyway. When you upload the charm to the charm store, you also have to manually upload any resources (including the image) or else it won't work. So it won't actually get it from here anyway.

Also, from the juju documentation it tells us that this field is ignored by juju. It's just for reference.

charm/src/charm.py
plars: Not from what I can tell. We should probably add some sort of get version api in testflinger itself. It's not critical but I liked this idea from the examples I saw and wanted to leave it there as a placeholder in the charm.

charm/src/charm.py
plars: I do run black but I don't think it complained about this. I removed the line though.

charm/src/charm.py
plars: done

charm/src/charm.py
plars: done

nadzyah (approved):
Thanks for answering to all my questions! +1

PR #54, opened 2023-06-09
Add support for nginx-ingress-integrator to the juju charm

plars:
I've tested this locally with microk8s, which requires the "ingress" plugin enabled to work with it. Once deployed, I was able to easily connect to it though.

### DIFF
--- a/charm/README.md
+++ b/charm/README.md
@@ -1,0 +1,68 @@
+# Deploying with Juju
+
+To deploy Testflinger server using Juju, there are 3 main charms you'll need:
+ - mongodb-k8s
+ - nginx-ingress-integrator
+ - testflinger-k8s
+
+All of these can currently be pulled from charmhub, but there are a few extra
+steps needed for configuration and integration listed below. You'll also need
+a juju and k8s environment set up ahead of time. There are good instructions
+for doing this with microk8s here: https://juju.is/docs/olm/microk8s
+
+Additionally, you'll need to enable the following add-ons in microk8s:
+ - dns
+ - hostpath-storage
+ - ingress
+
+## Deploy mongodb-k8s
+
+For a simple test deployment, it is sufficient to just run:
+```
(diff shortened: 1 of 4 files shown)


plars:
Added a README for the charm deployment. This will likely change quite a bit in the future and hopefully get easier. But this reflects the current state of how to deploy the charms and configure them.

charm/README.md
```diff
@@ -0,0 +1,68 @@
...
+## Integrate (relate) the Charms
+
+To make the charms talk together, you'll need to run:
+```
+    $ juju integrate testflinger-k8s nginx-ingress-integrator
+    $ juju integrate testflinger-k8s mongodb-k8s
+```
+
+After this, watch `juju status` and/or `juju debug-log` for progress.
+Once everthing is settled, you should be able to point your web browser
```
mz2, line 66: everything

mz2 (approved):
Looked good to me.

I didn't notice changes that remove traefik from there, not sure it was there before (reviewing on mobile) or there was no pre-existing ingress.

charm/README.md
plars: fixed, thanks!

charm/src/charm.py
```diff
@@ -56,6 +59,14 @@ def version(self) -> str:
         # TODO: get the version somehow and return it
         return "Version ?"
 
+    def _require_nginx_route(self):
+        require_nginx_route(
+            charm=self,
+            service_hostname=self.config["external_hostname"],
+            service_name=self.app.name,
+            service_port=5000,
```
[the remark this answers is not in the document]
plars: I considered that, but I couldn't come up with a good reason why it should be an extra config item since it's only the internal port. Is there ever a need to change this that you can think of?

plars:
I had a traefik version in a local branch but reconsidered whether to use traefik or nginx before actually submitting, so that one didn't make it to a PR just yet.

PR #59, opened 2023-06-20
Update upstream image source

plars:


omar-selo (approved):

plars:
> Does the association with 'main' imply this is the image that we expect to be used on staging?
In this case, it's just that we're tagging the image "main" because it's built from the main branch on GH.  Initially, I think that's likely what we'll track for staging, unless there's some reason later to change it.

PR #68, opened 2023-06-22
Add charm makefile and libs

plars:


charm/Makefile
```diff
@@ -0,0 +1,6 @@
+build: fetch-libs
+	charmcraft pack
+
+fetch-libs:
+	 charmcraft fetch-lib charms.data_platform_libs.v0.data_interfaces
+	 charmcraft fetch-lib charms.nginx_ingress_integrator.v0.nginx_route
```
omar-selo, line 6: Since the libs are already checked in. Is there a need to fetch them?

plars: It will also update them if there are any updates, and shows clearly how to get them which I personally find helpful :)

omar-selo (approved):

PR #71, opened 2023-06-23
Charm unit tests

plars:
Add unit tests for the juju charm

### DIFF
--- a/charm/src/charm.py
+++ b/charm/src/charm.py
@@ -104,9 +104,9 @@ class TestflingerCharm(ops.CharmBase):
             self.mongodb.fetch_relation_data()[event.relation.id]
         )
         if initial != self._stored.reldata["mongodb"]:
-            self._update_layer_and_restart(None)
+            self._update_layer_and_restart()
 
-    def _update_layer_and_restart(self, _) -> None:
+    def _update_layer_and_restart(self) -> None:
         """Define and start layer for testflinger using Pebble"""
         if not self.container.can_connect():
             self.unit.status = ops.WaitingStatus(
@@ -133,14 +133,22 @@ class TestflingerCharm(ops.CharmBase):
         self.unit.set_workload_version(self.version)
         self.unit.status = ops.ActiveStatus()
 
-    def _on_mongodb_client_relation_removed(self, _) -> None:
-        """Event is fired when relation with mongodb is broken."""
+    def _on_mongodb_client_relation_removed(
+        self, _: ops.framework.EventBase
+    ) -> None:
+        """
+        Event is fired when relation with mongodb is broken.
+        We need to accept the event as an argument, but we don't use it
+        """
         self.unit.status = ops.WaitingStatus("Waiting for database relation")
         sys.exit()
 
-    def _on_config_changed(self, event: ops.framework.EventBase) -> None:
-        """Handle config changed event"""
-        self._update_layer_and_restart(event)
+    def _on_config_changed(self, _: ops.framework.EventBase) -> None:
+        """
+        Handle config changed event
+        We need to accept the event as an argument, but we don't use it
+        """
+        self._update_layer_and_restart()
 
     @property
     def _pebble_layer(self):
--- a/charm/src/charm.py
+++ b/charm/src/charm.py
@@ -1,12 +1,29 @@
 #!/usr/bin/env python3
-# Copyright 2023 Paul Larson
-# See LICENSE file for licensing details.
+# Copyright (C) 2023 Canonical
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


charm/src/charm.py
```diff
@@ -56,7 +72,6 @@ def __init__(self, *args):
     @property
     def version(self) -> str:
         """Report the current version of the app"""
-        # TODO: get the version somehow and return it
```
nadzyah, line 59: Isn't it still TODO?

charm/src/charm.py
```diff
@@ -90,7 +105,7 @@ def _on_mongodb_client_relation_changed(
         if initial != self._stored.reldata["mongodb"]:
             self._update_layer_and_restart(None)
 
-    def _update_layer_and_restart(self, event) -> None:
+    def _update_layer_and_restart(self, _) -> None:
```
nadzyah, line 108: I’m wondering why you need to specify an argument you won't use in a private method. Can we just not pass it?

charm/src/charm.py
```diff
@@ -117,9 +132,7 @@ def _update_layer_and_restart(self, event) -> None:
         self.unit.set_workload_version(self.version)
         self.unit.status = ops.ActiveStatus()
 
-    def _on_mongodb_client_relation_removed(
-        self, event: ops.framework.EventBase
-    ) -> None:
+    def _on_mongodb_client_relation_removed(self, _) -> None:
```
nadzyah, line 135: Same question

charm/src/charm.py
plars: It is, but pylint doesn't like that and will block it. I could tell it to ignore it, but I'm not sure that's better than just leaving out the comment

charm/src/charm.py
plars: Ok, done, and added a few comments around all of this _ event nonsense. Let me know what you think.

charm/src/charm.py
nadzyah: Yeah, a comment would be great

charm/src/charm.py
nadzyah: Okay, we can remove it since the “?” version indicates that we need to change it

nadzyah:
Btw, what do you think about using `*args` instead of _ argument?

plars:
> Btw, what do you think about using `*args` instead of _ argument?

I don't like it for this case. That's for when you don't know how many arguments you will receive. We *know* how many args it will get, but we just don't need to actually use it. I think that is better conveyed by _

charm/tests/unit/test_charm.py
```diff
@@ -0,0 +1,129 @@
...
+import pytest
+import yaml
+
+import ops
+import ops.testing
+from charm import TestflingerCharm
+
+CHARMDIR = Path(__file__).parent.parent.parent.resolve()
+METADATA_FILE = CHARMDIR / "metadata.yaml"
+# METADATA = str(yaml.safe_load(Path("metadata.yaml").read_text()))
```
nadzyah, line 30: should we remove this comment bc it's needed anymore?

charm/tests/unit/test_charm.py
```diff
@@ -0,0 +1,129 @@
...
+    harness.begin()
+    with pytest.raises(SystemExit):
+        assert harness.container_pebble_ready("testflinger")
+    assert isinstance(harness.model.unit.status, ops.WaitingStatus)
+
+
+def test_good_relation_active_status(harness):
+    """status should be active when the relation is present"""
+    harness.container_pebble_ready("testflinger")
+    harness.set_can_connect("testflinger", True)
```
nadzyah, line 52: I'm wondering if it is worth making "testflinger" a constant (e.g. `PROJECT_NAME`) bc if you change it (for instance, make it uppercase or change spelling for some reason, you'll have to do it in all the places in the project

charm/tests/unit/test_charm.py
plars: done, thanks!

plars: done, thanks!

charm/tests/unit/test_charm.py
plars: I think it actually maps to the "PEBBLE_SERVICE_NAME" from a juju perspective if you look at the charm. I suppose if we were going to do that, we would define it in the charm and import it from there. However I'm not really sure that adds anything to the clarity (for me at least). Also, if we ever change the name from "testflinger" to something else, there will be a LOT more changes than this that are needed.

I'm not opposed to it though. If you think it makes it more readable for you, I'm happy to change it. Just that for me, I don't think it helps since I would need to go figure out what the value of that is whenever I see it.

nadzyah: if you're okay to keep it, we can keep it :)

nadzyah (approved):

PR #72, opened 2023-07-02
Add terraform deployment plan and README

plars:


terraform/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+The `testflinger.tf` plan in this directory can be used for deploying
+the Testflinger server into a dev, staging, or production environment.
+For development, it is recommended to use Juju with Microk8s.
+
+
+## Deploying an Environment
+
+The following assumes that you have an existing Juju + k8s environment
+with a `k8s-certification` credential imported in juju for connecting
```
mz2, line 11: Actually you lost me here, I have not had to do this with microk8s locally, I don't think. What needs to be done on Charmed Kubernetes or microk8s to accomplish this?

terraform/README.md
```diff
@@ -0,0 +1,67 @@
...
+a new model to the k8s environment.
+
+For purposes of this example, we will be deploying a **staging**
+environment.
+
+This also requires that you import the TLS certificate for https into
```
mz2, line 17: This would be clearer to explain where this is needed to be done (step 3), I'm not clear why this paragraph is here?

terraform/README.md
```diff
@@ -0,0 +1,67 @@
...
+    $ juju deploy -m testflinger-staging --channel=5/edge mongodb-k8s --storage mongodb=10G
+```
+
+3. (optional) Export environment variables to be used in the following steps. If
+you choose not to do this, you'll need to either specify them on the command line
+for each command, or answer the interactive prompts from terraform.
+```
+    $ export TF_VAR_environment=staging
+    $ export TF_VAR_external_ingress_hostname=testflinger-staging.canonical.com
+    $ export TF_VAR_tls_secret_name=testflinger-staging-canonical-com-tls
```
mz2, line 40: terraform state handling will add to this & the following `terraform init` step. Fine to get this merged in this shape but we'll need that really before we start using this in anger.

terraform/README.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+6. Running `terraform plan` will show us what it intends to do without
+actually doing it. Look at the output from this and make sure it looks
+correct.
+```
+    $ terraform plan
+```
+
+7. Finally, apply the terraform plan. After this settles, everything should
```
mz2, line 63: Worth also noting after this as a separate section what to do past this initial setup (just `terraform apply` all the way down after that) when you want to make changes to the system.

terraform/testflinger.tf
```diff
@@ -0,0 +1,84 @@
+terraform {
+  required_providers {
+    juju = {
+      version = "~> 0.7.0"
+      source  = "juju/juju"
+    }
+  }
```
mz2, line 7: There'll be a need to describe a "backend" here also when you add state handling support in. I don't mind if that's in this PR or in a follow-up one.

mz2 (approved):
Some suggested tweaks to the README, looks good and understandable with the exception of the bit about credentials which I'm too noob for.

terraform/README.md
plars: I can see how that would be confusing. I'll provide operational documentation that is specific to staging or production for things like this, and instead focus this in-repo README on deploying a dev environment.

terraform/README.md
plars: How does the state handling keep track of these vars? From what I've observed so far, I either need to provide them on the command line, in the environment, or answer prompts every time I run terraform in order for it to have these.

plars: How does the state handling keep track of these vars? From what I've observed so far, I either need to provide them on the command line, in the environment, or answer prompts every time I run terraform in order for it to have these.

terraform/testflinger.tf
plars: noted, still need to sort that bit out, but it can be part of a future update, or more likely as operational documentation with further details about staging/production deployments in our environment.

terraform/README.md
plars: Added a comment about this

terraform/README.md
plars: Moved this around a bit and reworded. I also added details on how to create and import a self-signed certificate, since I think it's helpful to have it here as an optional step. Hopefully more clear now?

plars:
I made some updates, see what you think if you have a chance.

PR #73, opened 2023-07-07
Fix issues with refreshing the charm and containers

plars:


omar-selo (approved):

charm/src/charm.py
```diff
@@ -191,26 +191,31 @@ def app_environment(self) -> dict:
...
     def fetch_mongodb_relation_data(self) -> dict:
         """Get relation data from the mongodb charm"""
-        data = self._stored.reldata.get("mongodb", {})
+        data = self.mongodb.fetch_relation_data()
         logger.debug("Got following database data: %s", data)
         if not data:
             self.unit.status = ops.WaitingStatus(
                 "Waiting for database relation"
             )
             raise SystemExit(0)
```
nadzyah, line 200: According to the doc, the best way to raise this exception is using `sys.exit(0)`
see https://docs.python.org/3/library/exceptions.html#SystemExit

nadzyah (changes requested):

nadzyah (approved):

PR #76, opened 2023-07-13
Detect if we have empty relation data and exit

plars:
There's a chance that the relation can exist but not yet have data in it. This can result in an error if we try to use that data later, and the charm will get stuck in an error state rather than retrying. Exiting here if there's no data for the relation forces it to retry getting the data later.

### DIFF
--- a/charm/src/charm.py
+++ b/charm/src/charm.py
@@ -210,6 +210,11 @@ class TestflingerCharm(ops.CharmBase):
             username = val.get("username")
             password = val.get("password")
             database = val.get("database")
+            break
+        else:
+            logger.error("No database relation data found yet")
+            sys.exit()
+
         db_data = {
             "db_host": host,
             "db_port": port,


charm/src/charm.py
```diff
@@ -210,6 +210,11 @@ def fetch_mongodb_relation_data(self) -> dict:
             username = val.get("username")
             password = val.get("password")
             database = val.get("database")
+            break
+        else:
+            logger.error("No database relation data found yet")
+            sys.exit()
```
nadzyah, line 216: Probably it's a personal preference, but to me this code is kinda hard to read. It took some time to understand what's going on in the loop.

nadzyah (commented):
Sorry if this issue is kind of my ignorance, I'm just wondering if we can make it simpler

charm/src/charm.py
plars: Yeah, I had a similar thought about it, but was relucatant to change from the pattern I've seen other charms use. This is looping through relation data that is sent to us by an event. I've only ever personally seen it send one k:v pair when it works, so in theory you could just count the entries and if it's >0 then take the first one.

However I'm not sure why every example I've seen seems to indicate that there *could* be more than one entry. I can change it, but either way, I think it's going to be very hard to debug if anything ever goes wrong with it. Forutnately we should only hit this section very soon after deployment

PR #77, opened 2023-07-13
Include the snap install steps for convenience

jocave:


terraform/README.md
```diff
@@ -5,16 +5,30 @@ the Testflinger server into a dev, staging, or production environment.
...
+
+```
+sudo snap install microk8s --channel=1.27-strict/stable
+sudo usermod -a -G snap_microk8s $USER
+sudo chown -f -R $USER ~/.kube
+newgrp snap_microk8s
+sudo microk8s enable hostpath-storage dns ingress
+sudo snap alias microk8s.kubectl kubectl
+sudo snap install juju --channel=3.1/stable
+mkdir -p ~/.local/share
```
plars, line 24: I don't recall ever having to do a lot of these steps, especially the one for .local/share. What's that for? Not a problem, just curious.

plars (approved):
Just a question, but nothing blocking. +1

terraform/README.md
jocave: That's a workaround that can be needed if you're using juju from the snap confined

jocave: The other steps are similarly probably only needed if running microk8s confined.

plars: ah, ok. Makes sense

PR #79, opened 2023-07-18
Add github actions for releasing the charm to edge

plars:


.github/workflows/charm_release_edge.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Upload charm to charmhub
+        uses: canonical/charming-actions/upload-charm@2.3.0
+        with:
+          charm-path: charm
+          credentials: "${{ secrets.CHARMHUB_TOKEN }}"
```
omar-selo, line 21: Just for reference, when implementing test observer's action. What's charmhub token?

omar-selo (approved):
LGTM

.github/workflows/charm_release_edge.yml
plars: You would need to upload a new one under the test_observer project called "CHARMHUB_TOKEN". I just created one with my own account for now by using `charmcraft login --export /tmp/charmcraft.auth --charm=testflinger-k8s --permission=package-manage --channel=edge --ttl=2592000`
However, I'm working at figuring out who to pester so that we can get the ownership changed on those charms. So hopefully we can use the bot account soon!

omar-selo: oh yeah forgot about `charmcraft login --export`. But yeah moving ownership is an issue

PR #83, opened 2023-08-03
Filter "complete" jobs in the job view

plars:


hum4n0id (approved):
Lftm

hum4n0id (approved):
lgtm

tests/test_v1.py
```diff
@@ -363,7 +363,7 @@ def test_get_invalid(mongo_app):
     assert 404 == output.status_code
 
 
-def test_cancel_job_completed(mongo_app):
+def test_cancel_job_complete(mongo_app):
```
jocave, line 366: seems like the test should stay as _completed ?

jocave (changes requested):

PR #88, opened 2023-08-23
Use a blueprint for v1 api

plars:
This is much easier to read, removes a lot of code, and sets things up better for generating openapi docs


omar-selo (approved):
Looks neat!

PR #94, opened 2023-09-13
Use vanilla

plars:
This adds vanilla framework css and adjusts some colors to make it more Ubuntu-like


src/static/assets/css/testflinger.css
```diff
@@ -26,12 +26,13 @@ main {
 
 /* Header bar related */
 .header {
+    color: white;
     overflow: hidden;
-    background-color: #3f3f3f;
+    background-color: #333;
```
nadzyah, line 31: Is this correct? Sorry if I don't know something about the hex format for colours

plars:
Here's what it looks like, if that helps:
image

nadzyah (approved):
Wow wow wow, really great changes! 
One minor question but still +1 from me

src/static/assets/css/testflinger.css
plars: correct? I have no idea. Fun fact, I'm told that I'm colorblind, which in my case means that I see colors, but I think not as many as others do. And none of them make any sense to me. I have exactly zero understanding of what colors look "good" or "bad" when used together.

However, it was intentional because I was trying to use https://vanillaframework.io/docs/settings/color-settings https://design.ubuntu.com/brand/colour-palette and https://ubuntu.com for ideas. IIRC, that one came from the header bar on ubuntu.com. However I see lots of differences when compared to other sites like snapcraft.io, maas.io, events.canonical.com, etc...

I think the point here is just to make it feel a bit more like something that fits in the canonical/ubuntu ecosystem

nadzyah: Yeah, I was wondering about 333 and not a 6-length number. Thanks for the clarification!

mz2: Yes, it's valid indeed: https://www.w3.org/TR/css-color-3/#rgb-color

PR #95, opened 2023-09-13
Don't show microseconds in datestamps

plars:


plars:
Previous:
image

Now:
image

mz2 (approved):

PR #96, opened 2023-09-13
Don't wrap table cell contents and increase row size

plars:


plars:
An unintended side-effect of styling that we got with vanilla makes things like this show up a lot more often:
image

This small fix makes it quite a bit nicer:
image

mz2:
Hmm, but the truncation I see in the "after" picture is unwanted too I think (you can't paste the UUID out if it's truncated). I think you want `overflow: hidden` or some other way to make it actually overflow.

plars:
I didn't have much luck with overflow: hidden, but I found another tag that proved to be even more useful for auto-resizing the columns. It won't stop wrapping completely if there's something really long, but it should do a lot more to avoid it when it would make things like the job id look strange without truncating the end of it. See if you like this better:
image

mz2 (approved):

PR #100, opened 2023-09-19
Minor terraform module stuff

plars:
A lot of this is just structural, but the important bit is removing the empty provider, which seems to interfere whenever I try to use this as a module from something else.


terraform/main.tf
```diff
@@ -1,38 +1,3 @@
-terraform {
-  required_providers {
-    juju = {
-      version = "~> 0.7.0"
-      source  = "juju/juju"
-    }
-  }
-}
-
-provider "juju" {}
```
omar-selo, line 10: Oh I had no clue that this is optional...

omar-selo (approved):
LGTM!

PR #112, opened 2023-10-12
add reference docs for device type and job examples

tang-mm:
Add doc for Supported Provisioning methods and example definition
(moved from https://github.com/canonical/certification-lab-docs/tree/main/reference)


docs/reference/provisioning-methods.rst
```diff
@@ -0,0 +1,74 @@
+Provisioning Methods
```
mz2, line 1: I think only this file actually really belongs in the testflinger docs as a reference, the others are a straightforward fit to Library.

This indeed should also be a _reference_ so that there's an example configuration included (and a hello-world type payload at the top of the file indicating where the choice of the provisioning method goes?)

mz2:
I think this should effectively be split in two: you have a short reference section there needed into this repo, but most of this should move into Library. When you make this move, please also make an associated removal at the other end (or link in those docs where the current location of  the up to date copy is).

tang-mm:
Updated per comments:
- kept only the provisioning reference doc
- added examples for `provision_data` fiedl

docs/reference/provisioning-methods.rst
```diff
@@ -0,0 +1,120 @@
...
+
+Provision data are specified as a key-value pair in the ``provisioning_data`` section of a Testflinger job definition file:
+
+.. code-block:: yaml
+
+  job_queue: my-queue-name
+  provision_data:
+    key: value    # Data to pass to the provisioning step
+
+The type of provision data usually varies on device types. For example, our MAAS provisioner currently only needs to to know which ``distro`` series to use, whereas the devices that install all-snaps images (such as Dragonboard) may need to know which channel to use when building the image instead.
```
jocave, line 14: The final sentence does not match the device_connector or the table below - Dragonboard takes a URL not a channel

jocave (changes requested):

docs/reference/provisioning-methods.rst
tang-mm: fixed

tang-mm:
Closed as some contents were covered by https://github.com/canonical/testflinger/pull/113

PR #128, opened 2023-10-16
Create .jira_sync_config.yaml

jocave:
Establishing and initial gh to jira sync config

## Description

Config for the gh jira sync bot


plars (approved):
Nice! I assume there are other steps to set up this linkage on the jira side, are those done already or did they need to wait for this to land?

jocave:
> Nice! I assume there are other steps to set up this linkage on the jira side, are those done already or did they need to wait for this to land?

I don't think there are any other steps - we would need to close & re-open existing issues if we want them synced though. If it does't "just work" we should get in touch with Maksim.

PR #131, opened 2023-10-17
Move testflinger-server code to monorepo layout

plars:
## Description

This moves the testflinger-server code into the structure that we need for monorepo

## Resolved issues

CERTTF-166 (partially)

## Documentation

Docs directory already exists and was moved to the root of the monorepo along with the .readthedocs.yaml

## Tests

In progress...


jocave (commented):
Bit of bike-shedding - I feel like the `testflinger-` prefix on all the component sub directories is redundant. Should we remove it?

plars:
> Bit of bike-shedding - I feel like the `testflinger-` prefix on all the component sub directories is redundant. Should we remove it?

Just to clarify, do you mean just from the directory name? For instance, things like "testflinger-agent" or especially the user facing "testflinger-cli" would be a little confusing if you executed them as just "agent" or "cli". So I think internally, keeping testflinger as part of the name is ok. I'm ok with removing it from the names of the directories if you think that looks cleaner though. The directory structure would then look like this:
/testflinger
    /agent
    /cli
    /device-connectors
    /server
...but the cmdline used to call them would still be the same as it was previously.
The is one small advantage I can see to keeping it as-is. We know the agent charm will need some modification. Each agent currently gets a dir on the agent host under /srv/testflinger/<agent_name>/... and under there, we have the git repo for testflinger-agent and snappy-device-agents. It would be nice to keep that mostly the same, although obviously the device agents dir is going to need to change. But I think I have a way of checking out the git repo so that only those two directories show up there, but we still get a shallow git tree, which can have the advantage of being able to update it later, pull other branches for testing, etc.  It's a small thing but could be useful. I'm not stuck on it either way though if you have a preference.

jocave:
> > Bit of bike-shedding - I feel like the `testflinger-` prefix on all the component sub directories is redundant. Should we remove it?
> 
> Just to clarify, do you mean just from the directory name? For instance, things like "testflinger-agent" or especially the user facing "testflinger-cli" would be a little confusing if you executed them as just "agent" or "cli". So I think internally, keeping testflinger as part of the name is ok. I'm ok with removing it from the names of the directories if you think that looks cleaner though. The directory structure would then look like this: /testflinger /agent /cli /device-connectors /server ...but the cmdline used to call them would still be the same as it was previously. The is one small advantage I can see to keeping it as-is. We know the agent charm will need some modification. Each agent currently gets a dir on the agent host under /srv/testflinger/<agent_name>/... and under there, we have the git repo for testflinger-agent and snappy-device-agents. It would be nice to keep that mostly the same, although obviously the device agents dir is going to need to change. But I think I have a way of checking out the git repo so that only those two directories show up there, but we still get a shallow git tree, which can have the advantage of being able to update it later, pull other branches for testing, etc. It's a small thing but could be useful. I'm not stuck on it either way though if you have a preference.

Yes, just meant the directory name

jocave (approved):
Thanks, looks good to me.

I think there is some cruft from old repositories (e.g. pmr-merge-hooks), but happy for the merge to proceed with what is in the source repos now and tidy these things up in later PRs.

PR #133, opened 2023-10-18
update tf charm to include metrics endpoint - Prometheus

hum4n0id:
## Description

Expose Prometheus metrics endpoint for COS lite

## Resolved issues

https://warthogs.atlassian.net/browse/CERTTF-203

## Tests

Tested via Juju / microK8s 

Added "api_port" to config.yaml to make the multiple references to this API port dynamic, helpful with add'l references for the metrics endpoint.


charm/config.yaml
```diff
@@ -7,3 +7,7 @@ options:
     default: 10
     description: Number of seconds to wait for keepalive connections
     type: int
+  api_port:
```
plars, line 10: Remove this section, we don't need api_port in the config, and it was actually recommended that we do not do this. Hardcoding things like this may seem unusual, but there's really no reason for it to change, and removing the option reduces the chances for config mismatches.

charm/src/charm.py
```diff
@@ -53,6 +54,19 @@ def __init__(self, *args):
             database_name="testflinger_db",
         )
 
+        self._prometheus_scraping = MetricsEndpointProvider(
+            self,
+            relation_name="metrics-endpoint",
+            jobs=[
+                {
+                    "static_configs": [
+                        {"targets": [f"*:{self.config['api_port']}"]}
```
plars, line 63: make sure to also remove the config item here

charm/src/charm.py
```diff
@@ -80,7 +94,7 @@ def _require_nginx_route(self):
             charm=self,
             service_hostname=self.config["external_hostname"],
             service_name=self.app.name,
-            service_port=5000,
+            service_port=self.config["api_port"],
```
plars, line 97: ...and here

charm/src/charm.py
```diff
@@ -162,7 +176,7 @@ def _pebble_layer(self):
                 "--keep-alive",
                 keepalive,
                 "--bind",
-                "0.0.0.0:5000",
+                f"0.0.0.0:{self.config['api_port']}",
```
plars, line 179: ...and here

plars (changes requested):
You'll see that I also pushed a few changes in here to make the tests pass again, and add the missing charm libraries.  There's one workflow still failing, but that one fails because you pushed this to your repo instead of the canonical one. Just make sure to push it to the canonical one next time

charm/config.yaml
hum4n0id: No problem

PR #135, opened 2023-10-19
update tf charm with prometheus scrap / metrics endpoint

hum4n0id:
## Description

Expose Prometheus metrics endpoint for COS lite

## Resolved issues

https://warthogs.atlassian.net/browse/CERTTF-203


charm/config.yaml
```diff
@@ -7,3 +7,7 @@ options:
     default: 10
     description: Number of seconds to wait for keepalive connections
     type: int
+  api_port:
```
plars, line 10: As in the previous PR, take this out - we don't need a config option for it because there's no reason for it to ever change.

charm/src/charm.py
```diff
@@ -53,6 +54,13 @@ def __init__(self, *args):
             database_name="testflinger_db",
         )
 
+        self._prometheus_scraping = MetricsEndpointProvider(
+            self,
+            relation_name="metrics-endpoint",
+            jobs=[{"static_configs": [{"targets": [f"*:5000"]}]}],
```
plars, line 60: I think this is what causes the lint failure. You have f"*:5000" but there's no placeholder for it.

plars (changes requested):
I added a few more comments, but be aware that this may need some rebasing or merging to make it work with the new monorepo layout.

PR #136, opened 2023-10-20
Post monorepo cleanup

plars:
This is for small cleanups that we need after moving to the monorepo. Larger changes should go in a separate PR


omar-selo (approved):

jocave (approved):
These changes look good.

Other things that could or should be done:

- top-level readme giving just intro to project, point to read-the-docs for rest
- sub-project readmes could also just link to the info in read-the-docs if all info has been transferred
- possibly consolidate the .gitignores?

plars:
@jocave I added something for 1 and 3 - I think 2 should be coordinated more with the doc efforts right now? See what you think.

jocave (approved):
Looks much cleaner, thanks

PR #137, opened 2023-10-21
Add state check to check_device_booted

kiya956:
impact:
    snappy-device-agents/src/testflinger_device_connectors/devices/oemrecovery/oemrecovery.py

description:
    currently check_device_booted not really check device current state
    and we found it could end up login into install mode state or run mode but before device ready.
    For making sure everything is ready before we start testing procedure, we add state checking.

test:
    test on x8high-pdk, x8high-jace9000, x8med-pdk, x6med-pdk pass

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/oemrecovery/oemrecovery.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/oemrecovery/oemrecovery.py
@@ -89,6 +89,72 @@ class OemRecovery:
         self._run_cmd_list(recovery_cmds)
         self.check_device_booted()
 
+    def check_generic_classic(self):
+        try:
+            test_username = self.job_data.get("test_data", {}).get(
+                "test_username", "ubuntu"
+            )
+            test_password = self.job_data.get("test_data", {}).get(
+                "test_password", "ubuntu"
+            )
+        except AttributeError:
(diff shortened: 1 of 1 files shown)


plars (commented):
Excellent, it looks like you moved it over from the old repo very nicely. Just one question - do you know if this was tested on any other existing oemrecover devices? or just the new ones that you are trying to use it with, and if so, which ones?

kiya956:
> Excellent, it looks like you moved it over from the old repo very nicely. Just one question - do you know if this was tested on any other existing oemrecover devices? or just the new ones that you are trying to use it with, and if so, which ones?

I only tested on x8high-pdk, x8high-jace9000 (Post GM, currently in cert Lab), x8med-pdk, X6med-pdk(still in pr-GM). But I didn't test on other devices in cert Lab.

plars (approved):
I found some other oemrecovery devices to test it on. It worked fine on the one with an ubuntu core image. I could only find one other device with a classic image but I think it was misconfigured and shouldn't have been oemrecovery. It didn't work at all, but it doesn't seem to be related to your change here. The other jobs actually using this are skipping provisioning, so I don't think we actually have any other classic oemrecovery images.
So with your testing and the extra that I did, +1

kiya956:
Paul,  Thank you so much.

PR #138, opened 2023-10-22
Fix oci image builds due to Dockerfile being in a subdir now

plars:
OCI images weren't working because we had "file: Dockerfile". This would normally be ok, except the context was set to ./server, so the file specified should have been {context}/Dockerfile. Turns out this is the default though, so all we need to do is remove it.


plars:
I tested this by adding the branch temporarily to the workflow file so that it got tested when I pushed it. The results are [here](https://github.com/canonical/testflinger/actions/runs/6600868488/job/17931328886?pr=138)

omar-selo (approved):

PR #139, opened 2023-10-23
Merge agent charms

plars:
This merges the existing testflinger agent charms from launchpad into the monorepo. I see this as a first step, and while they *will* work for now, they will still pull against the old github repos for now. I'll need to rework some things to make them properly work with the sources as we have them structured in the monorepo now. I'll do that in a separate PR later so that it doesn't have so much noise from merging all this git history from another repo and that will make it easier to review.


omar-selo (approved):
I see that some checks failed. But my understanding is that this PR is just to move code here, and later PRs will fix things

plars:
ok, fixed the complaints about the source layout without resorting to anything as invasive as I thought.

PR #143, opened 2023-10-26
Doc: add job definition schema and how-to guides

tang-mm:
## Description

Added a reference doc describing the job schema and a couple of how-to guides, including:
  - Install Testflinger CLI
  - Connect to a different server
  - Submit a job (with link to the job schema)
  - Cancel a job

## Resolved issues

Resolves [CERTTF-96](https://warthogs.atlassian.net/browse/CERTTF-96)

## Documentation

Preview version on Readthedocs:
- [Reference > Test job schema](https://canonical-testflinger.readthedocs-hosted.com/en/doc-preview/reference/job-schema)
- [How-to](https://canonical-testflinger.readthedocs-hosted.com/en/doc-preview/how-to)

## Tests

To check the documentation on a local sphinx site:
```
$ cd testflinger/docs
$ make install
$ make run     # Serve docs on http://127.0.0.1:8000/reference/test-phases/
```

[CERTTF-96]: https://warthogs.atlassian.net/browse/CERTTF-96?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/reference/job-schema.rst
+++ b/docs/reference/job-schema.rst
@@ -1,0 +1,84 @@
+Test job schema
+=================
+
+Test jobs can be defined in either YAML or JSON format.
+
+The following table lists the key elements that a job definition file should contain.
+
+.. list-table:: Test job schema
+  :header-rows: 1
+
+  * - Field
+    - Type 
+    - Description
+  * - ``job_queue``
+    - String
+    - Name of the job queue to which you want to submit the job. This field is mandatory.
+  * - ``global_timeout``
+    - integer
+    - | (Optional) Maximum time (in seconds) a job is allowed to run. If the global timeout is reached while a job is still running, Testflinger will close the job immediately. By default, the timeout for each Testflinger agent is 4 hours. 
+      | If your test job will presumably take more than 4 hours to complete, you can set a larger number in the ``global_timeout`` field in the test job. However, if a timeout is specified on a device, the value for this parameter will be overridden by the per-device configurations.
+  * - ``output_timeout``
+    - integer
+    - (Optional) Maximum time (in seconds) Testflinger should wait in the ``test`` phase to get output from the job. If the timeout is reached before the test phase has any output, Testflinger will cancel the job. This value should be smaller than, or equal to, the ``global_timeout``.  The default value is 15 minutes. 
+  * - ``allocation_timeout``
+    - integer
+    - (Optional) Maximum time (in seconds) Testflinger should wait in the ``allocate`` phase for multi-device jobs to reach the ``allocated`` state. If the timeout is reached before all devices are allocated, Testflinger will cancel the job. The default timeout to allocate all devices is 2 hours.
+  * - ``<phase>_data``
+    - dictionary
+    - | (Optional) Sections that define the configurations and commands about how the job should be executed in each of the test phases. If any phase encounters a non-zero exit code, no further phases will be executed.
+      | Supported test phases include: 
+      |   - provision
+      |   - firmware_update
+      |   - test
+      |   - allocate
+      |   - reserve 
+      | For detailed information about how to define the data to include in each test phase, see :doc:`test-phases`. 
+
+Environment variables
+----------------------------
+
+The following environment variables are supported in the test environment on the HOST system (not on the device under test). You can refer to these variables when defining a job:
+
+.. list-table:: Environment variables on the HOST system
+  :header-rows: 1
+
+  * - Name
+    - Description
+  * - ``WPA_BG_SSID``
+    -	SSID for BG WiFi access point
+  * - ``WPA_BG_PSK``
+    - WPA Key for BG WiFi access point
+  * - ``WPA_N_SSID``
+    - SSID for N WiFi access point
+  * - ``WPA_N_PSK``
+    - WPA Key for N WiFi access point
+  * - ``WPA_AC_SSID``
+    - SSID for AC WiFi access point
+  * - ``WPA_AC_PSK``
+    - WPA Key for AC WiFi access point
+  * - ``OPEN_BG_SSID``
+    - SSID for OPEN BG WiFi access point
+  * - ``OPEN_N_SSID``
+    - SSID for OPEN N WiFi access point
+  * - ``OPEN_AC_SSID``
+    - SSID for OPEN AC WiFi access point
+  * - ``DEVICE_IP``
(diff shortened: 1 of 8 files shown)


docs/reference/job-schema.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Test jobs can be defined in either YAML or JSON format.
+
+The following table lists the key elements that a job definition file should contain.
+
+.. list-table:: Test job schema
+  :header-rows: 1
+
+  * - Field
+    - Type
```
plars, line 12: Can we add a `- Default` column here also and put in the defaults there?

docs/reference/job-schema.rst
```diff
@@ -0,0 +1,84 @@
...
+  * - Field
+    - Type 
+    - Description
+  * - ``job_queue``
+    - String
+    - Name of the job queue to which you want to submit the job. This field is mandatory.
+  * - ``global_timeout``
+    - integer
+    - | (Optional) Maximum time (in seconds) a job is allowed to run. If the global timeout is reached while a job is still running, Testflinger will close the job immediately. By default, the timeout for each Testflinger agent is 4 hours. 
+      | If your test job will presumably take more than 4 hours to complete, you can set a larger number in the ``global_timeout`` field in the test job. However, if a timeout is specified on a device, the value for this parameter will be overridden by the per-device configurations.
```
plars, line 20: I would shorten this a bit if possible. Maybe something like...
"You can set a larger number in the ``global_timeout`` field, but per-device configurations may restrict it to something less than you specify."

docs/reference/job-schema.rst
```diff
@@ -0,0 +1,84 @@
...
+      |   - firmware_update
+      |   - test
+      |   - allocate
+      |   - reserve 
+      | For detailed information about how to define the data to include in each test phase, see :doc:`test-phases`. 
+
+Environment variables
+----------------------------
+
+The following environment variables are supported in the test environment on the HOST system (not on the device under test). You can refer to these variables when defining a job:
```
plars, line 41: This is not quite right - and it's very specific to our labs. Even then, I don't think the server cert lab has this wifi data in their agents, just DEVICE_IP. There's nothing in testflinger that requires or restricts any values from the environment settings.

I think we could put something like this in the internal documentation, but  even then, we would have to denote which labs it even applies to.

plars (changes requested):
Lots of good stuff here, thanks! A few comments below.

docs/reference/job-schema.rst
```diff
@@ -0,0 +1,84 @@
...
+  * - ``OPEN_AC_SSID``
+    - SSID for OPEN AC WiFi access point
+  * - ``DEVICE_IP``
+    - IP address of the device under test (used for SSH)
+
+
+Example job in YAML
+----------------------------
+
+The following example YAML file defines a job that provisions the Ubuntu 22.04-jammy system on the device, and print the distribution-specific information on the provisioned device:
```
kevinyehk, line 73: do we need to describe that if use "distro: jammy" means Ubuntu 22.04 server image not desktop.

kevinyehk (commented):
Just have a minor suggestion.

docs/reference/job-schema.rst
```diff
@@ -0,0 +1,61 @@
...
+
+The following example YAML file defines a job that retrieves the Ubuntu 22.04-jammy  system image from the given URL, provisions the image on the device at IP address ``DEVICE_IP``, and prints the distribution-specific information on the provisioned device:
+
+.. code-block:: yaml
+
+  job-queue: example-queue
+  global-timeout: 28800
+  output-timeout: 3600
+  provision_data:
+    url: https://cdimage.ubuntu.com/ubuntu/releases/jammy/release/example.img.xz
```
plars, line 58: iirc, this example use to show the use of a maas system. If you are are going to use a `url: ...` tag in the provision data, then I think we should be really careful to make it clear that this is a muxpi provisioned device like rpi or risc-v. For instance, maybe the queue name could be `rpi4b` and the url could actually point to the arm64-raspi image for jammy?

If you are trying to show a maas image, then the distro tag needs to be used. MaaS provisioned devices cannot specify a url, so we should not give the impression that this will work for all types of systems.

docs/reference/job-schema.rst
tang-mm: removed this section in appended commit

docs/reference/job-schema.rst
plars: I think that might be useful, yes. Maybe one with `job_queue: example-maas-queue` and another with `job_queue: rpi4b` or something? That might make it more obvious. You could even use something like a comment in the yaml `# this is a jammy server image` or something to make it clear and hopefully handle Kevin's concern.

docs/reference/job-schema.rst
```diff
@@ -44,18 +44,30 @@ The following table lists the key elements that a job definition file should con
...
+Example jobs in YAML
 ----------------------------
 
-The following example YAML file defines a job that retrieves the Ubuntu 22.04-jammy  system image from the given URL, provisions the image on the device at IP address ``DEVICE_IP``, and prints the distribution-specific information on the provisioned device:
+The following example YAML file defines a job that provisions the Ubuntu Core 22 system on a Raspberry Pi 4 device. The job retrieves the image from the given URL, provisions the image on the device at IP address ``DEVICE_IP``, and prints the distribution-specific information on the provisioned device:
 
 .. code-block:: yaml
 
-  job-queue: example-queue
+  job_queue: raspi4b
```
plars, line 54: Could you actually change this job_queue to "rpi4b"? Otherwise, this looks good to me!

plars (changes requested):

docs/reference/job-schema.rst
plars: Oh... and I just saw your updates after typing that ^
So I think you anticipated almost everything I was going to say, only one minor change to that rpi4b job queue name, but otherwise looks good to me!

docs/reference/job-schema.rst
tang-mm: Updated. I had to rebase all commits with a signature and force-pushed. Could you take a look again? Thanks!

plars (approved):
+1

PR #148, opened 2023-10-26
Send all queues in agent data

plars:
## Description
- Send all known queues for the agent instead of just the advertised ones to the server for the initial_agent_data
- Modify the agent_detail page so that it displays all of these queues for each agent, along with a link to the queue details page
- **BONUS** small change to speed up test_agent.py execution significantly (from 161s to 16s on my system)

## Resolved issues
Fixes CERTTF-141

## Documentation
N/A

## Tests
Unit test added for this method, and I also set it up locally and connected an agent to it so that I could see the change to the html. Here's an example of what it looks like:
image

### DIFF
--- a/agent/testflinger_agent/tests/test_agent.py
+++ b/agent/testflinger_agent/tests/test_agent.py
@@ -23,6 +23,7 @@ class TestClient:
             "polling_interval": "2",
             "server_address": "127.0.0.1:8000",
             "job_queues": ["test"],
+            "location": "nowhere",
             "execution_basedir": self.tmpdir,
             "logging_basedir": self.tmpdir,
             "results_basedir": os.path.join(self.tmpdir, "results"),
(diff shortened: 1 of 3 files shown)


agent/testflinger_agent/tests/test_agent.py
```diff
@@ -203,3 +206,16 @@ def test_recovery_failed(self, agent, requests_mock):
...
+        with patch.object(
+            testflinger_agent.client.TestflingerClient, "post_agent_data"
+        ) as mock_post_agent_data:
+            agent._post_initial_agent_data()
+            mock_post_agent_data.assert_called_with(
+                {
+                    "queues": self.config["job_queues"],
+                    "location": self.config["location"],
+                }
+            )
```
omar-selo, line 221: are you using `requests_mock` fixture here?

server/src/templates/agent_detail.html
```diff
@@ -2,17 +2,27 @@
...
-    <br>
+    </p>
+    <p>
+    Queues:
+    <ul>
+        {% for queue in agent.queues %}
+        <li><a href="/queues/{{ queue }}">{{ queue }}</a></li>
+        {% endfor %}
+    </ul>
+    </p>
```
omar-selo, line 18: I'm not aware what are advertised queues and how they are different from rest of queues. But would it be useful to show them separately or mark the queues that are advertised?

agent/testflinger_agent/tests/test_agent.py
plars: good catch! I've removed it now

server/src/templates/agent_detail.html
plars: Not really - in this case, we probably always should have been using all queues rather than just the advertised ones. Also, I expect advertised queues will work very differently in the future.
What this will allow is looking at an agent page and seeing all the possible queues you can use to submit a job there, which I think will be quite useful for end-users!

nadzyah (approved):
The changes look clear to me. I don't have any questions about them

PR #150, opened 2023-10-27
Doc: add getting-started tutorial

tang-mm:
## Description

Added a basic tutorial for first-time end users, covering:
- installing the CLI via Snap
- configure the default server
- creating a first job
- submit job
- check job status
- check job output

(This tutorial has some overlap with the how-to guides in #143 , with fewer options and more explanation for new users.) 

## Resolved issues

Resolves [CERTTF-222](https://warthogs.atlassian.net/browse/CERTTF-222)

## Documentation

Preview version on Readthedocs:

https://canonical-testflinger.readthedocs-hosted.com/en/doc-preview/tutorial

## Tests

To check the documentation on a local sphinx site:
```
$ cd testflinger/docs
$ make install
$ make run     # Serve docs on http://127.0.0.1:8000/reference/test-phases/
```

[CERTTF-222]: https://warthogs.atlassian.net/browse/CERTTF-222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/tutorial/index.rst
+++ b/docs/tutorial/index.rst
@@ -32,7 +32,7 @@ The most convenient way to install the CLI is through the *Snap Store*. Open a t
 
 Once the installation is finished, you can execute the ``testflinger-cli`` command to check the installation. For example, you can run with the ``--help`` or ``-h`` option to display the CLI help page, which is useful anytime when you need a reference about how to use the tool:
 
-.. code-block:: shell
+.. code-block::
 
   $ testflinger-cli --help
 
@@ -110,7 +110,7 @@ Run the following command in the terminal to retrieve the available job queues t
 
 If the connection is successful, a list of job queues is returned with their queue names and short descriptions:
 
-.. code-block:: shell
+.. code-block::
 
   Advertised queues on this server:
     example-queue-1 - for testing device model-1
--- a/docs/tutorial/index.rst
+++ b/docs/tutorial/index.rst
@@ -110,7 +110,7 @@ Run the following command in the terminal to retrieve the available job queues t
 
 If the connection is successful, a list of job queues is returned with their queue names and short descriptions:
 
-.. code-block::
+.. code-block:: text
 
   Advertised queues on this server:
     example-queue-1 - for testing device model-1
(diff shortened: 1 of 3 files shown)


docs/tutorial/index.rst
```diff
@@ -1,20 +1,228 @@
...
+
+The most convenient way to install the CLI is through the *Snap Store*. Open a terminal and run the following command:
+
+.. code-block:: shell
+  
+  $ sudo snap install testflinger-cli
+
+Once the installation is finished, you can execute the ``testflinger-cli`` command to check the installation. For example, you can run with the ``--help`` or ``-h`` option to display the CLI help page, which is useful anytime when you need a reference about how to use the tool:
+
+.. code-block:: shell
```
plars, line 35: Can we remove the syntax highlighting from this section? It's highlighting some strange words in the preview build.

docs/tutorial/index.rst
```diff
@@ -1,20 +1,228 @@
...
+
+  $ testflinger-cli list-queues
+
+.. note::
+  
+  If you want to temporarily use another server, add ``--server`` argument and the server URI in the command.
+
+If the connection is successful, a list of job queues is returned with their queue names and short descriptions:
+
+.. code-block:: shell
```
plars, line 113: Same here with the syntax highlighting. "for" gets highlighted below.

plars (changes requested):
Those preview links are super-helpful, thanks for including those!
This looks good other than some strange syntax highlighting. I would guess it could either be turned off for those code blocks or changed to some other type of highlighting.

plars (approved):
It looks like the earlier section for testflinger-cli looks fine for me now since you removed "bash", but the later one still had weird highlighting.
image

I found that if I use "text" instead, then it renders fine for me now. I hope you don't mind, I just pushed the change, since it was trivial. And I'm +1 on it now. But I'll let you make the final call :)

kevinyehk (approved):
LGTM +1

PR #151, opened 2023-10-27
Build cli snap workflow

plars:
## Description

Use a github workflow to build and publish the cli snap

## Resolved issues
Previously, I did this through the snapcraft.io builders and set them up to watch the testflinger-cli repo and build automatically. However, those do not work with a monorepo because they don't support building the snap if it's in a directory. So we need to move this to use the gh workflow instead.

## Documentation
N/A

## Tests

I'm testing this by temporarily adding the proposed branch to the list of ones to trigger it
**update**: tested here: https://github.com/canonical/testflinger/actions/runs/6670022069/job/18129054928?pr=151
I also installed the resulting snap and tested it locally, works great :)

### DIFF
--- a/.github/workflows/cli-publish-snap.yml
+++ b/.github/workflows/cli-publish-snap.yml
@@ -1,0 +1,22 @@
+name: Publish Testflinger CLI snap to edge
+on:
+  push:
+    branches: ["main"]
+    paths:
+      - cli/**
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: snapcore/action-build@v1
+      id: build
+      with:
+        path: cli
+    - uses: snapcore/action-publish@v1
+      env:
+        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAP_STORE_LOGIN }}
+      with:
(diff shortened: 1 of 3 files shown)


.github/workflows/cli-publish-snap.yml
```diff
@@ -0,0 +1,22 @@
...
+    - uses: snapcore/action-build@v1
+      id: build
+      with:
+        path: cli
+    - uses: snapcore/action-publish@v1
+      env:
+        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAP_STORE_LOGIN }}
+      with:
+        snap: ${{ steps.build.outputs.snap }}
+        release: edge
```
mz2, line 22: I think `edge` makes sense for now. Please describe this scheme however in the project README.

I've filed a follow-up for introducing semantically versioning and a stable channel: https://github.com/canonical/testflinger/issues/154

mz2 (dismissed):
Nice! I have filled a follow-up issue for introducing stable builds.

.github/workflows/cli-publish-snap.yml
plars: Updated the README

mz2 (approved):

PR #153, opened 2023-10-27
Enable self-hosted workflow runners

plars:
## Description
This enables the self-hosted runners for all of our existing workflows

## Resolved issues
Resolves long wait times which hit us frequently

## Documentation

I added a CONTRIBUTING.md file in the root of the repo with some useful information as well as a section about dealing with signed commits shamelessly stolen from the excellent one in checkbox.

## Tests
I tested this in a different branch to ensure everything works. You should still be able to see the results [here](https://github.com/canonical/testflinger/pull/152/checks)

### DIFF
--- a/.github/workflows/agent-tox.yml
+++ b/.github/workflows/agent-tox.yml
@@ -15,7 +15,7 @@ jobs:
     defaults:
       run:
         working-directory: agent
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, linux, X64]
     strategy:
       matrix:
         python: ["3.8", "3.10"]
--- a/.github/workflows/cli-tox.yml
+++ b/.github/workflows/cli-tox.yml
@@ -2,7 +2,7 @@ name: "[cli] Run unit tests"
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, try-self-hosted ]
     paths:
       - cli/**
   pull_request:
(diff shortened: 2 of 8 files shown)


mz2 (approved):
LGTM! The contribution guide looks somehow familiar :thinking:

kissiel (commented):
Worth adding is the specification of which size runner to use. We use "large" for toxxing and lightweight stuff (2cores, 8GB mem).

I've also noticed that the tests are being run when stuff gets pushed to main. Isn't it too late?

mz2:
> Worth adding is the specification of which size runner to use

As per https://github.com/canonical/test_observer/pull/55#issuecomment-1782944806 I actually think we shouldn't specify this unless we know there's some particular resource requirement (for example the checkbox builds waiting on remote builds similarly don't actually need any particularly high resource availability.

Let's merge in this shape, adjust in follow-up if build time or stability needs attention.

mz2:
> I've also noticed that the tests are being run when stuff gets pushed to main. Isn't it too late?

That's just github's diff presentation hiding the pull request build triggers, all looks to be in order. Merging ->

kissiel:
> > Worth adding is the specification of which size runner to use
> 
> As per [canonical/test_observer#55 (comment)](https://github.com/canonical/test_observer/pull/55#issuecomment-1782944806) I actually think we shouldn't specify this unless we know there's some particular resource requirement (for example the checkbox builds waiting on remote builds similarly don't actually need any particularly high resource availability).
> 
> Let's merge in this shape, adjust in follow-up if build time or stability needs attention.

ISD029 states that

>Please use the “large” label by default

So I was just parroting it :)

PR #155, opened 2023-11-02
Update agent charms

plars:
## Description

Update the testflinger-agent-charm to use the monorepo

## Resolved issues
There are some minor layout changes necessitated by moving to the monorepo. It is still the case that each agent gets installed under /srv/testflinger-agent/<agent_name>.
Previously, there was a `testflinger-agent` and `snappy-device-agents` directory under there. Now that we have a monorepo, there will just be a `testflinger` directory there with a sparse/shallow checkout that only includes the `agent` and `device-connectors` directories.
Additionally, the config files have now moved to the root agent-dir, so that they are easier to find and manage.

## Documentation

No real changes from a doc perspective. Installation and deployment are still the same.

## Tests
To test this, I installed the old charm locally and pushed the new one to the edge channel, then refreshed the charm to use the edge one instead. The upgrades went perfectly and the new version of the agent was running after the upgrade.

### DIFF
--- a/agent/charms/testflinger-agent-charm/src/charm.py
+++ b/agent/charms/testflinger-agent-charm/src/charm.py
@@ -41,11 +41,11 @@ class TestflingerAgentCharm(CharmBase):
 
     def __init__(self, *args):
         super().__init__(*args)
-        self.framework.observe(self.on.install, self._on_install)
-        self.framework.observe(self.on.config_changed, self._on_config_changed)
-        self.framework.observe(self.on.start, self._on_start)
-        self.framework.observe(self.on.remove, self._on_remove)
-        self.framework.observe(self.on.update_action, self._on_update_action)
+        self.framework.observe(self.on.install, self.on_install)
+        self.framework.observe(self.on.config_changed, self.on_config_changed)
+        self.framework.observe(self.on.start, self.on_start)
+        self.framework.observe(self.on.remove, self.on_remove)
+        self.framework.observe(self.on.update_action, self.on_update_action)
         self._stored.set_default(
             testflinger_agent_repo="",
             testflinger_agent_branch="",
@@ -59,7 +59,7 @@ class TestflingerAgentCharm(CharmBase):
             venv_path=f"/srv/testflinger-agent/{self.app.name}/env",
         )
 
-    def _on_install(self, _):
+    def on_install(self, _):
         """Install hook"""
         self.unit.status = MaintenanceStatus("Installing dependencies")
         # Ensure we have a fresh agent dir to start with
@@ -68,7 +68,7 @@ class TestflingerAgentCharm(CharmBase):
         os.makedirs("/home/ubuntu/testflinger", exist_ok=True)
         shutil.chown("/home/ubuntu/testflinger", "ubuntu", "ubuntu")
 
-        self._install_apt_packages(
+        self.install_apt_packages(
             [
                 "python3-pip",
                 "python3-virtualenv",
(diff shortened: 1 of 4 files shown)


agent/charms/testflinger-agent-host-charm/src/charm.py
```diff
@@ -30,29 +30,29 @@ class TestflingerAgentHostCharm(CharmBase):
 
     def __init__(self, *args):
         super().__init__(*args)
-        self.framework.observe(self.on.install, self._on_install)
-        self.framework.observe(self.on.start, self._on_start)
-        self.framework.observe(self.on.config_changed, self._on_config_changed)
+        self.framework.observe(self.on.install, self.on_install)
+        self.framework.observe(self.on.start, self.on_start)
+        self.framework.observe(self.on.config_changed, self.on_config_changed)
```
omar-selo, line 35: Just out of curiosity, is there a particular reason for avoiding methods that start with `_`

agent/charms/testflinger-agent-charm/src/charm.py
```diff
@@ -93,106 +91,114 @@ def _run_with_logged_errors(self, cmd):
...
-            multi_options=[f"-b {self._stored.testflinger_agent_branch}"],
+            url=self._stored.testflinger_repo,
+            to_path=repo_path,
+            no_checkout=True,
+            depth=1,
         )
-        self._run_with_logged_errors(
-            [f"{self._stored.venv_path}/bin/pip3", "install", "-I",
-             tf_agent_dir]
+        repo = Repo(repo_path)
```
omar-selo, line 149: very small optional note, but perhaps you could just do `repo = Repo.clone_from`

agent/charms/testflinger-agent-charm/src/charm.py
```diff
@@ -218,14 +224,15 @@ def _render_systemd_unit(self):
...
-    def _on_config_changed(self, _):
+    def on_config_changed(self, _):
         self.unit.status = MaintenanceStatus("Handling config_changed hook")
-        self._check_update_repos_needed()
-        self._write_config_files()
-        self._signal_restart_agent()
+        self.check_update_repos_needed()
+        self.write_config_files()
+        self.render_systemd_unit()
+        self.signal_restart_agent()
```
omar-selo, line 232: Just checking that the addition of `render_systemd_unit` here is intentional

omar-selo (dismissed):
Looks good

agent/charms/testflinger-agent-host-charm/src/charm.py
plars: mostly just opinion that I hate reading it  like that and it's kind of pointless since this isn't a module that other things are importing where we're trying to give hints about what should be called publicly vs what should be only used internally. I'm not stuck on that opinion, just what I was thinking when I started looking at it again.

agent/charms/testflinger-agent-charm/src/charm.py
plars: yes! There was a small change needed to the systemd unit template and we need to make sure it gets installed to work properly.  Most of the time, it would probably be overkill to install it on every config change, but it doesn't take long and has no effect if nothing changed.

agent/charms/testflinger-agent-charm/src/charm.py
plars: Done, thanks!

plars:
@omar-selo or others... one other small change I just added was to explicitly *skip* removing the old "testflinger-agent" and "snappy-device-agents" dirs for now. We will add those back after all agents have been deployed with this version. The reason we need to do it in two steps, is because of the possibility that a job could be running when we update the agent. We want to give it an opportunity to finish running before we remove the old device agents, and I think this would be a safer way to do that.

omar-selo (approved):
Looks good

PR #158, opened 2023-11-08
Use aproxy for the oci-publish step for self-hosted runners

plars:
self-hosted runners have very restrictive rules for network access. As a result, pulling docker base images to build and publish OCI images requires a proxy. IS devops has been working on this solution called aproxy, which will eventually be installed by default on the self-hosted runners. So eventually we can just remove this section from the workflow, but for now it's needed in order to get OCI image publishing working again.

## Testing
I ran the action against this branch and it passed here: https://github.com/canonical/testflinger/actions/runs/6795492316/job/18473690576

### DIFF
--- a/.github/workflows/server-publish-oci-image.yml
+++ b/.github/workflows/server-publish-oci-image.yml
@@ -27,6 +27,31 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


omar-selo (approved):
Nice find. Also glad IS is going to configure this by default in self hosted runners.

PR #162, opened 2023-11-21
Make the UI more closely follow vanilla

jocave:
## Description

Match some of other applications with a UI that more closely follows the vanilla guidelines. Use "standard" nav bar and components.

Although we have plans for possible integrating this with other UI (or implementing in technologies) this looks a bit more in-keeping for now.

## Resolved issues

No issues solved - just looks a bit nicer!

## Documentation

No documentation changes, just html template modifications

## Tests

Tested in the docker-compose development environment locally

### DIFF
--- a/server/src/templates/job_detail.html
+++ b/server/src/templates/job_detail.html
@@ -63,31 +63,78 @@
 {% endif %}
 
 <h2 class="p-muted-heading">Provision Phase</h2>
-provision_status: {{ job.result_data.provision_status }}<br>
-provision_output:
-<div class="center scrollable">
-    <span style="white-space: pre-line">
-        {{ job.result_data.provision_output }}
+<span class="p-chip">
+    {% if job.result_data.provision_status %}
+    <span class="p-chip__lead">
+        Success
     </span>
+    {% else %}
+    <span class="p-chip__lead">
+        Error
+    </span>
+    {% endif %}
+</span>
+<div class="p-code-snippet">
+    <div class="p-code-snippet__header">
+        <h5 class="p-code-snippet__title">Provision Output</h5>
+    </div>
+    <div class="scrollable">
+        <pre class="p-code-snippet__block">
+        <code>
+{{ job.result_data.provision_output }}
+        </code>
+    </pre>
+    </div>
 </div>
 
 <h2 class="p-muted-heading">Firmware Update Phase</h2>
-firmware_update_status: {{ job.result_data.firmware_update_status }}<br>
-firmware_update_output:
-<div class="center scrollable">
-    <span style="white-space: pre-line">
-        {{ job.result_data.firmware_update_output }}
+<span class="p-chip">
+    {% if job.result_data.provision_status %}
+    <span class="p-chip__lead">
+        Success
     </span>
+    {% else %}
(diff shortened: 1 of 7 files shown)


jocave:
needs a bit more working out how to put the main content in an appropriate column with some space on the borders..

plars (changes requested):
I like the updated logo and everything, but it seems to hard shift everything to the center. Is there a way to prevent that?
image

jocave:
> I like the updated logo and everything, but it seems to hard shift everything to the center. Is there a way to prevent that? image

I'm sure there must be :)

It would be handy to have a smallish dataset to populate the mongodb docker instance with for local testing and possibly integration testing - any chance could get such a thing?

plars (dismissed):
Looks much better now! +1

plars (commented):
One odd thing I noticed is this:
image

I wonder if there's a way we can fix the inconsistent wrapping? maybe the columns can be resized somehow? It even happens when I go full-screen. It didn't work like that on the previous one.

jocave:
> One odd thing I noticed is this: image
> 
> I wonder if there's a way we can fix the inconsistent wrapping? maybe the columns can be resized somehow? It even happens when I go full-screen. It didn't work like that on the previous one.



Could do something like this to reduce the size of other columns

Screenshot from 2024-01-24 17-52-41

plars (approved):
Looks much better now on that field, +1

server/src/templates/job_detail.html
```diff
@@ -1,81 +1,140 @@
...
-    provision_status: {{ job.result_data.provision_status }}<br>
-    provision_output:
-    <div class="center scrollable">
-        <span style="white-space: pre-line">
-            {{ job.result_data.provision_output }}
-        </span>
+<h2 class="p-muted-heading">Provision Phase</h2>
+<span class="p-chip">
+    {% if job.result_data.provision_status %}
+    <span class="p-chip__lead">
```
jocave, line 68: @plars know I review my own work I think I looked at converting these statuses to "chips" - but not sure it can really be called finished. The statuses are I think text descriptions. Maybe this part should be backed out.

plars: no, the statuses are basically like an exit code for that stage. So 0 for pass or anything else for fail. However, especially with things like the test phase, it's up to the user on the exit status.

So yeah, thanks for pointing this out. I didn't notice what this change was doing at first. I think this should be removed for now.

plars: @jocave I've updated this branch to remove the p-chip spans and replace them with another heading line (to keep it consistent looking) and the numerical exit code if it has it, otherwise it will put `[unknown]`. This is what it looks like now:
image

What do you think?

jocave: The text is great thanks for that. I think I wouldn't personally have it be a header, although I can see that the header that describes the output box does somewhat draw attention.

jocave: That's nice! The colour helps emphasise the outcome. I think the widgets I selected from the vanilla framework were being a bit stretched to be used as long console output holders anyway

plars (approved):

PR #168, opened 2023-12-01
Add an identifier field for associating devices based on CID

plars:
## Description
Add an identifier field that we can use for any generic serial number (like CID for us) that helps us match the agent up with other systems

## Resolved issues

Resolves CERTTF-245

## Documentation

Modified the readme as well as the reference section of the docs which describe these fields.

## Tests

Extended existing unit tests to ensure that this data is transmitted to the server if it exists.

### DIFF
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -37,7 +37,12 @@ class TestflingerAgent:
         self._post_advertised_images()
 
         queues = self.client.config.get("job_queues", [])
-        self.client.post_agent_data({"queues": queues, "location": location})
+        identifier = self.client.config.get("identifier")
+        agent_data = {"queues": queues, "location": location}
+        if identifier:
+            agent_data["identifier"] = identifier
+
+        self.client.post_agent_data(agent_data)
 
     def _post_advertised_queues(self):
         """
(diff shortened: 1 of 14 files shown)


agent/testflinger_agent/agent.py
```diff
@@ -37,7 +37,12 @@ def _post_initial_agent_data(self):
         self._post_advertised_images()
 
         queues = self.client.config.get("job_queues", [])
-        self.client.post_agent_data({"queues": queues, "location": location})
+        identifier = self.client.config.get("identifier")
+        agent_data = {"queues": queues, "location": location}
+        if identifier:
```
nadzyah, line 42: What do you think about using “is not None” here instead?

plars: Hmm, not sure. The other piece of this coming next is an API that allows pulling this data for all known agents so that c3 can import it all into a temporary db. My initial though on this was that we would not populate this field unless it was specified. Then that api could filter out ones that don't have an identifier. However, this means that the api that says "give me all the agent data" won't really give you *all* the agent data.

However, I think a case could be made to always populate this field (even though it's optional) and always transmit something, like an empty string or None if it doesn't exist. But then c3 would need to only consider agents that have something meaningful in this filed, and would have a lot of agent data that it can never match up against any other CID since it doesn't have anything in that field.

Do you have any opinion on this?

nadzyah: if we don't accept empty strings, I'm fine with this approach

nadzyah (approved):

PR #171, opened 2023-12-09
Add api to allow retrieving agent data

plars:
## Description

Add api to allow retrieving agent data

## Resolved issues

Resolves CERTTF-246

## Documentation

swagger docs and README updated

## Tests

Unit test added

### DIFF
--- a/server/src/api/schemas.py
+++ b/server/src/api/schemas.py
@@ -31,6 +31,15 @@ class AgentIn(Schema):
     log = fields.List(fields.String(), required=False)
 
 
+class AgentOut(Schema):
+    """Agent data input schema"""
+
+    state = fields.String(required=False)
+    queues = fields.List(fields.String(), required=False)
+    location = fields.String(required=False)
+    job_id = fields.String(required=False)
+
+
 class ActionIn(Schema):
     """Action data input schema"""
 
--- a/server/tests/test_v1.py
+++ b/server/tests/test_v1.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2016-2022 Canonical
+# Copyright (C) 2016-2023 Canonical
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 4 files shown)


server/tests/test_v1.py
```diff
@@ -401,3 +401,23 @@ def test_agents_post_bad(mongo_app):
...
+    """Test api to retrieve agent data"""
+    app, _ = mongo_app
+    agent_name = "agent1"
+    agent_data = {
+        "state": "provision",
+        "queues": ["q1", "q2"],
+        "location": "here",
+    }
+    output = app.post(f"/v1/agents/data/{agent_name}", json=agent_data)
+    assert 200 == output.status_code
```
omar-selo, line 416: I've typically shied away from setting up tests through other app code. So I'd create objects in db directly instead of calling post endpoints. As I felt that would be more of a unit test. Frankly, I'm not sure if there is really any benefit to that. From you're experience and knowledge, have I been wasting my time :sweat_smile: ?

server/src/api/schemas.py
```diff
@@ -31,6 +31,15 @@ class AgentIn(Schema):
     log = fields.List(fields.String(), required=False)
 
 
+class AgentOut(Schema):
+    """Agent data input schema"""
+
+    state = fields.String(required=False)
+    queues = fields.List(fields.String(), required=False)
+    location = fields.String(required=False)
+    job_id = fields.String(required=False)
```
omar-selo, line 40: does `required=False` mean that the returned json from this endpoint might not include this field at all?

omar-selo (commented):
I've noticed [CERTTF-246](https://warthogs.atlassian.net/browse/CERTTF-246) mentions a query parameter `?identifier=<cid>`. Is this no longer necessary or perhaps it's coming in a future PR?

plars:
> I've noticed [CERTTF-246](https://warthogs.atlassian.net/browse/CERTTF-246) mentions a query parameter `?identifier=<cid>`. Is this no longer necessary or perhaps it's coming in a future PR?

My understanding of the problem at the time we wrote that, was that they simply wanted to see if it was in testflinger or not on the c3 machine details page. After more discussion, it sounds like what they actually want is to extend the search API in a way that allows them to also query for systems that are in testflinger and retrieve the list of valid queues for that device as well. So eventually the solution will likely look more like what we discussed at the sprint - where c3 knows everything about every device, and testflinger generates configs for devices based off that data rather than through cc-lab-manager. But that will take quite a bit of time to implement. For now, this gives us a way that we can bulk-import the necessary data about agents so that c3 has more awareness and can hopefully include this in search results.

server/src/api/schemas.py
plars: it's possible that any of those fields might not be populated.  I'm not sure off the top of my head whether the GET would fail if it's not present, but at the very least, it communicates this possibility in the swagger docs.

server/tests/test_v1.py
plars: I don't have a strong opinion here, since we have other tests that test posting that data also. I feel like using the API is a little more realistic than direct manipulation of the db - when possible. But I've done both, as sometimes it's not easy to inject the data you actually want, because you need to simulate a very specific condition.

PR #172, opened 2023-12-12
Limit the charm libs check to server/charm/**

mz2:
## Description

Limits the charm libs check to `server/charm/**`

## Resolved issues

Just a small observation made when reviewing the fact that we have the action executed.

## Documentation

The workflow definition is the documentation.

## Tests

Fine to see if it just works.


plars (approved):
+1

PR #173, opened 2023-12-12
Fix: Change expected master boot string for netboot back

plars:
## Description

For netboot images (just plano gadgets use this), there's an initramfs with a small binary that let's us push around images, reformat the drive, etc. There's a string that it sends back if you hit it to confirm that we are in that image, and this device connector needs to expect that properly, even though it uses old names.

## Resolved issues

Right now, netboot provisioning won't work on a few devices because this is expecting the wrong thing back

## Tests

No changes, this is just changing the expected output back to how it previously was when it worked.


kevinyehk (approved):
LGTM +1

PR #175, opened 2023-12-13
Switch `testflinger-cli` to `core22` base

simondeziel:
Switching to `core22` will use `python3.10` instead of `python3.8` from `core20`.

### DIFF
--- a/cli/snapcraft.yaml
+++ b/cli/snapcraft.yaml
@@ -5,7 +5,7 @@ description: |
   server for submitting test jobs, checking status, getting results, and
   streaming output.
 confinement: strict
-base: core20
+base: core22
 adopt-info: testflinger-cli
 
 architectures:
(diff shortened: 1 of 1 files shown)


.github/workflows/cli-tox.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     runs-on: [self-hosted, linux, X64]
     strategy:
       matrix:
-        python: ["3.8", "3.10"]
```
plars, line 21: I'm fine with moving it to core22, but can we keep 3.8 in the matrix for unit tests for now? I think we might have some users that still need to run from source on older systems. Fortunately not so old that 3.8 has been a problem for them though.

plars (changes requested):

plars:
/canonical/self-hosted-runners/run-workflows 8269f5238b6bc1015434b1aa7d3ab58dffc7e913

plars (approved):

PR #177, opened 2023-12-15
Add add hp_oemscript device connector

plars:
## Description

This adds oemscript provisioning for hp devices explicitly. Functionally, there's no difference right now between this and the lenovo one since both require the ubr flag, but that might change in the future.

## Resolved issues
Allows hp oem systems to be configured for provisioning oem images.

## Documentation
Added a mention of it in the reference section of the docs like the other types have.

## Tests
This was tested a while back by Kevin, though I think at the time he was just using the lenovo one. Since all that's happening here are string changes, I don't suspect there should be a problem, but it might be useful to try it on one system first before rolling it out to all the others.


mz2 (approved):
LGTM!

PR #179, opened 2023-12-21
Add Dockerfile to create an OCI image of agent

hanhsuan:
## Description

Add Dockerfile to create an OCI image includes
1. testflinger-agent
2. testflinger-cli
3. device-connector

and could change the settings by environment variables while starting.


nancyc12:
/canonical/self-hosted-runners/run-workflows [802441a](https://github.com/canonical/testflinger/pull/179/commits/802441a2403d7e2b770cffc974929f150c14b900)

plars:
Can you tell me more about what this is for? I'd be happy to see something like this as a mechanism for integration testing, but for actual use - it's a bad pattern to have the entire agent stack run inside a container. In fact, most of the device connectors will not work well like this currently. What I would recommend for actual use, is to have *only* the test phase run inside a container. This way you get a clean, reproducible environment every time, while still allowing the other phases to operate in a less restricted way.

hanhsuan:
As discuss and the conclusion in Riga, OEM team will only use test command for now and ask me to put what OEM team has into this repository.

plars (changes requested):
After a bit more discussion today, here are some changes we are recommending in order to merge this:
1. Move this to agent/extra instead of agent/docker - this gives a better indication that this is "extra" and not needed for general use
2. Document at the top of the README that this is for a specific use case and that most device connectors will not function properly this way, so it's not recommended for use in most cases.

hanhsuan:
May I move to `agent/extra/docker` ? I think it might be more clear to know it is the extra container related items.

plars:
@hanhsuan yes, that should work

plars:
/canonical/self-hosted-runners/run-workflows 9e10be3f0e603125d1f9ddd0070a6f603a02ce7d

plars (approved):

PR #183, opened 2024-01-09
Add provision_type field to agent data

plars:
## Description
Add provision_type to the agent data that we store on the server

## Resolved issues
CERTTF-277

## Documentation
Nothing really, but openapi schema updated appropriately so that it will show up there for reference.

## Tests
Tested locally, and unit tests adjusted

### DIFF
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -32,13 +32,19 @@ class TestflingerAgent:
     def _post_initial_agent_data(self):
         """Post the initial agent data to the server once on agent startup"""
 
-        location = self.client.config.get("location", "")
         self._post_advertised_queues()
         self._post_advertised_images()
 
-        queues = self.client.config.get("job_queues", [])
         identifier = self.client.config.get("identifier")
-        agent_data = {"queues": queues, "location": location}
+        location = self.client.config.get("location", "")
+        provision_type = self.client.config.get("provision_type", "")
+        queues = self.client.config.get("job_queues", [])
+
+        agent_data = {
+            "location": location,
+            "queues": queues,
+            "provision_type": provision_type,
+        }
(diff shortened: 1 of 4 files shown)


server/src/api/schemas.py
```diff
@@ -27,6 +27,7 @@ class AgentIn(Schema):
     state = fields.String(required=False)
     queues = fields.List(fields.String(), required=False)
     location = fields.String(required=False)
+    provision_type = fields.String(required=False)
```
omar-selo, line 30: could `provision_type` be an enum? that would help with api docs too since it would show possible values

agent/testflinger_agent/agent.py
```diff
@@ -32,13 +32,19 @@ def __init__(self, client):
...
 
-        location = self.client.config.get("location", "")
         self._post_advertised_queues()
         self._post_advertised_images()
 
-        queues = self.client.config.get("job_queues", [])
         identifier = self.client.config.get("identifier")
-        agent_data = {"queues": queues, "location": location}
+        location = self.client.config.get("location", "")
+        provision_type = self.client.config.get("provision_type")
```
omar-selo, line 40: is it guaranteed to have a value or should we have some default?

omar-selo (commented):
Looks good, just a couple of questions

server/src/api/schemas.py
plars: At some point probably, it will be but that's something we just started discussing for other reasons. It will still likely need to convert to a string (string val enum, not int) though.

agent/testflinger_agent/agent.py
plars: Good idea, fixed. thanks!

omar-selo (approved):
Looks good!

PR #184, opened 2024-01-10
Add identifier to the schema so that it's imported properly

plars:
## Description
I think I may have just missed git adding the schema file from when I originally merged the support for sending identifer.

## Resolved issues
Without this, the identifier will be sent to the server, but it will just be thrown out as it's not part of the expected input data.

## Documentation
No changes, but adding it to the schema will automatically update the openapi docs

## Tests
Tested locally


omar-selo (approved):

plars:
needs re-review because I had to fix a merge conflict

PR #187, opened 2024-01-12
consolidate server github workflows to avoid race

plars:
## Description
Combine the workflow for building the oci image and pushing to ghcr into the one where we build and push the charm. It looks like sometimes there's a race where we don't get the latest image in time for the charm build process to see it. I'm hoping that this will enforce the order of these steps so that we don't have that problem

## Resolved issues
Described above

## Documentation
N/A

## Tests
Not sure about this... I thought I could just force the run manually, but it didn't do what I hoped. It uploaded the image to a different tag name based on the branch name - so not main. But the charm build needs to pull the image from main, so it didn't find the right sha256 sum for the latest image there, and then got an error about the resource not being available [here](https://github.com/canonical/testflinger/actions/runs/7496694073/job/20409131359). I suspect that this *should* work once it's properly merged, but I'm not sure how else to test it other than trying it - then either reverting it or fixing it if it doesn't work. If you have another idea though, I'm happy to try it :)


.github/workflows/server-charm-release-edge.yml
```diff
@@ -4,24 +4,60 @@ on:
   push:
     branches:
       - main
+      - consolidate-server-workflows
```
omar-selo, line 7: probably not needed anymore?

omar-selo (dismissed):
Looks good

.github/workflows/server-charm-release-edge.yml
plars: oops, I thought I had reverted that already. Done now, thanks!

mz2 (approved):
LGTM

PR #188, opened 2024-01-12
Add: iotscript connector with dedicated code owner

p-gentili:
## Description

This PR intends to bootstrap the IoT device connector maintained by the OEM Software Engineering team (@kiya956  mostly). 

I added an empty directory for their new connector, updated the README with a simple description and made their team code-owner of that particular directory. 

## Resolved issues

N/A

## Documentation

Updated the README with a small description, more documentation will come with the implementation I guess.

## Tests

N/A

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+# OEM Software Engineering
+device-connectors/src/testflinger_device_connectors/devices/iotscript @canonical/oem-swe-iot
--- a/device-connectors/README.rst
+++ b/device-connectors/README.rst
@@ -23,6 +23,7 @@ testing on other types of devices.
 - noprovision - devices which need to run tests, but can't be provisioned (yet)
 - oemrecovery - anything (such as core fde images) that can't be provisioned but can run a set of commands to recover back to the initial state
 - oemscript - uses a script that supports some oem images and allows injection of an iso to the recovery partition to install that image
+- iotscript - IoT devices which require a Zapper for HW manipulation and OEM-specific actions to get provisioned (UUU, seed-override, ...) 
 
 
 Exit Status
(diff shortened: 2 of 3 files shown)


plars (changes requested):
Most of the time, we would want it to be at least somewhat working before merging, but I can see the benefit of doing it like this for now. However, in this case, I think you either need something that either excludes this directory or implements just a bit more. Otherwise you end up with an error when trying to even run help:
```
❯ testflinger-device-connector -h                  
  File "/home/plars/work/cert/testflinger/env/bin/testflinger-device-connector", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/plars/work/cert/testflinger/device-connectors/src/testflinger_device_connectors/cmd.py", line 36, in main
    dev_subparser = dev_parser.add_parser(dev_name)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/argparse.py", line 1204, in add_parser
    raise ArgumentError(self, _('conflicting subparser: %s') % name)
argparse.ArgumentError: argument {lenovo_oemscript,oemrecovery,netboot,noprovision,hp_oemscript,multi,maas2,oemscript}: conflicting subparser: oemscript
```
Ironically, I'm planning to refactor this soon-ish, but probably not in the next cycle or two, as there's some more urgent things I'm supposed to do first. But once I do, this should be avoidable.

For now, this should get you going as a minimal stub, but in the future this should be easier:
```
from testflinger_device_connectors.devices import DefaultDevice

device_name = "iotscript"


class DeviceConnector(DefaultDevice):
    def provision(self, args):
        pass
```

p-gentili:
Oops sorry about that, I should have tried running the script. I've added what you suggested and it doesn't crash anymore. Thanks!

plars (approved):
+1

PR #191, opened 2024-01-25
Tagged job searching

plars:
## Description

The motivation for this is to help support the creation of a spread backend for testflinger. (see jira card below)
The spread backend would submit a job that requests the devices it wants but only have them provision and stay in the allocate state. Then it would later go clean up the ones that were left. Ideally, it should just use the job_id and cancel the jobs once it's done with them, but they have some circumstances where they are unable to retain those job IDs and need to find the jobs somehow. They accomplish this on some of the cloud backends they support by tagging the systems they are using, then searching for them by that tag name. They also need the timestamp when the job was created so they can automate cancellation of the ones that are old.

The search API here defaults to only consider jobs that are active (not cancelled or completed). However, it was pretty straightforward to also add support for specifying that you wan to search for jobs in a specific state. This could be used for all sorts of other things, and I think the search API might be extended in the future to cover some other needs that we haven't though of yet.

## Resolved issues

Fixes [CERTTF-192](https://warthogs.atlassian.net/browse/CERTTF-192https://warthogs.atlassian.net/browse/CERTTF-192)

## Documentation

Documentation added to the README and also to the job_schema reference docs.

## Tests

Unit tests modified and added to cover both the addition of tags to the job, as well as the search API and all the different ways it can be used. I also tried this out locally of course.

[CERTTF-192]: https://warthogs.atlassian.net/browse/CERTTF-192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/how-to/search-job.rst
+++ b/docs/how-to/search-job.rst
@@ -1,0 +1,63 @@
+Search Jobs
+=============
+
+You can search for jobs based on the following criteria:
+   * tags
+   * state
+
+
+Searching by Tags
+------------------
+
+Tags can make it easier to find related jobs later. For instance, tools like
+spread have a need to find currently running jobs so that they can be
+cancelled once they are done using those devices for their testing.
+
+To add tags to a job submission, add a section like this with one or more tags
+in your job yaml:
+
+.. code-block:: yaml
+
+      tags:
+        - myproject
+        - client
+
+Testflinger doesn't use any of this information, but it makes it easier for
+you to search for those jobs later.
+
+The search API allows you to search for jobs based on tags and state. To
+search for a jobs by tag, you can provide one or more tags in the query string:
+
+.. code-block:: console
+
+      $ curl 'http://localhost:8000/v1/job/search?tags=foo&tags=bar'
+
+By default, the search API will return jobs that match any of the tags. To
+require that all tags match, you can provide the "match" query parameter with
+the value "all":
+
+.. code-block:: console
+
+      $ curl 'http://localhost:8000/v1/job/search?tags=foo&tags=bar&match=all'
+
+Searching by Job State
+-----------------------
+
+By default, the search API will only return jobs that are not already marked as
+cancelled or completed. To specify searching for jobs in a specific state, you
+can provide the "state" query parameter with one of the
+:doc:`test phases <../reference/test-phases>`:
+
+.. code-block:: console
+
+      $ curl 'http://localhost:8000/v1/job/search?state=provision'
+
+You can also search specify more than one state to match against. Obviously,
+since a job can only be in one state at a given moment, the matching mode
+for this will always be "any".
+
+.. code-block:: console
+
+      $ curl 'http://localhost:8000/v1/job/search?state=cancelled&state=completed'
+
+This can be done with or without providing tags in the search query.
--- a/docs/how-to/search-job.rst
+++ b/docs/how-to/search-job.rst
@@ -43,10 +43,9 @@ the value "all":
 Searching by Job State
 -----------------------
 
-By default, the search API will only return jobs that are not already marked as
-cancelled or completed. To specify searching for jobs in a specific state, you
-can provide the "state" query parameter with one of the
-:doc:`test phases <../reference/test-phases>`:
+By default, the search API will match jobs in any state.  To specify searching
+for jobs in a specific state, you can provide the "state" query parameter with
+one of the :doc:`test phases <../reference/test-phases>`:
 
 .. code-block:: console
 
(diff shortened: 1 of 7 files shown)


plars:
Also added a how-to and tagged @tang-mm for the review on that

docs/how-to/search-job.rst
```diff
@@ -0,0 +1,55 @@
...
+the value "all":
+
+.. code-block:: console
+
+      $ curl 'http://localhost:8000/v1/job/search?tags=foo&tags=bar&match=all'
+
+Searching by Job State
+-----------------------
+
+By default, the search API will only jobs that are not already marked as
```
omar-selo, line 46:
```suggestion
By default, the search API will only return jobs that are not already marked as
```

docs/how-to/search-job.rst
```diff
@@ -0,0 +1,55 @@
...
+cancelled or completed. To specify searching for jobs in a specific state, you
+can provide the "state" query parameter with one of the
```
omar-selo, line 48: Following the principle of least astonishment, I would've expected the default to return all jobs regardless of state given that I didn't specify state

server/src/api/schemas.py
```diff
@@ -72,6 +73,24 @@ class JobId(Schema):
     job_id = fields.String(required=True)
 
 
+class JobSearchRequest(Schema):
+    """Job search request schema"""
+
+    tags = fields.List(fields.String, description="List of tags to search for")
+    match = fields.String(
+        description="Match mode - 'all' or 'any' (default 'any')"
+    )
```
omar-selo, line 82: We could enforce 'all' or 'any' via the `OneOf` [data validator](https://apiflask.com/schema/#data-validators). Then there is no need for us to check it in the code

server/src/api/schemas.py
```diff
@@ -72,6 +73,24 @@ class JobId(Schema):
...
+    state = fields.List(
+        fields.String, description="List of job states to include"
+    )
```
omar-selo, line 85: If the possible states are known then we could also use `OneOf` here

server/src/api/v1.py
```diff
@@ -146,6 +146,47 @@ def job_get_id(job_id):
...
 
 
+@v1.get("/job/search")
+@v1.input(schemas.JobSearchRequest, location="query")
+@v1.output(schemas.JobSearchResponse)
+def search_jobs(query_data):
+    """Search for jobs by tags"""
+    tags = query_data.get("tags")
+    match = request.args.get("match", "any")
+    states = request.args.getlist("state")
```
omar-selo, line 156: Oh reading the docs I didn't know that you can pass multiple states. Perhaps it would be useful to mention that there

server/src/api/v1.py
```diff
@@ -146,6 +146,47 @@ def job_get_id(job_id):
...
+
+    if match not in ["any", "all"]:
+        abort(400, "Invalid match mode")
+
+    query = {}
+    if tags and match == "all":
+        query["job_data.tags"] = {"$all": tags}
+    elif tags and match == "any":
+        query["job_data.tags"] = {"$in": tags}
```
omar-selo, line 165: What if no "tags" were passed. Since match defaults to any this would create the query `{"$in": []}` I assume? then would that not return anything?

omar-selo (changes requested):

server/src/api/schemas.py
plars: We don't currently have this as an enum or anything, and I could see it being soething that we change a bit soon. So it's only strings at the moment. Also, if we do end up wanting to add some "fake" ones that encompass more than one state it could affect this (see above)

server/src/api/v1.py
plars: no, because if tags is empty, then it won't execute this section - we check both that tags contains something and that match==any before doing this.

server/src/api/schemas.py
plars: we're also checking that "tag" exists at all, see the comments below.

docs/how-to/search-job.rst
omar-selo: > a really long search string for the common case. ex ?tags=foo&state=waiting&state=test&state=provision&state=allocated...

I see the issue. One idea then is to just expose through the api docs/schema that there is this default value for the state field. I was wondering whether we could use the property `dump_default` from `Field` model. But I'm not sure if it would work for lists. But if it does, then we could possibly specify the default through the schema and avoid handling it in code

server/src/api/schemas.py
omar-selo: Oh definitely, then something to consider once it's an enum

server/src/api/v1.py
omar-selo: Oh yeah I missed that :sweat_smile:

docs/how-to/search-job.rst
plars: Ok, check out the latest revision. I added a fake state called "active". If you specify this state, then it only considers jobs that are not "cancelled" or "completed".  Otherwise it considers all states unless you specify states to filter on.

omar-selo: I love it 👍

omar-selo (approved):
Looks great

PR #193, opened 2024-02-01
Use importlib instead of imp

plars:
## Description
- replace imp (deprecated in python 3.12) with importlib
- specify the list of supported connector modules rather than trying to detect them
- rework the argument subparsers to avoid loading unnecessary connectors
- add more tests

## Resolved issues

imp is deprecated and will no longer be available starting with python 3.12. That's not affecting us just yet, but noble looks likely to ship with python 3.12, so as an LTS it could be the new foundation that we deploy on very soon.
Also a good opportunity for some other related improvements

## Documentation
No real functional change, but should speed things up slightly when running the device connector.

## Tests
- added some unit tests, for argument parsing and initial load of the connectors
- tested on an agent running against staging

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/cmd.py
+++ b/device-connectors/src/testflinger_device_connectors/cmd.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-# Copyright (C) 2023 Canonical
+# Copyright (C) 2023-2024 Canonical
 #
 # This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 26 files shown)


plars:
...and quite a few black updates because it looks like there's a new version. The changes look fine though, and I kept it in a separate commit, in case you want to filter those out for easier review.

nadzyah (approved):
Haven't found anything questionable

PR #195, opened 2024-02-09
Encode datetime as ISO8601 when returning job search results

plars:
## Description

The spread backend wants RFC3339 compatible datetime fileds. This converts it to a format that is both compatible with that, as well as ISO8601. Nothing in the internal representation is changed, just the string that comes back in the json results from calls to the API.

## Documentation
N/A

## Tests
Unit test added to ensure the field is returned in the correct format

### DIFF
--- a/server/src/providers.py
+++ b/server/src/providers.py
@@ -1,0 +1,33 @@
+# Copyright (C) 2024 Canonical
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


server/src/providers.py
```diff
@@ -0,0 +1,33 @@
...
+
+class ISODatetimeProvider(
+    DefaultJSONProvider
+):  # pylint: disable=too-few-public-methods
+    """Return datetime objects as RFC3339/ISO8601 strings with a 'Z' suffix."""
+
+    def default(self, obj):
+        """Encode datetime objects as ISO8601, no effect on other types."""
+        if isinstance(obj, datetime):
+            return obj.isoformat(timespec="seconds") + "Z"
```
nadzyah, line 32: I'm just wondering what if timezone has already been specified?

plars: We currently grab the time from utcnow when the job is created. So it's not something the user sends, but we need to include the "Z" so that it's specific that it's UTC and so that it matches the expected format.

nadzyah (approved):

PR #197, opened 2024-02-13
remove logmsg helper function

plars:
## Description
There's not really much point to using this helper anymore. Let's remove it, make logging more clear, and also slightly improve the insertion of the agent string by using a formatter instead of a filter

PR #198, opened 2024-02-14
Add ZapperConnector base class for Zapper-based device connectors

p-gentili:
## Description

This PR introduces a new device connector base class for Zapper-based provisioning called `ZapperConnector`. Classes who inherits from it are responsible for config/data validation and preparation of arguments for the Zapper `provision` API. The actual logic of the provisioning procedure is implemented and executed on Zapper.

The base class implements the default `_run` method which connects to the Zapper RPyC service and runs the `provision` API. The local logger is passed via API as well so that the execution can be monitored from Testflinger.

Finally, I've removed the temporary _iotscript_ module, created a new starting point for the `zapper_iot` connector and adjusted CODEOWNER accordingly.

## Resolved issues

Resolves [ZAP-696](https://warthogs.atlassian.net/browse/ZAP-696)

## Documentation

Updated README.
Tests are included.

## Tests

Checked if CLI still works after the changes. Nothing else is really testable at this stage since this PR introduces only a base class and Zapper still doesn't provide any implementation.

[ZAP-696]: https://warthogs.atlassian.net/browse/ZAP-696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/__init__.py
@@ -39,7 +39,6 @@ DEVICE_CONNECTORS = (
     "dell_oemscript",
     "dragonboard",
     "hp_oemscript",
-    "iotscript",
     "lenovo_oemscript",
     "maas2",
     "multi",
@@ -48,6 +47,7 @@ DEVICE_CONNECTORS = (
     "noprovision",
     "oemrecovery",
     "oemscript",
+    "zapper_iot",
 )
 
 
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
@@ -74,6 +74,7 @@ class ZapperConnector(ABC, DefaultDevice):
         Validate the job config and data and prepare the arguments
         for the Zapper `provision` API.
         """
+        raise NotImplementedError
 
     def _run(self, zapper_ip, *args, **kwargs):
         """
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
@@ -1,0 +1,102 @@
+# Copyright (C) 2024 Canonical
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Package containing modules for implementing Zapper-driven device connectors.
+
+Modules inheriting from the provided abstract class will run Zapper-driven
+provisioning procedures via Zapper API. The provisioning logic is implemented
+in the Zapper codebase and the connector serves as a pre-processing step,
+validating the configuration and preparing the API arguments.
+"""
+
+import logging
+from abc import ABC, abstractmethod
+from typing import Any, Dict, Tuple
+
+import rpyc
+import yaml
+
+import testflinger_device_connectors
+from testflinger_device_connectors.devices import (
+    DefaultDevice,
+    RecoveryError,
+    catch,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class ZapperConnector(ABC, DefaultDevice):
+    """
+    Abstract base class defining a common interface for Zapper-driven
+    device connectors.
+    """
+
+    PROVISION_METHOD = ""  # to be defined in the implementation
+    ZAPPER_REQUEST_TIMEOUT = 60 * 90
+    ZAPPER_SERVICE_PORT = 60000
+
+    @catch(RecoveryError, 46)
+    def provision(self, args):
+        """Method called when the command is invoked."""
+        with open(args.config) as configfile:
+            config = yaml.safe_load(configfile)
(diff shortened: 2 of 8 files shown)


device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
```diff
@@ -0,0 +1,101 @@
...
+        (api_args, api_kwargs) = self._validate_configuration(
+            args.config, args.job_data
+        )
+
+        logmsg(logging.INFO, "BEGIN provision")
+        logmsg(logging.INFO, "Provisioning device")
+
+        self._run(args.config["controller_host"], *api_args, **api_kwargs)
+
+        logmsg(logging.INFO, "END provision")
```
nadzyah, line 66: This behaviour has been changed to logger.info in #197

p-gentili: ah, time to rebase

device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+        logger.info("END provision")
+
+    @abstractmethod
+    def _validate_configuration(
+        self, config, job_data
+    ) -> Tuple[Tuple[Any, ...], Dict[str, Any]]:
+        """
+        Validate the job config and data and prepare the arguments
+        for the Zapper `provision` API.
+        """
```
nadzyah, line 76: shall we raise NotImplementedError here as well?

p-gentili: The method is abstract, you'll get a TypeError if you try to create an object out of this class.

nadzyah: But what if a child class doesn’t implement this method?

p-gentili: Same effect, it'll raise TypeError.

nadzyah: Hmm, this is what's written in the [documentation](https://docs.python.org/3/library/abc.html#abc.abstractmethod):
> Note Unlike Java abstract methods, these abstract methods may have an implementation. This implementation can be called via the super() mechanism from the class that overrides it.

So, you can do something like this
image

I believe the desired behaviour is to tell a user that the method is not implemented, so the code would return an error if the exception was raised

p-gentili: In this case I don't want anything implemented in the abstract method (so `super()` is not relevant). I want the user to implement it. If they don't do it, they'll get a `TypeError` even before they try to run that method:

```
class A(ABC):
    @abstractmethod
    def mymethod(self):
        """Implement me."""
        
class B(A):
    ...
    
obj = B()  <-- this will raise TypeError
```

What I'm doing here is defining an interface so that every Zapper connector behave in the same way.

nadzyah: but in my case I also don't have anything implemented in my abstract class, and you still can call `super()` there. That's why I suggest raising an error in the abstract method itself, because marking it with `abstractmethod` doesn't mean we cannot call it

nadzyah: moreover, you can do something like this as well without seeing an error :D
 
```python3
>>> class A(ABC):
...     @abstractmethod
...     def mymethod(self):
...         """Implement me."""
... 
>>> A.mymethod(1)
```

p-gentili: Abstract class and abstract methods are designed for this exact purpose: making the child implement what's abstract and eventually override the default methods. I don't see benefits in raising an exception is someone runs `super()`, I'm actually fine with that as long as they implement their own stuff. Moreover, that method expects a return value, so a mere `return super().method()` implementation wouldn't be enough.

And for what concerns that example: not true, `mymethod` is not a static/class method so no way to call it with `A.mymethod(1)`. You need to create an object first which will raise `TypeError`.

p-gentili: You got it :) https://github.com/canonical/testflinger/pull/198/commits/a99f6ff643e5949a804e42b470c165b589415921

nadzyah (approved):
Thank you so much for the contribution and for your patience with my comments. Definitely looks good to me :D
But you probably want someone else to have a look at this as well

device-connectors/src/testflinger_device_connectors/devices/__init__.py
```diff
@@ -48,6 +47,7 @@
     "noprovision",
     "oemrecovery",
     "oemscript",
+    "zapper_iot",
```
plars, line 50: It looks like this is only meant as a base class for other device connectors to use, correct? If so, then you don't want to put it in DEVICE_CONNECTORS. These are intended for things that will be used by real devices and not just inherited, and will turn into subcommands for testflinger-device-connector cli

plars (commented):
One minor question, see below. Otherwise, looks great! and thanks for keeping that patches small and separate as much as possible. That helps a lot

device-connectors/src/testflinger_device_connectors/devices/__init__.py
p-gentili: The base class is just `zapper` and it's indeed not included in that list. On the other hand, `zapper_iot` would be the first real connector to inherit from it and it's the one owned by the OEM team. If you prefer, we could add it in there only once it gets properly implemented, for now I was just trying to set the ground for them.

plars: Ah, right. No I think that's fine then to go ahead and add it, since that's going to be the actual device agent that implements it. +1

plars (approved):
+1

PR #202, opened 2024-02-20
Include "completed" jobs in the ones that are not active

plars:
## Description
We added "completed" a while back and I forgot to include that here. This fixes it.

## Tests
Small adjustment to the unit tests to make sure this is covered.

### DIFF
--- a/server/src/api/v1.py
+++ b/server/src/api/v1.py
@@ -162,7 +162,9 @@ def search_jobs(query_data):
         query["job_data.tags"] = {"$in": tags}
 
     if "active" in states:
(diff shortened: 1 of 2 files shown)


omar-selo (commented):
What's the difference between "complete" and "completed"? do we need both? and if so can we somehow differentiate them?

plars:
> What's the difference between "complete" and "completed"? do we need both? and if so can we somehow differentiate them?

It used to be "complete" but it was recommended that I change it to "completed". However, since there could be some versions of the cli, agent, etc deployed that still expect "complete", then we're allowing lots of time before fully deprecating that version of it, just to be safe.

omar-selo (approved):

PR #208, opened 2024-02-27
remove oem image once provision is completed

nancyc12:
## Description
Currently testflinger agent won't remove the .iso images from /testflingere/run folder once provision is completed. Propose to add a step to remove image file at the end of provision phase for oemscript devices.

## Resolved issues
Reduce the manual effort to clean up accumulated image files in /testflinger/run folder.

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/oemscript/oemscript.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/oemscript/oemscript.py
@@ -103,6 +103,10 @@ class OemScript:
 
         self.check_device_booted()
 
+        # remove the .iso image
+        cmd = f"rm -f {image_file}"
+        subprocess.check_output(cmd, stderr=subprocess.STDOUT, timeout=60)
+
     def run_recovery_script(self, image_file):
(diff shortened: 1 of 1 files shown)


kevinyehk (approved):
Thanks for fixing this promptly, but I think it still needs @plars 's approval.

jocave (commented):
Seems like this could be a case for use of a context manager if we want to ensure cleanup happens even in an error situation.

plars (changes requested):
Yeah, a context manager would be nice but since you're just getting back the filename, it might not be practical. At the very least, I would definitely just use os.unlink() (or os.remove()) rather than running it as a subprocess command.

To get pretty much the same benefit as a context manager here, you could also wrap everything from the download() through self.check_device_booted() in a `try` block, and use something like:
```
finally:
    os.unlink(image_file)
```
to ensure that it gets removed regardless of anything else that happens there.

nancyc12:
Thank you @jocave and @plars  for the comments. It's truly something I missed.

plars (approved):
+1 thanks!

PR #213, opened 2024-02-29
Improve js filter code and use it with the new search box

plars:
## Description
Just realized today that the search box is broken now after the recent vanilla framework updates. This fixes it

## Tests
Tested locally

### DIFF
--- a/server/src/static/assets/js/filter.js
+++ b/server/src/static/assets/js/filter.js
@@ -1,19 +1,15 @@
-function runFilter() {
-  var input, filter, table, tr, td, i, txtValue;
-  input = document.getElementById("inputFilter");
-  filter = input.value.toUpperCase();
-  table = document.getElementById("filterTable");
-  tr = table.getElementsByTagName("tr");
+document.addEventListener('DOMContentLoaded', function() {
+    const searchBox = document.querySelector('.p-search-box__input');
+    const tableRows = Array.from(document.querySelectorAll('.p-table--mobile-card tbody tr'));
 
-  for (i = 0; i < tr.length; i++) {
-    td = tr[i].getElementsByTagName("td")[0];
-    if (td) {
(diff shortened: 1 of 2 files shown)


server/src/static/assets/js/filter.js
```diff
@@ -1,19 +1,15 @@
-function runFilter() {
-  var input, filter, table, tr, td, i, txtValue;
-  input = document.getElementById("inputFilter");
-  filter = input.value.toUpperCase();
-  table = document.getElementById("filterTable");
-  tr = table.getElementsByTagName("tr");
+document.addEventListener('DOMContentLoaded', function() {
+    const searchBox = document.querySelector('.p-search-box__input');
+    const tableRows = Array.from(document.querySelectorAll('.p-table--mobile-card tbody tr'));
```
omar-selo, line 3: While these selectors work, I'm worried that any changes to the style used would break this again. Wdyt if we set an id or class for searchBox (e.g. `search-box`) and a class for rows (e.g. `searchable-row`) and select on these instead? This way future modifications to style of these wont break functionality.

plars: Good idea, however I think it might be unnecessary for the existing ".p-search-box__input" ones since that's already clearly intended as a search box? See what you think, but I did add a "searchable-row" class and restricted the js search bits to only look at those. See what you think.

omar-selo: Yeah, actually it makes sense that search box is a search box (unless we end up with multiple search boxes in the future) so it's probably an overkill to create a class for it. I like the addition of the `searchable-row`. I was thinking of selecting on `searchable-row` only though. That way if any of the parents `.p-table--mobile-card tbody tr` change the search wont break. But that's optional.

omar-selo (approved):

PR #214, opened 2024-03-01
feat(muxpi): provisioning: skip user creation, boot check via URL

thp-canonical:
## Description

1. Allow skipping of user creation when deploying an image (for pre-configured images).
2. Allow a custom check URL to allow for testing services running in images.

## Documentation

Documentation with the newly-added keys is added. As part of this, the recognized image types (for user creation) are also documented.

## Tests

Untested proposal / request for comment.

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
@@ -19,6 +19,7 @@ import logging
 import subprocess
 import shlex
 import time
+import urllib.request
 from contextlib import contextmanager
 from pathlib import Path
 
@@ -162,11 +163,14 @@ class MuxPi:
         logger.info("Flashing Test image")
         try:
             self.flash_test_image(url)
-            with self.remote_mount():
-                image_type = self.get_image_type()
-                logger.info("Image type detected: {}".format(image_type))
-                logger.info("Creating Test User")
-                self.create_user(image_type)
+            if self.job_data["provision_data"].get("create_user", True):
+                with self.remote_mount():
+                    image_type = self.get_image_type()
+                    logger.info("Image type detected: {}".format(image_type))
+                    logger.info("Creating Test User")
+                    self.create_user(image_type)
+            else:
+                logger.info("Skipping test user creation (create_user=False)")
             self.run_post_provision_script()
             logger.info("Booting Test Image")
             cmd = self.config.get("control_switch_device_cmd", "stm -dut")
@@ -469,9 +473,30 @@ class MuxPi:
         test_password = self.job_data.get("test_data", {}).get(
             "test_password", "ubuntu"
         )
+
+        boot_check_url = self.job_data.get("provision_data", {}).get(
+            "boot_check_url", None
+        )
+        if boot_check_url is not None:
+            # We don't support full shell expansion of the URL, but just
+            # replace a literal $DEVICE_IP to the device's IP address
+            boot_check_url = boot_check_url.replace(
+                "$DEVICE_IP", self.config["device_ip"]
+            )
+
         while time.time() - started < 1200:
             try:
                 time.sleep(10)
+
+                if boot_check_url is not None:
+                    with urllib.request.urlopen(
+                        boot_check_url, timeout=5
+                    ) as response:
+                        if response.status == 200:
+                            return True
+
+                    continue
+
                 cmd = [
                     "sshpass",
                     "-p",
--- a/docs/reference/device-connector-types.rst
+++ b/docs/reference/device-connector-types.rst
@@ -19,7 +19,7 @@ To specify the commands to run by the device in each test phase, set the ``testf
    * - ``multi`` 
      - Experimental device type that is used for provisioning multiple other devices in order to coordinate a job across multiple devices at the same time.
    * - ``muxpi`` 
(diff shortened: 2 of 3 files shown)


docs/reference/device-connector-types.rst
```diff
@@ -32,3 +32,52 @@ To specify the commands to run by the device in each test phase, set the ``testf
...
+
+The ``muxpi`` device connector supports the following ``provision_data`` keys:
+
+.. list-table:: Supported ``provision_data`` keys for ``muxpi``
+   :header-rows: 1
+
+   * - Key
+     - Description
+   * - ``url``
+     - URL to a zstd-compressed disk image that is downloaded, decompressed and
```
plars, line 49: zstd handles multiple compression formats, but xz is typically used for images right now. I don't want to give the impression that it *must* be zstd compressed, so maybe we could change this a bit? Perhaps recommend xz but say that anything compatible with zstd is ok?

device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
```diff
@@ -434,22 +437,36 @@ def check_test_image_booted(self):
...
+                "sshpass",
+                "-p",
+                test_password,
+                "ssh-copy-id",
+                "-o",
+                "StrictHostKeyChecking=no",
+                "-o",
+                "UserKnownHostsFile=/dev/null",
+                "{}@{}".format(test_username, self.config["device_ip"]),
+            ]
```
plars, line 463: What do you think about f-stringing this as a DEFAULT_BOOT_CHECK_CMD or something, and making that the default above instead of None when we get the boot_check_cmd from the job?

plars (changes requested):
Overall, this looks pretty good and thanks a lot for also adding documentation around it! I have a couple of review comments below, but more generally, I was wondering what sort of command you would need to specify in order to determine that the device booted? especially since you  can't easily template the IP of the dut as part of it.
I'm wondering if we should just handle this as a special image type and check whether it booted in a static way, because we don't normally allow injection of commands like this during the provision process, and I think this could be fraught with problems, and potential for allowing someone to inadvertently render the device unusable until someone fixes it.

thp-canonical:
Thanks for your review. PR updated, responses to your questions inline below.

> I was wondering what sort of command you would need to specify in order to determine that the device booted?

Any command that e.g. checks if a HTTP endpoint is available:

```
curl --connect-timeout 2 "http://$DEVICE_IP:1234/foo/bar"
```

> especially since you can't easily template the IP of the dut as part of it.

That has now been fixed, I did have `env.update()` in there that exposes `$DEVICE_IP`, but I forgot `shell=True` in the `subprocess.check_output()` call, this should now work.

> I'm wondering if we should just handle this as a special image type and check whether it booted in a static way, because we don't normally allow injection of commands like this during the provision process, and I think this could be fraught with problems, and potential for allowing someone to inadvertently render the device unusable until someone fixes it.

One thing that could work is instead of allowing a generic `boot_check_cmd`, we just allow a `boot_check_tcp_port` (say), and if set, it will consider the device booted once a TCP connection to the destination port can be established (or `boot_check_http_port`, which will consider a successful `GET / HTTP/1.1` request on that port to mean the device is online). WDYT?

plars (changes requested):
I like your idea of replacing `boot_check_cmd` with `boot_check_tcp_port` or `boot_check_http_port`, however I think for the latter, it would suffice to just call it `boot_check_http` and then give it the URL with the port if needed.

Also, you probably saw, but it also looks like the auto-spell-check got tripped up on "xz" of all things.

thp-canonical:
> I like your idea of replacing `boot_check_cmd` with `boot_check_tcp_port` or `boot_check_http_port`, however I think for the latter, it would suffice to just call it `boot_check_http` and then give it the URL with the port if needed.

Went for `boot_check_url` (as it can be anything that's supported by `urlib.request.urlopen()`), but can change to `boot_check_http` if that sounds better to you. I reverted all the `DEFAULT_BOOT_CHECK_CMD` changes back to how it was initially to not unnecessarily touch existing code.

I opted to use `$DEVICE_IP` instead of any other kind of templating/replacement to be consistent with the usage of `$DEVICE_IP` in `test_cmds`, even though in `test_cmds`, it really is a proper environment variable, whereas here we just replace a literal `$DEVICE_IP` with the value (so e.g. `${DEVICE_IP}` wouldn't work here, even though it would work fine in `test_cmds`), this is hopefully made clear enough in the documentation:

<q>a literal ``$DEVICE_IP`` in the URL will be replaced with the IP address of the DUT.</q>

> Also, you probably saw, but it also looks like the auto-spell-check got tripped up on "xz" of all things.

Done.

plars:
/canonical/self-hosted-runners/run-workflows 552e8b3cc1ea4836f2853d66fffe9401ed25bdc1

plars (changes requested):
It looks like it's just failing the black formatting check. Would you mind re-running black on it and pushing it?

thp-canonical:
> It looks like it's just failing the black formatting check. Would you mind re-running black on it and pushing it?

Done. I also noticed that my PR fixes weren't signed commits, so I squashed all into a single signed commit.

plars:
/canonical/self-hosted-runners/run-workflows b8a0e204bd151525ce37b52d23f452356c332620'

PR #217, opened 2024-03-05
Fix jobid wrapping on agent page, and add link for jobs

plars:
## Description
Sometimes jobids might wrap on the agent page, which looks really inconsistent. This forces it to overflow instead, since I couldn't see that there was any good way to adjust the column sizes.
Also, it adds a link to the job page when a job exists.


omar-selo (approved):
Nice touch adding that link

PR #218, opened 2024-03-05
Make max pool size adjustable

plars:
## Description
In an attempt to still try to resolve why sometimes connections timeout, I'm trying to look at the possibility that the connections pools may be getting exhausted and staying that way for a while. I've tried increasing the number of pods to get more available connections and so far it hasn't helped too much. But I'm hoping maybe this could also give me something to tweak.

### DIFF
--- a/server/charm/config.yaml
+++ b/server/charm/config.yaml
@@ -7,3 +7,7 @@ options:
     default: 10
     description: Number of seconds to wait for keepalive connections
     type: int
+  max_pool_size:
+    default: 100
+    description: Maximum number of concurrent connections to the database
+    type: int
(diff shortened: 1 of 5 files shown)


omar-selo (approved):
Just for reference do you know for sure that the timeouts are caused by the database not the server?

plars:
> Just for reference do you know for sure that the timeouts are caused by the database not the server?

Not for certain, no. I do definitely see some random times where a query will take a long time, sometimes even as much as 3 seconds. However it's hard to see how some of the API calls timeout after 15s from something like that. The timeouts on those API calls happen very randomly. For example, someone showed me logs from yesterday where they saw one happen while it was waiting on a job to start, but I also have a long running job to try to see how often it occurs and haven't seen a single timeout in over 3 days.

As for the connection pool availability, I think I found a way on the db side to check that, and the typical usage is about 10 of >400 available now that I've deployed 4 units. However if I repeatedly check it for long enough, I'll sometimes see it jump up to almost 400 in use, but <1s later it's back down to about 10 or so.

I haven't found a way, so far, to see any kind of event where the connection pool is exhausted for a long duration (e.g. >1s), so figuring out whether this is actually the problem is a bit tricky. What I'm hoping, is that if I carefully increase the connection pool size while monitoring db server load, then perhaps I can find a point where it maxes out and *could* have created more, but doesn't ever have a need to.

PR #221, opened 2024-03-06
Add: Zapper KVM connector

p-gentili:
## Description

Introduces a Zapper connector for KVM provisioning. The connector inherits from the base class and just implements the function to prepare the API arguments.

## Resolved issues

Resolves [ZAP-699](https://warthogs.atlassian.net/browse/ZAP-699)

## Documentation

Updated the README.

## Tests
Tested running `testflinger-device-connector zapper_kvm provision -c config_30464.yaml job_30464.json` where

```
# config_30464.yaml
device_ip: "10.102.163.33"
control_host: "10.102.243.33"
reboot_script: 
  - snmpset -c private -v2c 10.102.196.162 .1.3.6.1.4.1.13742.6.4.1.2.1.2.1.7 i 0
  - sleep 5
  - snmpset -c private -v2c 10.102.196.162 .1.3.6.1.4.1.13742.6.4.1.2.1.2.1.7 i 1
```

```
# job_30464.json
{
"job_queue": "test",
"provision_data": {
  "url": "http://cdimage.ubuntu.com/daily-live/current/noble-desktop-amd64.iso",
  "robot_tasks": [
    "hp/pro/sff_400_g9/boot/boot_from_usb.robot",
    "common/grub/autoinstall/http/configure_autoinstall_http.robot",
    "hp/boot/wait_reboot.robot"
    ],
  "storage_layout": "lvm"
},
"test_data": {
  "test_username": "ubuntu",
  "test_password": "insecure"
  }
}
```

[ZAP-699]: https://warthogs.atlassian.net/browse/ZAP-699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
@@ -57,8 +57,6 @@ class ZapperConnector(ABC, DefaultDevice):
         with open(args.job_data, encoding="utf-8") as job_json:
             self.job_data = json.load(job_json)
 
-        testflinger_device_connectors.configure_logging(self.config)
-
         logger.info("BEGIN provision")
         logger.info("Provisioning device")
 
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
@@ -70,7 +70,7 @@ class ZapperConnector(ABC, DefaultDevice):
     @abstractmethod
     def _validate_configuration(
         self,
-    ) -> Tuple[Tuple[Any, ...], Dict[str, Any]]:
+    ) -> Tuple[Tuple, Dict[str, Any]]:
         """
         Validate the job config and data and prepare the arguments
         for the Zapper `provision` API.
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
@@ -21,6 +21,7 @@ in the Zapper codebase and the connector serves as a pre-processing step,
 validating the configuration and preparing the API arguments.
 """
 
+import json
 import logging
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Tuple
@@ -51,24 +52,24 @@ class ZapperConnector(ABC, DefaultDevice):
     @catch(RecoveryError, 46)
     def provision(self, args):
         """Method called when the command is invoked."""
-        with open(args.config) as configfile:
-            config = yaml.safe_load(configfile)
-        testflinger_device_connectors.configure_logging(config)
+        with open(args.config, encoding="utf-8") as configfile:
+            self.config = yaml.safe_load(configfile)
+        with open(args.job_data, encoding="utf-8") as job_json:
+            self.job_data = json.load(job_json)
 
-        (api_args, api_kwargs) = self._validate_configuration(
-            args.config, args.job_data
-        )
+        testflinger_device_connectors.configure_logging(self.config)
 
         logger.info("BEGIN provision")
         logger.info("Provisioning device")
 
-        self._run(args.config["controller_host"], *api_args, **api_kwargs)
+        (api_args, api_kwargs) = self._validate_configuration()
+        self._run(self.config["control_host"], *api_args, **api_kwargs)
 
         logger.info("END provision")
 
     @abstractmethod
     def _validate_configuration(
-        self, config, job_data
+        self,
     ) -> Tuple[Tuple[Any, ...], Dict[str, Any]]:
         """
         Validate the job config and data and prepare the arguments
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
@@ -29,7 +29,6 @@ from typing import Any, Dict, Tuple
 import rpyc
 import yaml
 
-import testflinger_device_connectors
 from testflinger_device_connectors.devices import (
     DefaultDevice,
     RecoveryError,
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
@@ -26,7 +26,7 @@ class DeviceConnector(ZapperConnector):
 
     def _validate_configuration(
         self,
-    ) -> Tuple[Tuple[Any, ...], Dict[str, Any]]:
+    ) -> Tuple[Tuple, Dict[str, Any]]:
         """
         Validate the job config and data and prepare the arguments
         for the Zapper `provision` API.
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
@@ -22,10 +22,10 @@ from testflinger_device_connectors.devices.zapper import ZapperConnector
 class DeviceConnector(ZapperConnector):
     """Tool for provisioning baremetal with a given image."""
 
-    PROVISION_METHOD = "zapper_iot"
+    PROVISION_METHOD = "ProvisioningIoT"
 
     def _validate_configuration(
-        self, config, job_data
+        self,
     ) -> Tuple[Tuple[Any, ...], Dict[str, Any]]:
         """
         Validate the job config and data and prepare the arguments
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
@@ -50,19 +50,14 @@ class DeviceConnector(ZapperConnector):
         for the Zapper `provision` API.
         """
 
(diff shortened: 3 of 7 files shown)


device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
```diff
@@ -51,24 +52,24 @@ class ZapperConnector(ABC, DefaultDevice):
...
+        (api_args, api_kwargs) = self._validate_configuration()
+        self._run(self.config["control_host"], *api_args, **api_kwargs)
 
         logger.info("END provision")
 
     @abstractmethod
     def _validate_configuration(
-        self, config, job_data
+        self,
     ) -> Tuple[Tuple[Any, ...], Dict[str, Any]]:
```
nadzyah, line 73: If you're using Python > 3.5, you can use it like this (see [docs](https://docs.python.org/3/library/typing.html) for more):
```suggestion
    ) -> tuple[tuple[Any, ...], dict[str, Any]]:
```

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -22,10 +22,10 @@
 class DeviceConnector(ZapperConnector):
     """Tool for provisioning baremetal with a given image."""
 
-    PROVISION_METHOD = "zapper_iot"
+    PROVISION_METHOD = "ProvisioningIoT"
 
     def _validate_configuration(
-        self, config, job_data
+        self,
     ) -> Tuple[Tuple[Any, ...], Dict[str, Any]]:
```
nadzyah, line 29: and here as well
```suggestion
    ) -> tuple[tuple[Any, ...], dict[str, Any]]:
```

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+            ]["base_user_data"]
+
+        with open(os.path.expanduser("~/.ssh/id_rsa.pub")) as pub:
+            autoinstall_conf["authorized_keys"] = [pub.read()]
+
+        return autoinstall_conf
+
+    def _validate_configuration(
+        self,
+    ) -> Tuple[Tuple[Any, ...], Dict[str, Any]]:
```
nadzyah, line 52:
```suggestion
    ) -> tuple[tuple[Any, ...], dict[str, Any]]:
```

nadzyah (commented):
Just a small suggestion c:

device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
p-gentili: Mh unfortunately the lower bound for this syntax (TIL is called [Generic Alias Type](https://docs.python.org/3/library/stdtypes.html#generic-alias-type)) is 3.9. What you can do since 3.5 is writing [your own alias](https://docs.python.org/3/library/typing.html#type-aliases) like `type MyDict = dict[T, T]` which is basically typing.Dict (https://docs.python.org/3/library/typing.html#aliases-to-built-in-types).

p-gentili:
I built the test configuration and job files by hand, please check if the format is right while reviewing.

device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
nadzyah: Oh, I see. We can deprecate it once we drop Python 3.8 support😌

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+        """Prepare autoinstall-related configuration."""
+        autoinstall_conf = {}
+
+        autoinstall_conf["storage_layout"] = self.job_data["provision_data"][
+            "storage_layout"
+        ]
+
+        autoinstall_conf["storage_password"] = self.job_data[
+            "provision_data"
+        ].get("storage_password")
```
nadzyah, line 38: I guess this part of code can be simplified the following way:

```suggestion
        autoinstall_conf = {
            "storage_layout": self.job_data["provision_data"]["storage_layout"],
            "storage_password": self.job_data["provision_data"].get("storage_password")
        }
```

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+        )
+        provisioning_data["password"] = self.job_data.get(
+            "test_password", {}
+        ).get("test_password", "ubuntu")
+        provisioning_data["autoinstall_conf"] = self._get_autoinstall_conf()
+        provisioning_data["reboot_script"] = self.config["reboot_script"]
+        provisioning_data["device_ip"] = self.config["device_ip"]
+        provisioning_data["robot_tasks"] = self.job_data["provision_data"][
+            "robot_tasks"
+        ]
```
nadzyah, line 71: Same here
```suggestion
        provisioning_data = {
            "url": self.job_data["provision_data"]["url"],
            "username": self.job_data.get("test_data", {}).get("test_username", "ubuntu"),
            "password": self.job_data.get("test_password", {}).get("test_password", "ubuntu"),
            "autoinstall_conf": self._get_autoinstall_conf(),
            "reboot_script": self.config["reboot_script"],
            "device_ip": self.config["device_ip"],
            "robot_tasks": self.job_data["provision_data"]["robot_tasks"]
        }
```

plars (changes requested):
Looking good! just a few small suggestions in addition to what Nadzeya already wrote.

Some of these are inline below, but also...

There are a lot of these device agents that aren't too easy to write unit tests for, but I'd like to refactor them in the future to help with that. For this one though, you've structured it nicely so that you could pass in some data and know what you should expect to get back for the configuration from the return in these methods. Would you mind adding some simple unit tests for these? It would be great to get that started for future developments on these connectors!

Also one small opportunistic change that I think you could make if you want - it's in the ../zapper module, I don't think you need the `testflinger_device_connectors.configure_logging(self.config)` any longer if you want to remove that. This is now executed when the command runs.

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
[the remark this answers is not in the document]
p-gentili: Makes sense.

p-gentili:
> There are a lot of these device agents that aren't too easy to write unit tests for, but I'd like to refactor them in the future to help with that. For this one though, you've structured it nicely so that you could pass in some data and know what you should expect to get back for the configuration from the return in these methods. Would you mind adding some simple unit tests for these? It would be great to get that started for future developments on these connectors!

Done! Plus, I've found a problem while writing them :)
 
> Also one small opportunistic change that I think you could make if you want - it's in the ../zapper module, I don't think you need the `testflinger_device_connectors.configure_logging(self.config)` any longer if you want to remove that. This is now executed when the command runs.

Ah, that's why I was getting duplicates, removed.

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/tests/test_zapper_kvm.py
```diff
@@ -0,0 +1,135 @@
...
+            },
+        }
+
+        with patch("builtins.open") as mock_open:
+            conf = connector._get_autoinstall_conf()
+        self.assertEqual(conf["storage_layout"], "lvm")
+        self.assertIsNone(conf["storage_password"])
+        self.assertNotIn("base_user_data", conf.keys())
+        self.assertIn(
+            mock_open.return_value.__enter__.return_value.read.return_value,
```
nadzyah, line 97: IDK, this is probably correct, but seems somewhat wrong to me. Is there any way to make it simpler? Like for instance, move it to the context manager itself

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/tests/test_zapper_kvm.py
```diff
@@ -0,0 +1,135 @@
...
+            },
+        }
+
+        with patch("builtins.open") as mock_open:
+            conf = connector._get_autoinstall_conf()
+        self.assertEqual(conf["storage_layout"], "lvm")
+        self.assertEqual(conf["storage_password"], "luks")
+        self.assertEqual(conf["base_user_data"], "base data content")
+        self.assertIn(
+            mock_open.return_value.__enter__.return_value.read.return_value,
```
nadzyah, line 133: Same here

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/tests/test_zapper_kvm.py
p-gentili: It's the more general way to handle context managers but [mock_open](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.mock_open) is definitely better in this case. Thanks!

plars (approved):
Thanks for the speedy fixes, it looks like you got everything I mentioned. +1 from me!

nadzyah (approved):
+1, thanks for your contribution and patience🙏

PR #222, opened 2024-03-07
Reboot the control host before provisioning

plars:
This helps avoid situations where networking on nanopi devices goes crazy. Sometimes it inexplicably gets slow, but rebooting fixes it. The hope is that this will give us a little more stability if we avoid this problem.

## Tests
I tested this on several devices in the lab. The only issue I found is that occasionally running reboot over ssh would close the connection before the command is even finished, resulting in an ssh error which might be hard to distinguish from other errors. I switched to using shutdown -r for this reason.

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
@@ -86,6 +86,38 @@ class MuxPi:
             raise ProvisioningError(e.output)
         return output
 
+    def check_control_alive(self):
+        """Check if the control host is alive"""
+        try:
+            self._run_control("true")
+        except (ProvisioningError, subprocess.SubprocessError):
+            raise ProvisioningError(
+                "Control host is not responding, provisioning can't proceed!"
+            )
+
+    def reboot_control_host(self):
+        """
+        Reboot the control host
+
+        Sometimes the control host can end up in an unstable condition which
+        isn't easy to detect until it's too late. Since nothing else should
+        be using the control host, reboot it before provisioning to ensure
+        it's in a known good state.
+        """
+        self.check_control_alive()
+        logger.info("Rebooting control host")
+        # Use shutdown instead of reboot to avoid ssh disconnecting too quickly
+        self._run_control("sudo shutdown -r +1")
+        for _ in range(3):
+            time.sleep(60)
+            try:
+                self.check_control_alive()
+                break
+            except ProvisioningError:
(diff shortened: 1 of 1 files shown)


device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
```diff
@@ -86,6 +86,38 @@ def _run_control(self, cmd, timeout=60):
...
+        logger.info("Rebooting control host")
+        # Use shutdown instead of reboot to avoid ssh disconnecting too quickly
+        self._run_control("sudo shutdown -r +1")
+        for _ in range(3):
+            time.sleep(60)
+            try:
+                self.check_control_alive()
+                break
+            except ProvisioningError:
+                logger.info("Waiting for control host to become active...")
```
kissiel, line 117: I know I'm crazy, but since it's already a busy loop, can we reduce the sleep time?
Because we need to wait a bit longer after issuing the shutdown, I'd go with big wait outside the loop, and then denser checks.

device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
```diff
@@ -86,6 +86,38 @@ def _run_control(self, cmd, timeout=60):
...
+
+        Sometimes the control host can end up in an unstable condition which
+        isn't easy to detect until it's too late. Since nothing else should
+        be using the control host, reboot it before provisioning to ensure
+        it's in a known good state.
+        """
+        self.check_control_alive()
+        logger.info("Rebooting control host")
+        # Use shutdown instead of reboot to avoid ssh disconnecting too quickly
+        self._run_control("sudo shutdown -r +1")
```
kissiel, line 110: This with the sleep(60) introduces a perfect race :)
There's a small, but possible chance of python resuming before the shutdown machine completed, yielding a premature completion of `check_control_alive()`

kissiel (changes requested):
Awesome stuff. There's a tiny race condition in the code + one request :)

device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
kissiel: Oh, and fixing this may render the last check not needed.

plars: Good points, thanks. The changes I just pushed should take care of both I think

kissiel (approved):
Awesome stuff!

PR #225, opened 2024-03-18
Refactor run with log

plars:
## Description
I've been looking at some ways to refactor some of the hairy pieces currently in the agent code, improve modularity, testability, etc. I think this improves this also gives us a better foundation to build on with the rest of it.

## Tests
Added some unit tests, and also tested this with a real device connector and several full runs of provision+test+reserve in the staging instance.

### DIFF
--- a/agent/testflinger_agent/runner.py
+++ b/agent/testflinger_agent/runner.py
@@ -26,11 +26,14 @@ from typing import Callable, Optional, List
 
 logger = logging.getLogger(__name__)
 
+OutputHandlerType = Callable[[str], None]
+StopConditionType = Callable[[], Optional[str]]
+
 
 class CommandRunner:
     def __init__(self, cwd: Optional[str], env: Optional[dict]):
-        self.output_handlers: List[Callable] = []
-        self.stop_condition_checkers: List[Callable] = []
+        self.output_handlers: List[OutputHandlerType] = []
+        self.stop_condition_checkers: List[StopConditionType] = []
         self.process: Optional[subprocess.Popen] = None
         self.cwd = cwd
         self.env = os.environ.copy()
@@ -39,29 +42,25 @@ class CommandRunner:
                 {k: str(v) for k, v in env.items() if isinstance(v, str)}
             )
 
-    def register_output_handler(self, handler: Callable[[str], None]):
+    def register_output_handler(self, handler: OutputHandlerType):
         self.output_handlers.append(handler)
 
     def post_output(self, data: str):
         for handler in self.output_handlers:
             handler(data)
 
-    def register_stop_condition_checker(
-        self, checker: Callable[[], Optional[str]]
-    ):
+    def register_stop_condition_checker(self, checker: StopConditionType):
         self.stop_condition_checkers.append(checker)
 
-    def check_stop_conditions(self) -> bool:
+    def check_stop_conditions_and_post_output(self) -> bool:
         for checker in self.stop_condition_checkers:
             output = checker()
             if output:
-                # This shouldn't happen, but makes mypy happy
-                assert self.process is not None
                 self.post_output(output)
                 return True
         return False
 
-    def check_output(self):
+    def check_and_post_output(self):
         raw_output = self.process.stdout.read()
         if not raw_output:
             return
@@ -83,6 +82,7 @@ class CommandRunner:
         self.process.wait()
 
     def cleanup(self):
+        signal.signal(signal.SIGTERM, signal.SIG_DFL)
         if self.process is not None:
             self.process.kill()
 
@@ -102,17 +102,16 @@ class CommandRunner:
         while self.process.poll() is None:
             time.sleep(10)
 
-            if self.check_stop_conditions():
+            if self.check_stop_conditions_and_post_output():
                 self.cleanup()
                 break
 
-            self.check_output()
+            self.check_and_post_output()
+
         # Check for any final output before exiting
-
         run_cmd_thread.join()
-        self.check_output()
-
-        signal.signal(signal.SIGTERM, signal.SIG_DFL)
+        self.check_and_post_output()
+        self.cleanup()
 
         return self.process.returncode
 
--- a/agent/testflinger_agent/runner.py
+++ b/agent/testflinger_agent/runner.py
@@ -1,0 +1,128 @@
+# Copyright (C) 2024 Canonical
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>
+
+import logging
+import os
+import fcntl
+import sys
+import shlex
+import signal
+import subprocess
+import threading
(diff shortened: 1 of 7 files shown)


agent/testflinger_agent/runner.py
```diff
@@ -0,0 +1,128 @@
...
+        self.process: Optional[subprocess.Popen] = None
+        self.cwd = cwd
+        self.env = os.environ.copy()
+        if env:
+            self.env.update(
+                {k: str(v) for k, v in env.items() if isinstance(v, str)}
+            )
+
+    def register_output_handler(self, handler: Callable[[str], None]):
+        self.output_handlers.append(handler)
```
omar-selo, line 43: What do you think of defining a type alias for output handlers?

agent/testflinger_agent/runner.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def post_output(self, data: str):
+        for handler in self.output_handlers:
+            handler(data)
+
+    def register_stop_condition_checker(
+        self, checker: Callable[[], Optional[str]]
+    ):
+        self.stop_condition_checkers.append(checker)
```
omar-selo, line 52: Same point here, what do you think of defining a type alias for stop condition checker?

agent/testflinger_agent/runner.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def check_stop_conditions(self) -> bool:
+        for checker in self.stop_condition_checkers:
+            output = checker()
+            if output:
+                # This shouldn't happen, but makes mypy happy
+                assert self.process is not None
```
omar-selo, line 59: I'm not sure I understand why `assert self.process is not None` is necessary. This function isn't using self.process right? If it really is a requirement, then we can even pull it out of the for loop right?

agent/testflinger_agent/runner.py
```diff
@@ -0,0 +1,128 @@
...
+                self.post_output(output)
```
omar-selo, line 60: Small point here, but the name `check_stop_conditions` doesn't sound that it performs side effects. Perhaps renaming it would be helpful

agent/testflinger_agent/runner.py
```diff
@@ -0,0 +1,128 @@
...
+                return True
+        return False
+
+    def check_output(self):
+        raw_output = self.process.stdout.read()
+        if not raw_output:
+            return
+
+        output = raw_output.decode(sys.stdout.encoding, errors="replace")
+        self.post_output(output)
```
omar-selo, line 70: Same point here `check_output` doesn't sound like it does any side effects

agent/testflinger_agent/runner.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # Make sure to wait until the process actually starts
+        while self.process is None:
+            time.sleep(1)
+
+        while self.process.poll() is None:
+            time.sleep(10)
+
+            if self.check_stop_conditions():
+                self.cleanup()
```
omar-selo, line 106: should we also cleanup if any exception was raised?

agent/testflinger_agent/runner.py
```diff
@@ -0,0 +1,128 @@
...
+                break
+
+            self.check_output()
+        # Check for any final output before exiting
+
+        run_cmd_thread.join()
+        self.check_output()
+
+        signal.signal(signal.SIGTERM, signal.SIG_DFL)
```
omar-selo, line 115: this too seems part of cleanup, and perhaps it doesn't get called if something was raised

agent/testflinger_agent/runner.py
```diff
@@ -0,0 +1,128 @@
...
+
+def set_nonblock(fd: int):
+    """Set the specified fd to nonblocking output
+
+    :param fd:
+        File descriptor that should be set to nonblocking mode
+    """
+
+    fl = fcntl.fcntl(fd, fcntl.F_GETFL)
+    fcntl.fcntl(fd, fcntl.F_SETFL, fl | os.O_NONBLOCK)
```
omar-selo, line 128: does this action need cleaning? haven't got the chance to dabble with nonblocking mode before

agent/testflinger_agent/stop_condition_checkers.py
```diff
@@ -0,0 +1,51 @@
...
+
+class OutputTimeoutChecker:
+    def __init__(self, timeout: int):
+        self.timeout = timeout
+        self.start_time = time.time()
+
+    def __call__(self) -> Optional[str]:
+        if time.time() - self.start_time > self.timeout:
+            return f"\nERROR: Output timeout reached! ({self.timeout}s)\n"
+        return None
```
omar-selo, line 51: While start_time is initialized here, there are a few actions that happen before the process starts including some sleeping to wait for the process to start. Realistically I'm not sure how much difference there can be between start_time and the actual process start time. Perhaps it's not an issue. I was wondering however if there is some way to get the time spent from the process itself without measuring it ourselves.

omar-selo (commented):
Overall a great refactor, the code is now much easier to understand and modular. This is my first time going through agent code so bear with me as I had lots of questions.

agent/testflinger_agent/runner.py
plars: Sure, added `OutputHandlerType`

agent/testflinger_agent/runner.py
plars: Added `StopConditionType`

agent/testflinger_agent/runner.py
plars: Ah, yeah. I used to have something that used self.process here but it was moved. Good catch

agent/testflinger_agent/runner.py
plars: Changed to `check_stop_conditions_and_post_output` which I think is kinda wordy. Happy to take suggestions if you can think of something that's more clear and succinct

agent/testflinger_agent/runner.py
plars: changed to `check_and_post_output`

agent/testflinger_agent/runner.py
plars: I *think* we should be pretty safe here since we are just calling process.poll() right? If an exception is raised in the execution of the process itself, it wouldn't be raised here. Instead, the process would just exit and we would see that it's ended and exit the loop. Is there a specific case you're concerned about?

agent/testflinger_agent/runner.py
plars: hmm, ok how about this. We can move the reset of the signal to the default handler in cleanup function itself.

agent/testflinger_agent/runner.py
plars: No, it's just acting on the stdout from the process.

agent/testflinger_agent/stop_condition_checkers.py
plars: I'm not aware of a way to do that, but that would be very nice indeed! I don't think it's *too* big of a concern here. We're not working on the granularity of milliseconds here, but more like 15min+. If we spend even a few extra seconds before the process actually starts (unlikely) then I don't think that would be a problem.

agent/testflinger_agent/runner.py
omar-selo: Makes sense actually

agent/testflinger_agent/stop_condition_checkers.py
omar-selo: Makes sense, I found Popen.wait can take a timeout but that wouldn't work for OutputTimeoutChecker if I understand it correctly, so this solution suffices.

omar-selo (approved):
Looks great, and thanks for resolving my comments

PR #226, opened 2024-03-18
Disable Unattended Upgrades

rmartin013:
This upgrade process, when triggered, can block a job for half an hour (sometimes even more). This is unecessarely annoying

## Description

Disable unattended upgrade for Riverside jobs (see https://linuxhint.com/enable-disable-unattended-upgrades-ubuntu/)

## Resolved issues
* https://warthogs.atlassian.net/browse/PERI-415

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
@@ -377,6 +377,13 @@ class MuxPi:
                 )
                 self._run_control(cmd)
 
+                # Disable Unattended Upgrades
+                cmd = (
+                    'sudo sed -i \'s/"1"/"0"/g\' '
+                    f"{base}/etc/apt/apt.conf.d/20auto-upgrades"
+                )
+                self._run_control(cmd)
+
                 self._configure_sudo()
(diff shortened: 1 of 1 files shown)


mz2:
> BTW is any test plan happy with that feature? My gut feeling is that we could as well generalize this config change, and every project would benefit from it. 

This was going to be my question too having read your comment, @plars. Why would we want unattended updates on (aside from perhaps a hypothetical specific scenario intended to test it?)

plars:
> > BTW is any test plan happy with that feature? My gut feeling is that we could as well generalize this config change, and every project would benefit from it.
> 
> This was going to be my question too having read your comment, @plars. Why would we want unattended updates on (aside from perhaps a hypothetical specific scenario intended to test it?) any job?

I don't know of a definite case where we want to keep unattended updates, however they are also notoriously difficult to disable completely, and the process for doing so could change from one release to another. For that reason, I think it's more maintainable and less error prone to keep that process out of the provisioning step, and also gives more control and transparency about when and how it's being done.
We could do this during provisioning for only riverside if that's really what we want, because it only uses classic images from one series right now. For other device types, we would need to first determine what type of image and what release to ensure we can do it, and do it the right way.

PR #227, opened 2024-03-18
Muxpi improvement

rmartin013:
## Description

* Improved little bit the way we avoid rebooting zapper controllers
* Introduced a new "media" parameter in 'provision_data' section (as discussed with @mz2). This optional parameter is currently quite specific to Riverside devices, but is designed to be as generic as possible. The existing behavior is preserved if omitted. It is intended to select one of the 2 zapper tools as a boot media:
1. typecmux + usb stick (usb): the usb stick must be symlinked in zapper to /dev/tc-disk
2. sdwire + sd card (sd): sdwire must be symlinked in zapper to /dev/sd-disk

## Resolved issues
* https://warthogs.atlassian.net/browse/PERI-436

## Tests
Tested with agents:
* nvidia-jetson-agx-orin-c31646
* nvidia-jetson-orin-nano-c33506

Tested combinations:
* no 'media' parameter
* media : usb
* media : sd

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
@@ -131,8 +131,8 @@ class MuxPi:
             media = self.job_data["provision_data"]["media"]
             try:
                 # If media option is provided, then DUT is probably capable of
-                # booting from different media, we should switch both of them to TS
-                # side regardless of which one was previously used
+                # booting from different media, we should switch both of them
+                # to TS side regardless of which one was previously used
                 cmd = "zapper sdwire plug_to_self"
                 sd_node = self._run_control(cmd)
                 cmd = "zapper typecmux plug_to_self"
--- a/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
@@ -129,13 +129,12 @@ class MuxPi:
             self._run_control(cmd)
         else:
             media = self.job_data["provision_data"]["media"]
-            match media:
-                case "sd":
-                    self.test_device = "/dev/sd-disk"
-                case "usb":
-                    self.test_device = "/dev/tc-disk"
-                case _:
-                    self.test_device = self.config["test_device"]
+            if media == "sd":
+                self.test_device = "/dev/sd-disk"
+            elif media == "usb":
+                self.test_device = "/dev/tc-disk"
+            else:
+                self.test_device = self.config["test_device"]
             # If media option is provided, then DUT is probably capable of
             # booting from different media, we should switch both of them to TS
             # side regardless of which one was previously used
--- a/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
@@ -129,22 +129,29 @@ class MuxPi:
             self._run_control(cmd)
         else:
             media = self.job_data["provision_data"]["media"]
-            if media == "sd":
-                self.test_device = "/dev/sd-disk"
-            elif media == "usb":
-                self.test_device = "/dev/tc-disk"
-            else:
-                self.test_device = self.config["test_device"]
-            # If media option is provided, then DUT is probably capable of
-            # booting from different media, we should switch both of them to TS
-            # side regardless of which one was previously used
-            cmd = "zapper sdwire set ZAPPER; zapper typecmux set TS"
             try:
-                self._run_control(cmd)
+                # If media option is provided, then DUT is probably capable of
+                # booting from different media, we should switch both of them to TS
+                # side regardless of which one was previously used
+                cmd = "zapper sdwire plug_to_self"
+                sd_node = self._run_control(cmd)
+                cmd = "zapper typecmux plug_to_self"
+                usb_node = self._run_control(cmd)
             except Exception:
                 pass
+            if media == "sd":
+                self.test_device = sd_node.decode()
+            elif media == "usb":
+                self.test_device = usb_node.decode()
+            else:
+                raise ProvisioningError(
+                    'The "media" value in '
+                    'the "provision_data" section of '
+                    "your job_data must be either "
+                    "'sd' or 'usb'"
+                )
         time.sleep(5)
-        logger.info("Flashing Test image")
+        logger.info(f"Flashing Test image on {self.test_device}")
         try:
             self.flash_test_image(url)
             if self.job_data["provision_data"].get("create_user", True):
--- a/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
@@ -122,8 +122,28 @@ class MuxPi:
                 'the "provision_data" section of '
                 "your job_data"
             )
-        cmd = self.config.get("control_switch_local_cmd", "stm -ts")
-        self._run_control(cmd)
+        if "media" not in self.job_data["provision_data"]:
+            media = None
+            self.test_device = self.config["test_device"]
+            cmd = self.config.get("control_switch_local_cmd", "stm -ts")
+            self._run_control(cmd)
+        else:
+            media = self.job_data["provision_data"]["media"]
+            match media:
+                case "sd":
+                    self.test_device = "/dev/sd-disk"
+                case "usb":
+                    self.test_device = "/dev/tc-disk"
+                case _:
+                    self.test_device = self.config["test_device"]
+            # If media option is provided, then DUT is probably capable of
+            # booting from different media, we should switch both of them to TS
+            # side regardless of which one was previously used
+            cmd = "zapper sdwire set ZAPPER; zapper typecmux set TS"
+            try:
+                self._run_control(cmd)
+            except Exception:
+                pass
         time.sleep(5)
         logger.info("Flashing Test image")
         try:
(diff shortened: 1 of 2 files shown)


plars (changes requested):
I like the general idea, but I'm wondering if it should be in muxpi or if this is something we should add to the zapper device connector that @p-gentili is working on?

I have a few notes below, but also...
1. This is a good opportunity to break out some pieces that can be easily tested - specifically, I would think about having a method that looks at the job and config to determine the right command to use and the right data (media) to send it. This way we can remove some of the conditionals here from the main path and put them into a method for which we can also easily write a unit test.
2. Please add documentation about the new parameter to docs/reference/device-connector-types.rst

device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
```diff
@@ -159,8 +161,28 @@ def provision(self):
...
-        cmd = self.config.get("control_switch_local_cmd", "stm -ts")
-        self._run_control(cmd)
+        if "media" not in self.job_data["provision_data"]:
+            media = None
+            self.test_device = self.config["test_device"]
+            cmd = self.config.get("control_switch_local_cmd", "stm -ts")
+            self._run_control(cmd)
+        else:
+            media = self.job_data["provision_data"]["media"]
+            match media:
```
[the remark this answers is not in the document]
rmartin013: Thanks for the feedback, I didn't know but I can easily update that

rmartin013:
> I like the general idea, but I'm wondering if it should be in muxpi or if this is something we should add to the zapper device connector that @p-gentili is working on?
> 
> I have a few notes below, but also...
> 
> 1. This is a good opportunity to break out some pieces that can be easily tested - specifically, I would think about having a method that looks at the job and config to determine the right command to use and the right data (media) to send it. This way we can remove some of the conditionals here from the main path and put them into a method for which we can also easily write a unit test.
> 2. Please add documentation about the new parameter to docs/reference/device-connector-types.rst

As I wrote on MM, I think the most effective plan was to use muxpi to (relatively) quickly provide a solution.
1. Unfortunately I don't see what you mean... Are you talking about IMAGE_PATH_IDS ? I guess this would be a subject for another PR anyway
2. Sure

mz2:
> I'm wondering if it should be in muxpi or if this is something we should add to the zapper device connector that @p-gentili is working on?

Not touching on the technical merits of the proposed code, but...

- We should not tie Remy's progress on upcoming work where schedule is an open question (this particular use of Zapper -- we are focused right now on provisioning, but immediate focus is on PC devices).
- Also, especially since the direction is away from muxpi to zapper used for provisioning, I also don't understand why we would hesitate on the basis of the Zapper connector coming up sometime soon? (i.e. we'll be moving away from this connector type, can replicate what we want in that Zapper based connector when it does become available, but it is not available now).

What helps at least me understand your concern on this better, Paul, is explaining concrete negative impact either in maintainability or some incident potential if this were to be merged.

device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
```diff
@@ -159,8 +161,27 @@ def provision(self):
...
+            if media == "sd":
+                self.test_device = "/dev/sd-disk"
+            elif media == "usb":
+                self.test_device = "/dev/tc-disk"
+            else:
+                self.test_device = self.config["test_device"]
+            # If media option is provided, then DUT is probably capable of
+            # booting from different media, we should switch both of them to TS
+            # side regardless of which one was previously used
+            cmd = "zapper sdwire set ZAPPER; zapper typecmux set TS"
```
plars, line 180: A couple of things I just noticed about this section:
1. calling two commands there with ; in the middle might result in something unexpected. If the last one fails, then the whole thing will fail. But if the first one fails, it will pass. You can use && instead I think to ensure that either command failing will fail properly
2. Do you really want to catch any exception here and ignore it? It seems like if the command to set up zapper or sdwire for provisioning failed, that that should probably be a fatal error right?

plars (changes requested):
Not sure if you were ready for this to be reviewed again or not, but I noticed one other thing that I added a comment on below.

device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
rmartin013: > A couple of things I just noticed about this section:
> 
> 1. calling two commands there with ; in the middle might result in something unexpected. If the last one fails, then the whole thing will fail. But if the first one fails, it will pass. You can use && instead I think to ensure that either command failing will fail properly
> 2. Do you really want to catch any exception here and ignore it? It seems like if the command to set up zapper or sdwire for provisioning failed, that that should probably be a fatal error right?

It's intentional indeed. This change can be useful even if the DUT's zapper is only connected to one of the 2 addons. Using the new parameter, you don't have to request and get a TF config change to replace the default stm commands, you can also easily switch from one to the other later on (that's typically what we did, switching from sd card to USB). So one of the 2 commands can fail legitimately. Even if the 2 of them fail, then the related /dev/xx-disk will not be detected, so the whole thing will fail anyway.

rmartin013:
> Not sure if you were ready for this to be reviewed again or not, but I noticed one other thing that I added a comment on below.

Thanks for the review Paul, yes it is ready. But I guess the final decision of integrating this change or not will be discussed next Monday.

device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
```diff
@@ -159,8 +161,28 @@ def provision(self):
...
+            self.test_device = self.config["test_device"]
+            cmd = self.config.get("control_switch_local_cmd", "stm -ts")
+            self._run_control(cmd)
+        else:
+            media = self.job_data["provision_data"]["media"]
+            match media:
+                case "sd":
+                    self.test_device = "/dev/sd-disk"
+                case "usb":
+                    self.test_device = "/dev/tc-disk"
```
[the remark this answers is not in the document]
p-gentili: [This is something I've recently implemented in Zapper](https://github.com/canonical/zapper/blob/7351643f097d498673375f2309be654ed325839e/zapper-control/zapper/provisioning/download.py#L61-L77), not exposed via API since it's part of the provisioning package, but it could be useful for the time being. Basically it hotplugs the device on the Zapper side while monitoring new device add events via udev.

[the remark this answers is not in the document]
rmartin013: Thanks for this comment, I didn't know we could use something like this, but that would be definitely better! What would be the ETA to get this?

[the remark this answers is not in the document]
p-gentili: Let's see how it goes https://github.com/canonical/zapper/pull/240

device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
```diff
@@ -149,7 +149,9 @@ def _copy_to_control(self, local_file, remote_file):
 
     def provision(self):
         # If this is not a zapper, reboot before provisioning
-        if "zapper" not in self.config.get("control_switch_local_cmd", ""):
```
plars, line 152: We needed to revert this part for now, so you'll want to rebase on current main. This section will likely conflict and need to be removed since we need to re-implement reboot_control_host() a different way.

plars (changes requested):
Not sure what the decision on monday is that you mentioned? but I did notice something that's changed that will likely require a very small bit of this to be removed. See the comment below

device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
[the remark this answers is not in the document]
p-gentili: The new API/CLI is ready and available in the latest Zapper snap.

device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
rmartin013: Thanks, rebased my branch and removed that commit.

docs/reference/device-connector-types.rst
```diff
@@ -117,6 +117,10 @@ The ``muxpi`` device connector supports the following ``provision_data`` keys:
        ``unzstd`` (**xz** format is recommended, but any format supported by
        the ``zstd`` tool is supported) and
        flashed to the SD card, which will be used to boot up the DUT.
+   * - ``media``
+     - Optional paramer to indicate on which boot media the disk image should
```
plars, line 121:
```suggestion
     - Optional parameter to indicate on which boot media the disk image should 
```

plars (changes requested):
One small typo, otherwise looks fine to me

docs/reference/device-connector-types.rst
rmartin013: Nice catch, sorry I missed it.

plars:
/canonical/self-hosted-runners/run-workflows 2baa3b21cd59d147b8a0f7f490ef73cb21f3e40d

plars:
@rmartin013 looks like it has a line in a comment on muxpy.py that is too long and the linter is complaining (recommend running tox before pushing), but also it looks like the linter for the docs is failing because of a broken link? It seems to work for me locally but failing on github somehow. If you can get these fixed up, then we should be able to land it.

rmartin013:
> @rmartin013 looks like it has a line in a comment on muxpy.py that is too long and the linter is complaining (recommend running tox before pushing), but also it looks like the linter for the docs is failing because of a broken link? It seems to work for me locally but failing on github somehow. If you can get these fixed up, then we should be able to land it.

Sorry, this time I forgot to run tox indeed... it's fixed now.
About the linter  for the docs, I don't know what to do with it, this is not a section of doc that I updated. I was wondering if that GH check used to pass before..?

plars:
/canonical/self-hosted-runners/run-workflows f610f09c77bc7f66f9fa50fda08154d8b4404f18

plars (approved):

plars:
hah, found it! Looks like it was some sort of cache issue that we can hit if only re-running part of the documentation linters. Definitely not something you affected here.

rmartin013:
Thanks Paul!

PR #229, opened 2024-03-19
Revert patches to reboot the control host before provisioning

plars:
Instability with rebooting the nanopi devices has been found where they don't reliably connect to the network after a soft reboot. I'm proposing this as a draft for review, in case we need to go ahead and revert this sooner than later. However it would be nice if we could find another way to deal with the instability that we previously saw where the nanopi devices would get slow network.


omar-selo (approved):
LGTM

PR #230, opened 2024-03-20
Add: provision OEM images via Zapper KVM connector

p-gentili:
## Description

_Draft until https://github.com/canonical/zapper/pull/239 gets merged (only minor problems, but some argument name might change)._

This PR introduces the support for OEM images via the Zapper KVM provisioning method.

For __22.04__ the procedure is split in two:
1. Zapper installs via KVM automations the `alloem` image which restores the backup partition and install a default Ubuntu 22.04 OEM image
2. If another `url` is passed, Testflinger will run the normal OemScript to install it on top of (1).

I preferred keeping (2) on the TF side to avoid code duplication and big downloads on Zapper "internal" storage (like the OemScript is doing).

On the other hand, __24.04__ is similar to provision since it's using `autoinstall` with a custom `user-data-base` and some additional kernel arguments (hence the new `cmdline_append` parameter). I haven't had the chance to test it since our main focus for now is 22.04 and this method is still under development by the OEM team.

## Resolved issues

Part of [ZAP-722](https://warthogs.atlassian.net/browse/ZAP-722)

## Documentation

Added a README which describes the available arguments, the supported scenarios and versions.

## Tests

Test are updated accordingly. Also, one can test it manually with the following configuration:

```
# config.yaml
device_ip: "10.102.163.33"
control_host: "10.102.243.33"
reboot_script: 
  - snmpset -c private -v2c 10.102.196.162 .1.3.6.1.4.1.13742.6.4.1.2.1.2.1.7 i 0
  - sleep 5
  - snmpset -c private -v2c 10.102.196.162 .1.3.6.1.4.1.13742.6.4.1.2.1.2.1.7 i 1
```

```
#job.json
{
"job_queue": "test",
"provision_data": {
  "alloem_url": "http://10.102.196.9/somerville/Platforms/jellyfish-edge-alloem-init/dell-bto-jammy-jellyfish-edge-alloem-init-X104-20230908-19.iso",
  "robot_tasks": [
    "hp/pro/sff_400_g9/boot/boot_from_usb.robot",
    "common/oem/jammy/alloem.robot",
    "hp/boot/wait_reboot.robot"
    ],
  "skip_download": true
},
"test_data": {
  "test_username": "ubuntu",
  "test_password": "insecure"
  }
}
```

[ZAP-722]: https://warthogs.atlassian.net/browse/ZAP-722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
@@ -62,6 +62,8 @@ class ZapperConnector(ABC, DefaultDevice):
         (api_args, api_kwargs) = self._validate_configuration()
         self._run(self.config["control_host"], *api_args, **api_kwargs)
 
+        self._post_run_actions(args)
+
         logger.info("END provision")
 
     @abstractmethod
@@ -99,3 +101,6 @@ class ZapperConnector(ABC, DefaultDevice):
             logger=logger,
             **kwargs,
         )
+
+    def _post_run_actions(self, args):
+        """Run further actions after Zapper API returns succesfully."""
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
@@ -103,4 +103,4 @@ class ZapperConnector(ABC, DefaultDevice):
         )
 
     def _post_run_actions(self, args):
-        """Run further actions after Zapper API returns succesfully."""
+        """Run further actions after Zapper API returns successfully."""
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/README.md
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/README.md
@@ -9,6 +9,8 @@ Support for vanilla Ubuntu is provided by [autoinstall](https://canonical-subiqu
 - Desktop >= 23.04
 - Server >= 20.04
 
+Unless specified via _autoinstall_ storage filter, the tool will select the largest storage device on the DUT. See [supported layouts](https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-reference.html#supported-layouts) for more information.
+
 ### Job parameters
 
 - __url__: URL to the image to install
@@ -28,6 +30,12 @@ Ubuntu OEM 22.04 is a two step process:
 
 1. Install using Zapper automations the `alloem` image which creates the reset partition on DUT and installs Ubuntu Jammy
 2. If URL is provided, run the OEM script to install an updated image on top of (1)
+
+The tool will select the storage device with the following priority:
+
+1. RAID
+2. NVME
+3. SATA
 
 #### Job parameters
 
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/README.md
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/README.md
@@ -1,0 +1,41 @@
+# Zapper KVM
+
+Zapper-driven provisioning method that makes use of KVM assertions and actions.
+
+## Ubuntu Desktop / Server
+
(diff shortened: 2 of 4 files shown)


device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
```diff
@@ -99,3 +101,6 @@ def _run(self, zapper_ip, *args, **kwargs):
             logger=logger,
             **kwargs,
         )
+
+    def _post_run_actions(self, args):
+        """Run further actions after Zapper API returns succesfully."""
```
nadzyah, line 106: Is it supposed to be an empty function?

p-gentili: Default implementation is an empty function yes, it's up to the implementor to eventually override it. I'm not marking it as abstract like the other one because it's not mandatory to implement it.

nadzyah (dismissed):
Don’t see anything to complain about, but probably you want to get an approval from @plars as well

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
```diff
@@ -50,18 +60,116 @@ def _validate_configuration(
...
+            "-p",
+            test_password,
+            "ssh-copy-id",
+            "-o",
+            "StrictHostKeyChecking=no",
+            "-o",
+            "UserKnownHostsFile=/dev/null",
+            f"{test_username}@{self.config['device_ip']}",
+        ]
+        subprocess.check_output(cmd, stderr=subprocess.STDOUT, timeout=60)
```
plars, line 153: Have you looked at these calls to check_output to ensure you are ok with the exceptions that could get raised? I'm finding it much nicer to call subprocess.run in a lot of cases (if possible), which still handling things like timeout. And either raising a more predictable exception, or even returning the return_code and output is often more useful than using check_output()

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Unless specified via _autoinstall_ storage filter, the tool will select the largest storage device on the DUT. See [supported layouts](https://canonical-subiquity.readthedocs-hosted.com/en/latest/reference/autoinstall-reference.html#supported-layouts) for more information.
+
+### Job parameters
+
+- __url__: URL to the image to install
+- __username__: username to configure
+- __password__: password to configure
+- **storage_layout**: can be either `lvm`, `direct`, `zfs` or `hybrid` (Desktop 23.10+)
+- **storage_password** (optional): only available with *storage_layout=lvm*, it's the password for LUKS encryption
```
plars, line 20: This one is slightly worrisome from an automation standpoint. Do we currently, or eventually have plans to add something that can automatically enter this when the system is rebooted?  I could see how zapper might have a chance at doing that, but if it's not going to, then maybe it's not an option we should provide because it could make the system unusable until someone reinstalls it.

plars (changes requested):
A couple of questions below. In particular I think you might want to go ahead and do something better with the check_output() things unless I'm missing something. Otherwise you won't get much information if they fail

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
p-gentili: In this case I tend to prefer `check_output` since I care about the result only when it fails. I wasn't handling at all the exceptions though, thanks for noticing. Updated with proper logging and exception.

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/README.md
p-gentili: Good point, we're not there yet.

plars (approved):
+1

PR #233, opened 2024-03-22
Implement zapper_iot method

kiya956:

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
@@ -18,7 +18,7 @@ from typing import Any, Dict, Tuple
 from testflinger_device_connectors.devices.zapper import ZapperConnector
 from testflinger_device_connectors.devices import ProvisioningError
 from testflinger_device_connectors.devices.zapper_iot.parser import (
-    validate_data,
+    validate_tplan,
     validate_url,
 )
 
@@ -39,7 +39,7 @@ class DeviceConnector(ZapperConnector):
         """
         try:
             tplan = self.job_data["provision_data"]["iot_provision"]["tplan"]
-            validate_data(tplan)
+            validate_tplan(tplan)
         except KeyError as e:
             raise ProvisioningError from e
 
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
@@ -13,10 +13,16 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """Zapper Connector for IOT provisioning."""
-
+import logging
 from typing import Any, Dict, Tuple
-
 from testflinger_device_connectors.devices.zapper import ZapperConnector
+from testflinger_device_connectors.devices import ProvisioningError
+from testflinger_device_connectors.devices.zapper_iot.parser import (
+    validate_data,
+    validate_url,
+)
+
+logger = logging.getLogger(__name__)
 
 
 class DeviceConnector(ZapperConnector):
@@ -31,4 +37,16 @@ class DeviceConnector(ZapperConnector):
         Validate the job config and data and prepare the arguments
         for the Zapper `provision` API.
         """
-        raise NotImplementedError
+        try:
+            tplan = self.job_data["provision_data"]["iot_provision"]["tplan"]
+            validate_data(tplan)
+        except KeyError as e:
+            raise ProvisioningError from e
+
+        try:
+            url = self.job_data["provision_data"]["iot_provision"]["url"]
+            validate_url(url)
+        except KeyError:
+            url = []
+
+        return ((tplan, url), {})
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/parser.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/parser.py
@@ -1,0 +1,198 @@
+"""provision yaml verify"""
+
+import logging
+import validators
+import jsonschema
+from jsonschema import validate
+
+logger = logging.getLogger(__name__)
+
+LAUNCHER_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "config": {
+            "type": "object",
+            "properties": {
+                "project_name": {"type": "string"},
+                "username": {"type": "string"},
+                "password": {"type": "string"},
+                "serial_console": {
+                    "type": "object",
+                    "properties": {
+                        "port": {"type": "string"},
+                        "baud_rate": {
+                            "type": "integer",
+                            "enum": [115200, 9600],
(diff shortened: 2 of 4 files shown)


device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -13,22 +13,84 @@
...
+    def provision(self, args):
+        """Method called when the command is invoked."""
+        with open(args.config, encoding="utf-8") as configfile:
+            self.config = yaml.safe_load(configfile)
+        with open(args.job_data, encoding="utf-8") as job_json:
+            self.job_data = json.load(job_json)
 
-    def _validate_configuration(
-        self,
-    ) -> Tuple[Tuple, Dict[str, Any]]:
```
nadzyah, line 29: is there any reason why typing hints were removed?

kiya956: Sorry I didn't notice that, will fix it later.

AristoChen (approved):
LGTM

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
kiya956: @nadzyah Hi Nadzyah
I have updated the changes, could you help to review is there any inappropriate 
Thank you

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -13,22 +13,91 @@
...
+        try:
+            url = job_data["provision_data"]["iot_provision"]["url"]
+            validate_url(url)
+        except KeyError:
+            url = []
+
+        return tplan, url
+
+    # pylint: disable=W0221
+    def _run(self, zapper_ip, tplan, url):
```
p-gentili, line 80: You shouldn't override this one, it's provided by `ZapperConnector`.

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -13,22 +13,91 @@
...
 
     PROVISION_METHOD = "ProvisioningIoT"
+    ZAPPER_REQUEST_TIMEOUT = 60 * 70
+    ZAPPER_SERVICE_PORT = 60000
+
+    def __init__(self):
+        self.config = None
+        self.job_data = None
+
+    def provision(self, args):
```
p-gentili, line 42: You shouldn't override this one, it's provided by `ZapperConnector`.

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -13,22 +13,91 @@
...
 
 
 class DeviceConnector(ZapperConnector):
     """Tool for provisioning baremetal with a given image."""
 
     PROVISION_METHOD = "ProvisioningIoT"
+    ZAPPER_REQUEST_TIMEOUT = 60 * 70
+    ZAPPER_SERVICE_PORT = 60000
+
+    def __init__(self):
```
p-gentili, line 38: You shouldn't override this one, it's provided by `ZapperConnector`.

p-gentili (changes requested):
Hey @kiya956 thanks for this. You shouldn't override many of the methods you're proposing here. The default implementation is provided by the base class and it should be maintained.

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -13,22 +13,91 @@
...
+
+logger = logging.getLogger(__name__)
 
 
 class DeviceConnector(ZapperConnector):
     """Tool for provisioning baremetal with a given image."""
 
     PROVISION_METHOD = "ProvisioningIoT"
+    ZAPPER_REQUEST_TIMEOUT = 60 * 70
+    ZAPPER_SERVICE_PORT = 60000
```
p-gentili, line 36: Same for this one, not needed because you're not varying from the base class.

kiya956:
> Thanks @kiya956!, Another thing: I noticed you're not pulling anything from `config` but the `device_ip` will come from there (C3 actually). Is it something you don't need?

@p-gentili Hi Paolo, due to our tool can get device ip directly from target device through serial, so we don't need device_ip. Thanks.

p-gentili (approved):
LGTM, thanks!

Just one question for @plars before merging, are you ok with the added dependencies?

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/parser.py
```diff
@@ -0,0 +1,198 @@
...
+    try:
+        validate(instance=data, schema=LAUNCHER_SCHEMA)
+        logger.info("the JSON data is valid")
+    except jsonschema.exceptions.ValidationError as err:
+        raise ValueError("the JSON data is invalid") from err
+
+
+def validate_url(url):
+    """for verify url"""
+    for lnk in url:
```
plars, line 196: I guess "lnk" is link? If so, let's write that out in the name of the variable - shortening it by one character doesn't help here

Do we know, for certain, that url is iterable? If we get a string there, then we're going to try to call that validator on every character, which is going to produce some really confusing errors. Should we confirm that it's a list or some other expected type first?

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/parser.py
```diff
@@ -0,0 +1,198 @@
...
+            ],
+        },
+    },
+}
+
+
+def validate_data(data):
+    """for verify provision yaml"""
+    try:
+        validate(instance=data, schema=LAUNCHER_SCHEMA)
```
plars, line 188: Since this seems to be looking at the "tplan" key from the data that we were passed, it seems to me like it might help if we:
1. name the function validate_tplan
2. rename LAUNCHER_SCHEMA to TPLAN_SCHEMA

I think it's ok to have this one schema here if that's the only one, but if we want to add others, then I would recommend at least putting them in a schemas module, and importing them from there. That way you can just focus on the code here.

plars (changes requested):
A few minor comments below, but I think I see what you're trying to do here.

In the future, it would be really helpful to at least provide a quick summary in the PR that says what you are trying to accomplish. No need to write anything too detailed or long in most cases, but providing hints about the intent would be helpful :)

plars (changes requested):
Also, as we add things like device connectors and extra options that can be provided to device connectors - could you please also add documentation under the /docs dir for them as appropriate. At the very least, these additional provisioning options need to be in the reference/device-connector-conf.rst file and the device connector itself should be listed in reference/device-connector-types.rst

plars (approved):
I'm assuming you are ready to merge this? It looks like you fixed the things I mentioned. Let's get it merged asap since I know you are trying to iterate on this still

PR #237, opened 2024-03-25
Temporarily disable output timeout checking until we have a better fix

plars:
This to work around an issue with how we do output timeout checking for now, until we have a better way.

When I originally refactored this, the output timeout checking was on another thread, but it was a bit hairy. That was one of the pieces I was trying to clean up, so I ended up consolidating it as a "stop_condition_checker" that we could register, but I missed the part where it needs to have awareness of the latest time new output was seen.
Will look into whether output timeout checking makes sense really, or at least providing a better fix.
But for now, it's very likely this could cause things to fail prematurely, so we need to address that immediately


nadzyah (dismissed):

nadzyah (approved):

omar-selo (approved):
LGTM

PR #242, opened 2024-04-02
Add the jammy tag to all self-hosted runner invocations

mthaddon:
## Description

Update all workflows using self-hosted runners to include the jammy tag. This is in preparation for the availability of noble runners


jocave (approved):
Thanks, +1

PR #243, opened 2024-04-02
All queues view

plars:
## Description
- Make the queues view include all known queues, not just advertised_queues
- queues will still have a description if they are also advertised
- some updates to improve testability

## Resolved issues
Resolves part of CERTTF-298

## Documentation
N/A

## Tests
Unit tests added for the function that generates data for the queues view


server/src/application.py
```diff
@@ -0,0 +1,126 @@
...
+        mongo_uri = (
+            f"mongodb://{mongo_creds}{mongo_host}:{mongo_port}/{mongo_db}"
+            f"?authSource={mongo_auth}"
+        )
+
+    mongo.init_app(
+        application,
+        uri=mongo_uri,
+        uuidRepresentation="standard",
+        serverSelectionTimeoutMS=2000,
```
nadzyah, line 122: What do you think about making this a configurable parameter?

server/tests/test_views.py
```diff
@@ -0,0 +1,77 @@
...
+
+import mongomock
+from mock import patch
+from src.views import queues_data
+
+
+def test_queues():
+    """
+    Test that the queues view gets the right data from both advertised and
+    unadvertised queues
```
omar-selo, line 28: I don't fully understand what's the difference between advertised and unadvertised queues. But reading the docstring here suggests that there should be unadvertised queues too. Yet below I see you're only inserting advertised queues

omar-selo (commented):
Interesting approach to testing with mongo in flask. The only issue is that every endpoint has to be written this way (separate view out as a function) and every test has to include that particular patch (could get hairy when we split views into files or when mongo client starts being used from different locations). It would be much cleaner if somehow we didn't have to do any of the above. A related question is whether we have to mock mongo at all? as in why not use pymongo and just direct it to a test database instead (each test case would have a setup and teardown phase that empties the db). On TO we take that approach and we inject the db client into our fastapi client [here](https://github.com/canonical/test_observer/blob/main/backend/tests/conftest.py#L90). If flask has no such dependency injection mechanism could we instead modify `setup_mongodb` or `create_flask_app` to get a similar result. wdyt?

server/src/application.py
plars: we could, but the default is 30 and we had to set it very high because of some PS5 issue IIRC and haven't had to touch it since then. It's not new - just the file is moving to a new name, so that's why you're seeing it now. I think I'd prefer to defer that to a later change if it becomes something we really need to have a knob for adjusting frequently, otherwise we had previously been advised to keep the number of configuration variables limited if possible.

server/tests/test_views.py
plars: So the agents have options that let them "advertise" queues with a name and description. It was easier to keep up with when we didn't have so many agents, but still kind of exists, and otherwise, the server didn't even have any way previously to know about whether a queue exists or not. There's still no concept of "registering" a queue with the server or anything like that, because it's completely opposite of how we built this. The server doesn't *tell* the queues to run a job, the agents declare which queues they know about and ask the server for a job matching one of those tags. But now, the agents also tell the server which ones they are listening to when they start up along with some other data like provisioning type, so that we can inform other systems about this.

...that's a lot of background, but this agent data is where we get the other queues from. It's the second query you see there to retrieve that. Then we make sure not to overwrite any of the ones from advertised queues, because those actually have a description. This was all for a request to make sure that all queues were visible from the UI on this view, and not just the advertised ones.

plars: As for the other part of the question, flask doesn't have a native dependency injection system like fastAPI does. There are some 3rd party ones, but I haven't really looked into those yet. I would much rather just redo all this in fastAPI actually - for a host of other reasons, but that's one of them :)

As for the question about mocking vs. using a real mongodb server - that was because these are unit tests and having them run quickly as unit tests from something like pytest means that I can't depend on there being a mongodb server running. The mock just provides the functionality that it would provide if it were there, while also allowing us to inject or read back data from it if needed.

The other question here was about potential for causing problems if we split the views into more files? I'm not sure I follow... In this case, I suppose I could have done something like beautiful soup to try to parse the html and make sure it matched what I wanted, but I think it would be pretty hard to maintain and read those. What I'm much more interested in is whether we are sending the right data to be rendered. I'm not sure how splitting the file would affect that. If we want to split the data gathering part into a different file from the html-rendering part of the view, I think we could do that. I don't see how that would make this harder but I'm happy to take a look if you have more details?

omar-selo: Thanks for the very detailed response. If I understand things correctly, now advertised queues and unadvertised queues have become closer as they both are shown on the server and UI. Perhaps we'll some day merge them completely.

I see the benefit of mocking mongo, it really does simplify things as there's no need to deploy a database to run tests.

Regarding the point on splitting the views, I should've been clearer. I was worried about patching the mongo instance as we'd have to patch different places. But this wouldn't be an issue in the future if we move to FastAPI and use dependency injection.

omar-selo (approved):
Thanks for your detailed response! LGTM

plars:
> Thanks for your detailed response! LGTM
Thanks for the review, great questions!

PR #245, opened 2024-04-03
Add docs about agents and queues to the explanation section

plars:
## Description
There are certainly more things we should add here, but this gives us a start on the explanation section of the documentation (which previously didn't have anything yet). We can build on that from here.

## Resolved issues
Also, these specific sections were requested in CERTTF-298

## Documentation
Yes

## Tests
Passed make linkcheck for me locally

### DIFF
--- a/docs/explanation/agents.rst
+++ b/docs/explanation/agents.rst
@@ -1,0 +1,11 @@
+Testflinger Agents
+==================
+
+The Testflinger agent is a per-machine service that runs on an agent host
+system. The agent is configured with a set of :doc:`queues <queues>` for which
+it knows how to run jobs. When it is not running a job, the agent:
+
+   * Asks the server for a job to run from the list of configured :doc:`queues <queues>`
+   * Dispatches the device connector to execute each :doc:`phase <../reference/test-phases>` of the job
+   * Reports the results of the job back to the server
+   * Uploads artifacts (if any) saved from the job to the server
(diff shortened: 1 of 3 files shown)


plars:
@tang-mm do you know what is causing this link-check failure? I saw this earlier today on someone else's PR, and figured out that I could reproduce it locally if I run make linkcheck and then run it again a second time. But I was able to make it pass on that other PR by rerunning all jobs on the github action rather than just the failed one.

I tried that here also, but it didn't work. I didn't chanage any of the things that it seems to be complaining about, and it works when I try to run it locally.

tang-mm:
> @tang-mm do you know what is causing this link-check failure? I saw this earlier today on someone else's PR, and figured out that I could reproduce it locally if I run make linkcheck and then run it again a second time. But I was able to make it pass on that other PR by rerunning all jobs on the github action rather than just the failed one.
> 
> I tried that here also, but it didn't work. I didn't chanage any of the things that it seems to be complaining about, and it works when I try to run it locally.

@plars The failure in link checking was caused by the anchors in URLs that already exist in `main`. I updated the doc config file previously to ignore them, which made the original PR build successful but somehow no longer works. I added a fix in https://github.com/canonical/testflinger/pull/248, hopefully, it will solve the issue once merged/rebased

tang-mm (approved):

PR #246, opened 2024-04-04
Fix an unlikely corner-case where an agent has no queues declared

plars:
## Description
Strange thing I hit when pushing this to staging. For some reason, there was an agent that had no queues associated with it there! That's unlikely, and wouldn't be very useful, but it's obviously possible and we apparently had one there that fit this profile.
This accounts for that possibility and skips it appropriately, while also adding a test for this corner-case

## Tests
Updated the unit tests with something that triggered this corner case


omar-selo (approved):
woah staging comes for the rescue

PR #232, opened 2024-03-21
Add more details to the reference section about provision_data

plars:
## Description
- move the maas storage custom config documentation out to the main documentation directory
- reformat the maas storage custom config docs and link to TOC
- add per-device-connector documentation for all the support provision_data options we currently support

## Resolved issues
Resolves one part of CERTTF-298

## Documentation
yes

## Tests
Built locally using `make html` from the docs directory

PR #248, opened 2024-04-08
Fix documenation checking on link anchors

tang-mm:
## Description

Update `custom_conf.py` to ignore anchors when running link checking on MAAS docs

(The initial config was introduced by https://github.com/canonical/testflinger/pull/232 but no longer works.)

## Resolved issues

N/A

## Documentation

Updated README to include customized docs checking parameters

## Tests

Automatic doc checks


plars (approved):
Nice, thanks for this fix!

PR #250, opened 2024-04-10
[CERTTF-303] Introduce support for file attachments

boukeas:
## Description

This PR introduces functionality that allows a testflinger client (e.g. the testflinger CLI) to submit a job that includes attachments, with these attachments eventually finding their way into the testflinger agent.

## Resolved issues

[CERTTF-303](https://warthogs.atlassian.net/browse/CERTTF-303)

- Removes the need to create configuration files as [here documents](https://tldp.org/LDP/abs/html/here-docs.html) within the `test_cmds` field; the configuration files can be attached instead.
- Bash scripts can be attached and used in the `test_cmds` field
- Attached files can also be useful in the `provisioning` or `firmware_update` phases
- Will serve as a step for addressing [RTW-282](https://warthogs.atlassian.net/browse/RTW-282)

### Server

- Extend the `Job` submission schema so that `attachments` fields are allowed under `test_data`, specifying which local files should be packed into an archive and submitted along with a job. For example:
```
test_data:
  attachments:
    - local: "/var/log/syslog"
      agent: "logs/syslog"
    - local: "data/config.json"
    - local: "data/scripts/script.sh"
      agent: "test.sh"
```
- Such fields are also allowed under the provisioning and firmware update data but the current schema doesn't validate against them because too many arbitrary fields are used in these phases, depending on the use case. This validation issue is not relevant to this PR.

- In the implementation of the endpoint for submitting a job, a new _derived_ `attachments` status field is stored in the job data, taking the values:

  - `none` when there are no attachments to the job
  - `waiting` when there are attachments but the corresponding archive file hasn't been submitted yet and
  - `complete` when there are attachments and the corresponding archive file has been submitted

  Only the values `none` or `complete` allow a job to be removed from a queue.

- Introduce two endpoints for submitting and retrieving an archive file that contains the attachments for a specific job. The submission of an attachments archive sets the value of the `attachments` status field for the job to `complete`.

- Move database-related code into a separate `database` module, so that the API implementation is database-agnostic. This has been partly realised: the database interactions _that are relevant to this PR_ have been moved into reusable functions of the `database` module. If approved, this change can also be applied to the implementation of all API endpoints, eventually removing all references to MongoDB from the API implementation _and_ making the database an injectable dependency into the app.

### CLI

The CLI checks if the job submission includes references to attachments in the provisioning, firmware and test phases and, if so, packs the attachments into an archive and follows the job submission with an additional attachments archive submission to the server (using the new endpoint described above)

### Agent

The agent checks if the job submission is associated with attachments and, if so, retrieves the attachments archive from the server, unpacks it and optionally moves/renames the attachments. Note that all attachments are placed under an `attachments` folder in the agents "rundir". They are additionally separated into phase-related folders, e.g. `attachments/test` for the attachments provided under `test_data`.

## Points for discussion
- The intention is that the current functionality will be later extended to attaching files that are intended to end up in the device (rather than the host).
- Efficiency concerns: how will the server behave when large files are submitted and retrieved? This is the main reason that the job submission still only includes the JSON, without the attachments archive (as a multipart-encoded request). If the job submission was bundled with a large attachments archive then it would take much longer to complete. Attachment archives are submitted and retrieved separately. In the future, this might also allow for asynchronous transfers.
- There is an evident need for tooling modules that are common among the CLI, the server and the agent. There is code that would be (re)usable and is currently isolated and repeated.
- How to test in a non-local staging environment

## Documentation

Currently only inline documentation is provided. If the approach is approved the documentation will be extended under a separate issue/PR.

## Tests

- The test suite has been extended on the CLI, server and agent sides to provide new unit tests for the new functionality. Some additional tests should be added to test for negative scenarios, i.e. behaviour under unexpected circumstances.
- Manual tests have been performed in a local testflinger deployment, using this job YAML, to verify that the specified files do indeed find their way to the agent, under the specified names.

```
job_queue: myqueue
provision_data:
  attachments:
    - local: "README.md"
firmware_update_data:
  attachments:
    - local: "CONTRIBUTING.md"
test_data:
  attachments:
    - local: "/var/log/syslog"
      agent: "logs/syslog"
    - local: "data/config.json"
    - local: "data/scripts/script.sh"
      agent: "test.sh"
  test_cmds: |
    ls -alR
    echo "*** These are the contents of the script file"
    cat attachments/test/test.sh
    echo "*** Running the script file"
    echo "(it's supposed to print the contents of the config file)"
    chmod u+x attachments/test/test.sh
    attachments/test/test.sh attachments/test/data/config.json
```

[CERTTF-303]: https://warthogs.atlassian.net/browse/CERTTF-303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RTW-282]: https://warthogs.atlassian.net/browse/RTW-282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -123,12 +123,13 @@ class TestflingerAgent:
         """Download and unpack the attachments associated with a job"""
 
         # download attachment archive to a unique temporary folder
+        job_id = job_data["job_id"]
         archive_dir = tmp_dir()
-        archive_path = self.client.get_attachments(
-            job_id=job_data["job_id"], path=archive_dir
-        )
+        archive_path = self.client.get_attachments(job_id, path=archive_dir)
         if archive_path is None:
-            raise FileNotFoundError
+            raise FileNotFoundError(
+                f"Unable to retrieve attachments for job {job_id}"
+            )
         # extract archive data to a unique temporary folder and clean up
         extracted_dir = tmp_dir()
         with tarfile.open(archive_path, "r:gz") as tar:
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -122,46 +122,43 @@ class TestflingerAgent:
 
     def unpack_attachments(self, job_data: dict, cwd: Path):
         """Download and unpack the attachments associated with a job"""
-
-        # download attachment archive to a unique temporary folder
         job_id = job_data["job_id"]
-        archive_dir = tmp_dir()
-        archive_path = self.client.get_attachments(job_id, path=archive_dir)
-        if archive_path is None:
-            raise FileNotFoundError(
-                f"Unable to retrieve attachments for job {job_id}"
-            )
-        # extract archive data to a unique temporary folder and clean up
-        extracted_dir = tmp_dir()
-        with tarfile.open(archive_path, "r:gz") as tar:
-            tar.extractall(extracted_dir, filter="data")
-        shutil.rmtree(archive_dir)
 
-        attachment_dir = cwd / ATTACHMENTS_DIR
-
-        # move/rename extracted archive files to their specified destinations
-        for phase in ("provision", "firmware_update", "test"):
-            try:
-                attachments = job_data[f"{phase}_data"]["attachments"]
-            except KeyError:
-                continue
-            for attachment in attachments:
-                original = Path(attachment["local"])
-                if original.is_absolute():
-                    # absolute filenames become relative
-                    original = original.relative_to(original.root)
-                # use renaming destination, if provided
-                # otherwise use the original one
-                destination_path = (
-                    attachment_dir / phase / attachment.get("agent", original)
-                )
-                # create intermediate path to destination, if required
-                destination_path.resolve().parent.mkdir(
-                    parents=True, exist_ok=True
-                )
-                # move file
-                source_path = extracted_dir / phase / original
-                shutil.move(source_path, destination_path)
+        with tempfile.NamedTemporaryFile(suffix="tar.gz") as archive_tmp:
+            archive_path = Path(archive_tmp.name)
+            # download attachment archive
+            self.client.get_attachments(job_id, path=archive_path)
+            with tempfile.TemporaryDirectory() as extracted_tmp:
+                extracted_dir = Path(extracted_tmp)
+                # extract archive into a temporary folder
+                with tarfile.open(archive_path, "r:gz") as tar:
+                    tar.extractall(extracted_dir, filter="data")
+                # move/rename extracted archive files to their destinations
+                attachment_dir = cwd / ATTACHMENTS_DIR
+                for phase in ("provision", "firmware_update", "test"):
+                    try:
+                        attachments = job_data[f"{phase}_data"]["attachments"]
+                    except KeyError:
+                        continue
+                    for attachment in attachments:
+                        original = Path(attachment["local"])
+                        if original.is_absolute():
+                            # absolute filenames become relative
+                            original = original.relative_to(original.root)
+                        # use renaming destination, if provided
+                        # otherwise use the original one
+                        destination_path = (
+                            attachment_dir
+                            / phase
+                            / attachment.get("agent", original)
+                        )
+                        # create intermediate path to destination, if required
+                        destination_path.resolve().parent.mkdir(
+                            parents=True, exist_ok=True
+                        )
+                        # move file
+                        source_path = extracted_dir / phase / original
+                        shutil.move(source_path, destination_path)
 
     def process_jobs(self):
         """Coordinate checking for new jobs and handling them if they exists"""
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -144,7 +144,7 @@ class TestflingerAgent:
         attachment_dir = cwd / "attachments"
 
         # move/rename extracted archive files to their specified destinations
-        for phase in ["provision", "firmware_update", "test"]:
+        for phase in ("provision", "firmware_update", "test"):
             try:
                 attachments = job_data[f"{phase}_data"]["attachments"]
             except KeyError:
@@ -159,11 +159,10 @@ class TestflingerAgent:
                 destination_path = (
                     attachment_dir / phase / attachment.get("agent", original)
                 )
-                try:
-                    # create intermediate path to destination, if required
-                    destination_path.resolve().parent.mkdir(parents=True)
-                except FileExistsError:
-                    pass
+                # create intermediate path to destination, if required
+                destination_path.resolve().parent.mkdir(
+                    parents=True, exist_ok=True
+                )
                 # move file
                 source_path = extracted_dir / phase / original
                 shutil.move(source_path, destination_path)
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -22,6 +22,7 @@ import tempfile
 
 from testflinger_agent.job import TestflingerJob
 from testflinger_agent.errors import TFServerError
+from testflinger_agent.config import ATTACHMENTS_DIR
 
 logger = logging.getLogger(__name__)
 
@@ -136,8 +137,7 @@ class TestflingerAgent:
             tar.extractall(extracted_dir, filter="data")
         shutil.rmtree(archive_dir)
 
-        # [TODO] clarify if this is an appropriate destination for extraction
-        attachment_dir = cwd / "attachments"
+        attachment_dir = cwd / ATTACHMENTS_DIR
 
         # move/rename extracted archive files to their specified destinations
         for phase in ("provision", "firmware_update", "test"):
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -31,11 +31,6 @@ def tmp_dir() -> Path:
     return Path(tempfile.mkdtemp())
 
 
-def secure_filter(member, path):
-    """Combine the `data` and `tar` filter from `tarfile`"""
-    return tarfile.tar_filter(tarfile.data_filter(member, path), path)
-
-
 class TestflingerAgent:
     def __init__(self, client):
         self.client = client
@@ -137,7 +132,7 @@ class TestflingerAgent:
         # extract archive data to a unique temporary folder and clean up
         extracted_dir = tmp_dir()
         with tarfile.open(archive_path, "r:gz") as tar:
-            tar.extractall(extracted_dir, filter=secure_filter)
+            tar.extractall(extracted_dir, filter="data")
         shutil.rmtree(archive_dir)
 
         # [TODO] clarify if this is an appropriate destination for extraction
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -35,7 +35,15 @@ def secure_filter(member, path):
     has been created properly and no attachment will be extracted to an
     unexpected location.
     """
-    if not member.name.startswith(("provision", "firmware_update", "test")):
+    try:
+        resolved = Path(member.name).resolve().relative_to(Path.cwd())
+    except ValueError as error:
+        # essentially trying to extract higher than the attachments folder
+        raise tarfile.OutsideDestinationError(member, path) from error
+    if not str(resolved).startswith(
+        ("provision/", "firmware_update/", "test/")
+    ):
+        # trying to extract in an invalid folder, under the attachments folder
         raise tarfile.OutsideDestinationError(member, path)
     return tarfile.data_filter(member, path)
 
@@ -168,10 +176,6 @@ class TestflingerAgent:
 
                 self.client.post_agent_data({"job_id": job.job_id})
 
-                # handle job attachments, if any
-                if job_data.get("attachments", "none") == "complete":
-                    self.unpack_attachments(job_data, cwd=Path(rundir))
-
                 # Dump the job data to testflinger.json in our execution dir
                 with open(os.path.join(rundir, "testflinger.json"), "w") as f:
                     json.dump(job_data, f)
@@ -180,6 +184,12 @@ class TestflingerAgent:
                     os.path.join(rundir, "testflinger-outcome.json"), "w"
                 ) as f:
                     json.dump({}, f)
+
+                # handle job attachments, if any
+                # (always after creating "testflinger.json", for reporting
+                # in case of an unpacking error)
+                if job_data.get("attachments", "none") == "complete":
+                    self.unpack_attachments(job_data, cwd=Path(rundir))
 
                 for phase in TEST_PHASES:
                     # First make sure the job hasn't been cancelled
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -27,9 +27,17 @@ from testflinger_agent.config import ATTACHMENTS_DIR
 logger = logging.getLogger(__name__)
 
 
-def tmp_dir() -> Path:
-    """Create a temporary directory and return the path to it"""
-    return Path(tempfile.mkdtemp())
+def secure_filter(member, path):
+    """Combine the `data` filter with custom attachment filtering
+
+    Makes sure that the starting folder for all attachments coincides
+    with one of the supported phases, i.e. that the attachment archive
+    has been created properly and no attachment will be extracted to an
+    unexpected location.
+    """
+    if not member.name.startswith(("provision", "firmware_update", "test")):
+        raise tarfile.OutsideDestinationError(member, path)
+    return tarfile.data_filter(member, path)
 
 
 class TestflingerAgent:
@@ -128,37 +136,9 @@ class TestflingerAgent:
             archive_path = Path(archive_tmp.name)
             # download attachment archive
             self.client.get_attachments(job_id, path=archive_path)
-            with tempfile.TemporaryDirectory() as extracted_tmp:
-                extracted_dir = Path(extracted_tmp)
-                # extract archive into a temporary folder
-                with tarfile.open(archive_path, "r:gz") as tar:
-                    tar.extractall(extracted_dir, filter="data")
-                # move/rename extracted archive files to their destinations
-                attachment_dir = cwd / ATTACHMENTS_DIR
-                for phase in ("provision", "firmware_update", "test"):
-                    try:
-                        attachments = job_data[f"{phase}_data"]["attachments"]
-                    except KeyError:
-                        continue
-                    for attachment in attachments:
-                        original = Path(attachment["local"])
-                        if original.is_absolute():
-                            # absolute filenames become relative
-                            original = original.relative_to(original.root)
-                        # use renaming destination, if provided
-                        # otherwise use the original one
-                        destination_path = (
-                            attachment_dir
-                            / phase
-                            / attachment.get("agent", original)
-                        )
-                        # create intermediate path to destination, if required
-                        destination_path.resolve().parent.mkdir(
-                            parents=True, exist_ok=True
-                        )
-                        # move file
-                        source_path = extracted_dir / phase / original
-                        shutil.move(source_path, destination_path)
+            # extract archive into the attachments folder
+            with tarfile.open(archive_path, "r:gz") as tar:
+                tar.extractall(cwd / ATTACHMENTS_DIR, filter=secure_filter)
 
     def process_jobs(self):
         """Coordinate checking for new jobs and handling them if they exists"""
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -15,12 +15,25 @@
 import json
 import logging
 import os
+from pathlib import Path
 import shutil
+import tarfile
+import tempfile
 
 from testflinger_agent.job import TestflingerJob
 from testflinger_agent.errors import TFServerError
 
 logger = logging.getLogger(__name__)
+
+
+def tmp_dir() -> Path:
+    """Create a temporary directory and return the path to it"""
+    return Path(tempfile.mkdtemp())
+
+
+def secure_filter(member, path):
+    """Combine the `data` and `tar` filter from `tarfile`"""
+    return tarfile.tar_filter(tarfile.data_filter(member, path), path)
 
 
 class TestflingerAgent:
@@ -111,6 +124,50 @@ class TestflingerAgent:
         # Create the offline file, this should work even if it exists
         open(self.get_offline_files()[0], "w").close()
 
+    def unpack_attachments(self, job_data: dict, cwd: Path):
+        """Download and unpack the attachments associated with a job"""
+
+        # download attachment archive to a unique temporary folder
+        archive_dir = tmp_dir()
+        archive_path = self.client.get_attachments(
+            job_id=job_data["job_id"], path=archive_dir
+        )
+        if archive_path is None:
+            raise FileNotFoundError
+        # extract archive data to a unique temporary folder and clean up
+        extracted_dir = tmp_dir()
+        with tarfile.open(archive_path, "r:gz") as tar:
+            tar.extractall(extracted_dir, filter=secure_filter)
+        shutil.rmtree(archive_dir)
+
+        # [TODO] clarify if this is an appropriate destination for extraction
+        attachment_dir = cwd / "attachments"
+
+        # move/rename extracted archive files to their specified destinations
+        for phase in ["provision", "firmware_update", "test"]:
+            try:
+                attachments = job_data[f"{phase}_data"]["attachments"]
+            except KeyError:
+                continue
+            for attachment in attachments:
+                original = Path(attachment["local"])
+                if original.is_absolute():
+                    # absolute filenames become relative
+                    original = original.relative_to(original.root)
+                # use renaming destination, if provided
+                # otherwise use the original one
+                destination_path = (
+                    attachment_dir / phase / attachment.get("agent", original)
+                )
+                try:
+                    # create intermediate path to destination, if required
+                    destination_path.resolve().parent.mkdir(parents=True)
+                except FileExistsError:
+                    pass
+                # move file
+                source_path = extracted_dir / phase / original
+                shutil.move(source_path, destination_path)
+
     def process_jobs(self):
         """Coordinate checking for new jobs and handling them if they exists"""
         TEST_PHASES = [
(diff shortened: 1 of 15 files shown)


agent/testflinger_agent/agent.py
```diff
@@ -111,6 +124,50 @@ def mark_device_offline(self):
...
+    def unpack_attachments(self, job_data: dict, cwd: Path):
+        """Download and unpack the attachments associated with a job"""
+
+        # download attachment archive to a unique temporary folder
+        archive_dir = tmp_dir()
+        archive_path = self.client.get_attachments(
+            job_id=job_data["job_id"], path=archive_dir
+        )
+        if archive_path is None:
+            raise FileNotFoundError
```
nadzyah, line 136: Could you please also add any logging or exception message here to include the path to the file that is not found? I guess it would be helpful for a user to see what caused the failure

agent/testflinger_agent/agent.py
```diff
@@ -111,6 +124,50 @@ def mark_device_offline(self):
...
+        # extract archive data to a unique temporary folder and clean up
+        extracted_dir = tmp_dir()
+        with tarfile.open(archive_path, "r:gz") as tar:
+            tar.extractall(extracted_dir, filter=secure_filter)
+        shutil.rmtree(archive_dir)
+
+        # [TODO] clarify if this is an appropriate destination for extraction
+        attachment_dir = cwd / "attachments"
```
nadzyah, line 144: I suggest to move this string to the file with constants

agent/testflinger_agent/client.py
```diff
@@ -105,6 +106,30 @@ def check_jobs(self):
...
+        :param path:
+            Where to save the attachment archive. If it is a folder,
+            the default filename `attachments.tar.gz` is used.
+        :returns path or None:
+            Where the attachment archive was saved or `None` id nothing
+            was retrieved
+        """
+        uri = urljoin(self.server, f"/v1/job/{job_id}/attachments")
+        if path.is_dir():
+            path = path / "attachments.tar.gz"
```
nadzyah, line 123: Same with attachments file name

cli/testflinger_cli/__init__.py
```diff
@@ -351,6 +414,32 @@ def submit_job_data(self, data):
...
+                ) from exc
+            if exc.status == 404:
+                raise SystemExit(
+                    "Received 404 error from server. Are you "
+                    "sure this is a testflinger server?"
+                ) from exc
+            # This shouldn't happen, so let's get more information
+            raise SystemExit(
+                "Unexpected error status from testflinger "
+                "server: {}".format(exc.status)
```
nadzyah, line 440: As I understand from the docs, SystemExit should be raised with `sys.exit` 

https://docs.python.org/3/library/exceptions.html#SystemExit

cli/testflinger_cli/tests/test_cli.py
```diff
@@ -85,6 +89,61 @@ def test_submit(capsys, tmp_path, requests_mock):
...
+
+        # extract the binary file data from the request
+        # (`requests_mock` only provides access to the `PreparedRequest`)
+        match = re.search(b"\r\n\r\n(.+)\r\n--", history[-1].body, re.DOTALL)
+        data = match.group(1)
+        # write the binary data to a file
+        with open("attachments.tar.gz", "wb") as attachments:
+            attachments.write(data)
+        # and check that the contents match the originals
+        with tarfile.open("attachments.tar.gz") as attachments:
```
nadzyah, line 139: I can see that the tmp archive name string is used a lot in your changes. So I definitely suggest moving it to one place, so if you want to change it, you don’t have to do it everywhere across the codebase

nadzyah (changes requested):
Thanks for your contribution! I’ve left several suggestions to make it more clear

agent/testflinger_agent/agent.py
```diff
@@ -111,6 +124,50 @@ def mark_device_offline(self):
...
+        archive_path = self.client.get_attachments(
+            job_id=job_data["job_id"], path=archive_dir
+        )
+        if archive_path is None:
+            raise FileNotFoundError
+        # extract archive data to a unique temporary folder and clean up
+        extracted_dir = tmp_dir()
+        with tarfile.open(archive_path, "r:gz") as tar:
+            tar.extractall(extracted_dir, filter=secure_filter)
+        shutil.rmtree(archive_dir)
```
plars, line 141: I think we can do more to ensure this gets cleaned up. If we raise that FileNotFoundError above, or if anything else raises an exception, then the archive_dir will just linger around. I would suggest either using tempfile.TemporaryDirectory() as a context manager and converting the string path it gives you to a Path object if you need.
Otherwise, it might be possible to use a try/finally here to make sure we clean it up, but I generally like the context manager approach better since you don't have to know every possible exception

cli/testflinger_cli/__init__.py
```diff
@@ -303,6 +312,53 @@ def configure(self):
             print("{} = {}".format(k, v))
         print()
 
+    @staticmethod
+    def pack_attachments(job_data: dict) -> Optional[Path]:
+        """Return the path to a compressed tarball of attachments"""
+
+        # pull together the attachement data per phase
```
plars, line 319: *attachment

cli/testflinger_cli/__init__.py
```diff
@@ -317,8 +373,16 @@ def submit(self):
...
                 ) from exc
-        job_id = self.submit_job_data(data)
-        queue = yaml.safe_load(data).get("job_queue")
+        job_dict = yaml.safe_load(data)
+
+        archive_path = self.pack_attachments(job_dict)
+        job_id = self.submit_job_data(job_dict)
+        if archive_path is not None:
+            # follow job submission with attachment submission, if required
+            self.submit_job_attachments(job_id, archive_path)
```
plars, line 382: I'm thinking that we may want to handle failures with submit_job_attachments a bit differently. If we have it detect http errors and exit, then we've already transmitted the job_data and it's just stuck in a waiting state for the attachments. Eventually, I know the db ttl should clean it up, but in the meantime, users would just see that their job is out there waiting to run but never does.
Perhaps we should retry here, and if we still don't succeed, then attempt to cancel the job and warn the user that it was unsuccessful before exiting. What do you think?

plars (changes requested):
I really love some of the improvements you made here, great work! I have a few comments below, let me know what you think.

cli/testflinger_cli/__init__.py
```diff
@@ -328,10 +392,10 @@ def submit(self):
         if self.args.poll:
             self.do_poll(job_id)
 
-    def submit_job_data(self, data):
+    def submit_job_data(self, data_dict):
```
thp-canonical, line 395:
```suggestion
    def submit_job_data(self, data_dict: dict):
```

thp-canonical: .. or even:

```suggestion
    def submit_job_data(self, data: dict):
```

cli/testflinger_cli/__init__.py
```diff
@@ -351,6 +415,32 @@ def submit_job_data(self, data):
...
+            self.client.post_attachment(job_id, path)
+        except client.HTTPError as exc:
+            if exc.status == 400:
+                raise SystemExit(
+                    f"Unable to submit attachment archive for {job_id}"
+                ) from exc
+            if exc.status == 404:
+                raise SystemExit(
+                    "Received 404 error from server. Are you "
+                    "sure this is a testflinger server?"
```
thp-canonical, line 436: This can probably also happen if we're running against an old Testflinger server that doesn't support attachments yet. Not sure if we want to reflect this in the error message?

```suggestion
                    "Received 404 error from server. Are you "
                    "sure this is a testflinger server with attachments support?"
```

cli/testflinger_cli/__init__.py
```diff
@@ -351,6 +415,32 @@ def submit_job_data(self, data):
...
+                ) from exc
+            # This shouldn't happen, so let's get more information
+            raise SystemExit(
+                "Unexpected error status from testflinger "
+                "server: {}".format(exc.status)
```
thp-canonical, line 441:
```suggestion
                "Unexpected error status from testflinger "
                f"server: {exc.status}"
```

cli/testflinger_cli/client.py
```diff
@@ -86,6 +87,30 @@ def put(self, uri_frag, data, timeout=15):
...
+                files = {"file": (path.name, file)}
+                response = requests.post(uri, files=files, timeout=timeout)
+            except requests.exceptions.ConnectTimeout:
+                logger.error(
+                    "Timout while trying to communicate with the server."
+                )
+                sys.exit(1)
+            except requests.exceptions.ConnectionError:
+                logger.error("Unable to communicate with specified server.")
+                sys.exit(1)
```
thp-canonical, line 109: You use `raise SystemExit()` above, but `logger.error()` + `sys.exit(1)` here. Not sure which one is preferred by the testflinger project, but we should be consistent with one or the other if possible.

cli/testflinger_cli/tests/test_cli.py
```diff
@@ -85,6 +89,61 @@ def test_submit(capsys, tmp_path, requests_mock):
...
+        # - there is a request to the job submission endpoint
+        # - there a request to the attachment submission endpoint
+        history = mocker.request_history
+        assert len(history) == 2
+        assert history[-2].path == "/v1/job"
+        assert history[-1].path == f"/v1/job/{job_id}/attachments"
+
+        # extract the binary file data from the request
+        # (`requests_mock` only provides access to the `PreparedRequest`)
+        match = re.search(b"\r\n\r\n(.+)\r\n--", history[-1].body, re.DOTALL)
```
thp-canonical, line 133: Is this the only way how to retrieve it? Seems fragilie (what if chunked encoding is used? etc etc...)

server/src/api/schemas.py
```diff
@@ -64,6 +64,30 @@ class ActionIn(Schema):
     action = fields.String(required=True, validate=OneOf(["cancel"]))
 
 
+class Attachment(Schema):
+    """Attachment pathnames schema
+
+    - `local`: path to attachment on the runner
+    - `agent`: path to copy the attachment in the testflinger agent (optional)
+    - `device`: path to copy the attachment in the device under test (optional)
```
thp-canonical, line 72: What's the difference between "local" (the "runner") and "agent" here?

server/src/database.py
```diff
@@ -27,6 +32,46 @@
...
+    mongo_port = os.environ.get("MONGODB_PORT", "27017")
+    mongo_uri = os.environ.get("MONGODB_URI")
+    mongo_auth = os.environ.get("MONGODB_AUTH_SOURCE", "admin")
+
+    if not mongo_uri:
+        if not (mongo_host and mongo_db):
+            raise SystemExit("No MongoDB URI configured!")
+        mongo_creds = (
+            f"{mongo_user}:{mongo_pass}@" if mongo_user and mongo_pass else ""
+        )
```
thp-canonical, line 58: Does mogodb not support an empty password? Right now, if `MONGODB_USERNAME=foo` and `MONGODB_PASSWORD=` (empty), the username would be silently dropped?

server/src/database.py
```diff
@@ -51,3 +96,80 @@ def create_indexes():
...
+            projection={"job_id": True, "job_data": True, "_id": False},
+        )
+    except TypeError:
+        return None
+    if not response:
+        return None
+    # Flatten the job_data and include the job_id
+    job = response.get("job_data")
+    job_id = response["job_id"]
+    job["job_id"] = job_id
```
thp-canonical, line 174: Since we assume `job["job_id"]` can be set (so it's not `None`), probably `response[...]` is OK instead of `response.get(...)`?

```suggestion
    job = response["job_data"]
    job_id = response["job_id"]
    job["job_id"] = job_id
```

thp-canonical (changes requested):
This is a welcome feature addition, thanks for working on this! See comments.

server/src/api/schemas.py
boukeas: The local/runner filenames refer to files on the machine where the job is submitted. That's where the attachments are collected and packed. That could be on the machine running the testflinger CLI or a Github runner. You may want to rename/move these files on the testflinger agent, so then you can specify their new name and location (always under `attachments/{phase}` with the `agent` field.

server/src/database.py
boukeas: I agree but this is existing code, I simply refactored it into the `database` module.

cli/testflinger_cli/__init__.py
boukeas: @plars Any thoughts on this? `SystemExit` is used throughout `testflinger_cli`, whereas `exit` is used throughout `testflinger_cli.client`. I followed the existing approach within each module but @thp-canonical made a similar comment regarding consistency. We could open a separate issue for this but I don't feel that addressing this is within the scope of the current PR.

agent/testflinger_agent/agent.py
boukeas: You're right, I will add a message to the exception. Logging is taken care of by the `except` block that catches the exception (and will use that message)

cli/testflinger_cli/tests/test_cli.py
boukeas: This is a unit test to intercept (and check) the request sent by the Testflinger CLI. I'm afraid there's currently no way to retrieve the actual `Request` (as opposed to the binary `PreparedRequest`) and I couldn't come up with a better alternative. If there is a change in the way the CLI makes the request then the test might also need to change to reflect that.

cli/testflinger_cli/__init__.py
```diff
@@ -303,6 +312,53 @@ def configure(self):
...
+                )
+                # add the corresponding files to the archive
+                # using a phase-prefixed archive filename
+                for filename in filenames:
+                    if filename.is_absolute():
+                        # absolute filenames become relative
+                        relative = filename.relative_to(filename.root)
+                        archive_name = phase_path / relative
+                    else:
+                        archive_name = phase_path / filename
```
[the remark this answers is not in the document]
boukeas: This is not really a check for client-submitted data. In both cases, (for different reasons) we need to add a prefix path A to a path B and the `/` operator won't work if B is an absolute path. That's the only reason why these absolute paths need to be made relative.

cli/testflinger_cli/__init__.py
nadzyah: I suggest using sys.exit still, since you can also specify the exit code there

cli/testflinger_cli/tests/test_cli.py
boukeas: In this case it's just a test, the file is created and consumed (checked) on the spot.

plars (changes requested):
This looks reasonable to me. I haven't had a chance to try it end-to-end yet, but I know you did and we will also want to do it in staging as a final sanity check. However I had one other thing that I wanted to ask below about how we ensure the attachments are removed once they are uploaded?
I did see that you have an uploadDate timestamp on it to use for TTL. Would be nice if we could delete-on-use, but I'm ok with the TTL, but didn't see where it was getting set in the index data. Did I just miss it?

plars (dismissed):

cli/testflinger_cli/__init__.py
boukeas: @nadzyah I have now used `sys.exit` in the Testflinger methods affected by this PR but there are dozen more instances of this throughout Testflinger that are out of scope.

agent/testflinger_agent/agent.py
```diff
@@ -15,14 +15,31 @@
...
+def secure_filter(member, path):
+    """Combine the `data` filter with custom attachment filtering
+
+    Makes sure that the starting folder for all attachments coincides
+    with one of the supported phases, i.e. that the attachment archive
+    has been created properly and no attachment will be extracted to an
+    unexpected location.
+    """
+    if not member.name.startswith(("provision", "firmware_update", "test")):
+        raise tarfile.OutsideDestinationError(member, path)
```
thp-canonical, line 39: 1. What if my file is called `provisionfoo/bar`?
2. What if my file is called `provision/../foo`?

The [`data_filter()`](https://docs.python.org/3/library/tarfile.html#tarfile.data_filter) and [`tar_filter()`](https://docs.python.org/3/library/tarfile.html#tarfile.tar_filter) functions don't seem to block that (tar_filter is mostly concerned with absolute paths, and data_filter is mostly concerned with links and special files).

Something like [`os.path.normpath()`](https://docs.python.org/3/library/os.path.html#os.path.normpath) (to resolve case 2 above) and checking for `.startswith("something/")` (to resolve case 1 above) might be a first step, haven't thought about whether that's good enough).

agent/testflinger_agent/agent.py
```diff
@@ -136,7 +165,13 @@ def process_jobs(self):
                     self.client.config.get("execution_basedir"), job.job_id
                 )
                 os.makedirs(rundir)
+
                 self.client.post_agent_data({"job_id": job.job_id})
+
+                # handle job attachments, if any
+                if job_data.get("attachments", "none") == "complete":
```
thp-canonical, line 172: The `"none"` literal string seems weird, as it's not used for comparison, and `dict.get()` has a default value of `None` (the value, not the string) anyway:

```suggestion
                if job_data.get("attachments") == "complete":
```

agent/testflinger_agent/client.py
```diff
@@ -105,6 +106,27 @@ def check_jobs(self):
...
             time.sleep(60)
 
+    def get_attachments(self, job_id: str, path: Path):
+        """Download the attachment archive associated with a job
+
+        :param job_id:
+            Id for the job
+        :param path:
+            Where to save the attachment archive.
+        """
```
thp-canonical, line 116: Return type and return value docstring could be added here.

cli/testflinger_cli/__init__.py
```diff
@@ -303,6 +320,44 @@ def configure(self):
...
+    def extract_attachment_data(job_data: dict) -> Optional[dict]:
+        """Pull together the attachment data per phase from the `job_data`"""
+        attachment_data = {}
+        for phase in ("provision", "firmware_update", "test"):
+            phase_str = f"{phase}_data"
+            try:
+                attachment_data[phase] = job_data[phase_str]["attachments"]
+            except KeyError:
+                pass
+        return attachment_data if attachment_data else None
```
thp-canonical, line 333: Shortcut:

```suggestion
        return attachment_data or None
```

cli/testflinger_cli/__init__.py
```diff
@@ -328,29 +402,77 @@ def submit(self):
...
+
+        :param job_id:
+            ID for the test job
+        :param path:
+            The path to the attachment archive
+        """
+        # defaults for retries
+        wait = self.config.get("attachments_retry_wait") or 10
+        timeout = self.config.get("attachments_timeout") or 600
+        tries = self.config.get("attachments_tries") or 3
```
thp-canonical, line 439: It would seem more idiomatic to use the `default` parameter of `dict.get()` for this:

```suggestion
        wait = self.config.get("attachments_retry_wait", 10)
        timeout = self.config.get("attachments_timeout", 600)
        tries = self.config.get("attachments_tries", 3)
```

cli/testflinger_cli/__init__.py
```diff
@@ -328,29 +402,77 @@ def submit(self):
...
+                    ) from error
+                # This shouldn't happen, so let's get more information
+                sys.exit(
+                    "Unexpected error status from testflinger server "
+                    f"({error.response.status_code}): {error.response.text}"
+                )
+            except (requests.Timeout, requests.ConnectionError):
+                # recoverable errors, try again
+                time.sleep(wait)
+                wait *= 0.3
```
thp-canonical, line 465: Is that intended for the wait to get shorter and shorter? Usually for retry logic like this, [Exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) is the way to go, but this seems to be the opposite of that (try more aggressively and frequently)? Or was that supposed to be `wait *= 1.3` or some other factor >= 1.0?

cli/testflinger_cli/client.py
```diff
@@ -112,19 +143,31 @@ def post_job_state(self, job_id, state):
         data = {"job_state": state}
         self.put(endpoint, data)
 
-    def submit_job(self, job_data):
+    def submit_job(self, data: dict) -> str:
         """Submit a test job to the testflinger server
 
         :param job_data:
-            String containing json or yaml data for the job to submit
+            Dictionary containing json or yaml data for the job to submit
```
thp-canonical, line 150: The string contains YAML (or JSON, as YAML is compatible with it), but the dictionary is just a data structure, and doesn't "contain" YAML or JSON itself (it will be serialized as JSON during the request).

This might be seen as nitpick-y, but it would be more correct as:

```suggestion
            Dictionary containing the job to submit
```

cli/testflinger_cli/tests/test_cli.py
```diff
@@ -85,6 +90,215 @@ def test_submit(capsys, tmp_path, requests_mock):
...
+                    # include the submission JSON itself as a test attachment
+                    "local": str(job_file)
+                }
+            ]
+        },
+    }
+    job_file.write_text(json.dumps(job_data))
+
+    sys.argv = ["", "submit", str(job_file)]
+    tfcli = testflinger_cli.TestflingerCli()
```
thp-canonical, line 167: Out of scope for this PR, but it might be useful for `TestflingerCli()` to accept an `argv: list[str]` parameter instead of relying on overwriting `sys.argv`.

cli/testflinger_cli/__init__.py
boukeas: Great spot, it should have been `1.3`. Will fix.

agent/testflinger_agent/agent.py
boukeas: `"none"` is actually an (admittedly poorly chosen) constant that signifies that this job does not contain attachments. I added that default here so that existing tests that don't involve attachments don't fail. Having said that, you're probably right: we can just assume that the `attachments` field is simply not set when attachments are not present, rather than setting it to `None`.

agent/testflinger_agent/client.py
boukeas: Attempting to address this, I realised that function doesn't need to return anything -- that was left over from a previous iteration. So I removed the return statement rather than modifying anything here.

cli/testflinger_cli/tests/test_cli.py
boukeas: @plars

cli/testflinger_cli/client.py
```diff
@@ -86,6 +89,34 @@ def put(self, uri_frag, data, timeout=15):
...
+                files = {"file": (path.name, file)}
+                response = requests.post(uri, files=files, timeout=timeout)
+            except requests.exceptions.ConnectTimeout:
+                logger.error(
+                    "Timeout while trying to communicate with the server."
+                )
+                raise
+            except requests.exceptions.ReadTimeout:
+                logger.error("Timeout while communicating with the server.")
+                raise
```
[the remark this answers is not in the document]
boukeas: The messages are slightly different but they _are_ different. The `ConnectTimeout` doesn't manage to connect to the server (hence the "trying to"), whereas the `ReadTimeout` means that a connection was established but the response wasn't completely received within the timeout limit (hence the "while communicating). So we want to keep them separate.

agent/testflinger_agent/agent.py
```diff
@@ -15,14 +15,39 @@
...
+def secure_filter(member, path):
+    """Combine the `data` filter with custom attachment filtering
+
+    Makes sure that the starting folder for all attachments coincides
+    with one of the supported phases, i.e. that the attachment archive
+    has been created properly and no attachment will be extracted to an
+    unexpected location.
+    """
+    try:
+        resolved = Path(member.name).resolve().relative_to(Path.cwd())
```
plars, line 39: Nice! I see you already have a test for things that don't begin with a valid name, but this was a really nice fix. Would it possible to also add a test to ensure that it fails properly if you try to give it an attachment for something like `test/../bad` as the path?

plars (changes requested):
One minor suggestion about a test for the secure_filter() update that you just made, otherwise looks good

cli/testflinger_cli/client.py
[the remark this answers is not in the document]
nadzyah: I see that they are different, but this difference is almost unnoticeable. As I understand, the first message should give a user the understanding that "We've tried to communicate with the server, but cannot even reach it." 
The second one is about "The server is reachable and we've even managed to establish the session, but got en error while reading the application data from the server".

I suggest updating the messages and add more context

plars (approved):
Thanks for adding that! +1

cli/testflinger_cli/client.py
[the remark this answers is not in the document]
boukeas: Alright, I've based the modified messages on the [exception explanations](https://requests.readthedocs.io/en/latest/api/#exceptions) from the `requests` library. So it should be as clear and accurate as possible.

plars (approved):

cli/testflinger_cli/__init__.py
```diff
@@ -119,6 +125,19 @@ def _print_queue_message():
...
+    """Return resolved relative `path`"""
+    if path.is_absolute():
+        # strip leading '/' from absolute path
+        path = path.relative_to(path.root)
+    # resolve the path and make relative again
+    return path.resolve().relative_to(Path.cwd())
+
+
+class AttachmentError(Exception):
+    """Exception thrown when attachments fail to be submitted"""
```
nadzyah, line 138: Is it supposed to have an empty body?

server/src/api/v1.py
```diff
@@ -146,6 +157,54 @@ def job_get_id(job_id):
...
+        UUID as a string for the job
+    :return:
+        send_file stream of attachment tarball to download
+    """
+    if not check_valid_uuid(job_id):
+        return "Invalid job id\n", 400
+    try:
+        file = database.retrieve_file(filename=f"{job_id}.attachments")
+    except FileNotFoundError:
+        return "", 204
```
nadzyah, line 174: I just want to say that I like this approach with this status code so much!

cli/testflinger_cli/__init__.py
boukeas: Yes, the linter doesn't like `pass` along with the docstring (which is also considered a statement essentially).

nadzyah: I actually didn't mean to “pass” keywork but the class implementation itself😅

PR #254, opened 2024-04-12
Add a fake device connector for testing purposes

plars:
## Description
Add a fake device connector that might be useful for local testing

## Documentation
Added a blurb about it in the README and in the reference docs

## Tests
Tested locally for now, but this could be used for some better tests in the future


docs/reference/device-connector-types.rst
```diff
@@ -73,6 +73,13 @@ The ``dragonboard`` device connector supports the following ``provision_data`` k
        the ``zstd`` tool is supported) and
        flashed to the SD card, which will be used to boot up the DUT.
 
+.. _fake_connector:
+
+_fake_connector
+---------------
```
nancyc12, line 79:
```suggestion
fake_connector
--------------
```

nancyc12 (changes requested):
I like the idea and can imagine this will benefit developers! Mostly looks good to me, just one suggestion.

nancyc12 (approved):
LGTM+1

PR #258, opened 2024-04-19
fix: match python version in agent test workflow to version on agent

boukeas:
## Description and resolved issues

Merging [#250](https://github.com/canonical/testflinger/pull/250) and deploying to staging uncovered certain issues.
- Unpacking attachments _fails_ on the agent used for testing because unpacking uses the `filter` argument of `tarfile.extractall` which is not supported in the Python version installed on the agent, i.e. `3.8.10`.
- This issue would normally have surfaced in the [agent testing workflow](https://github.com/canonical/testflinger/blob/main/.github/workflows/agent-tox.yml), prior to deployment. It did not because that workflow does not specify a Python patch version (only `3.8`), resulting in the _latest_ patch version being installed, i.e. `3.8.18`). The `filter` argument was added in version `3.8.17`, so it is supported in the environment used for testing in the workflow but not in the actual agent.
- Testing on staging revealed that the provisioning and firmware update stages were triggered when attachments were provided under the `provision_data` and `firmware_update_data` fields. The code assumes that any data under these fields suggests that the corresponding stages should be triggered.

## Changes

- A fully qualified Python version (`3.8.12`) is specified in the agent testing workflow. It does not match the Python version installed on the agent because that version is not available on our self-hosted runners, but it is sufficient for testing because the filtering feature was introduced in `3.8.17`.
- The `filter` functionality provides an important safety feature. Instead of re-implementing what would necessarily be a subset of that functionality, I decided to use a minimal subset of the [implementation provided in the `main` branch of the Python standard library](https://github.com/python/cpython/commits/main/Lib/tarfile.py). A check in the imports section decides whether or not to use the `tarfile` that comes with the current Python installation (if it supports `filter`) or the alternative one otherwise. The major advantages to this solution is that (a) the actual unpacking code remains unchanged and uses the `filter` functionality and (b) this is not a version-sensitive solution, i.e. whenever it can be guaranteed that the agents will run a Python version supporting `filter`, the unpacking code will remain the same and only the conditional import will be removed.
- After attachments are unpacked, attachment data is stripped from the job data dict, restoring it to a state that is identical to how it would be if attachments had not been included (i.e. like the good old days). This respects any assumptions that Testflinger makes about the contents of the job data dict and removes any possible interference with other stages. Note that attachment data doesn't even need to be included in the job data dict in the first place (it's all in the gzipped-tar archive) and it may be removed anyway in a later version. The reasons it has been kept are that (a) it is useful to display the attachments under the `Provision Data` and `Firmware update data` sections of the Job page on Testflinger and (b) it is yet unclear how that attachment data might be used by a device connector later down the line.


plars (commented):
patching things up like this is certainly not my favorite thing to do, but I'm ok with it as a temporary measure if we don't have a better approach. I was really hoping we could use the get_members workaround that I suggested before, but you're right that there's a lot going on in the existing data_filter and get_filtered_attrs that we would miss if we just did that.

agent/testflinger_agent/tarfile_patch.py
```diff
@@ -0,0 +1,260 @@
+try:
+    # attempt importing a filter, to check if the feature is present
+    from tarfile import data_filter
+except ImportError:
```
thp-canonical, line 4: This "file-level" try..except makes everything indented. It's not like we're running out of storage space for 0x20 bytes, but maybe this try..except should be done in `agent.py`, like this:

```python3
try:
    from tarfile import data_filter
    import tarfile
    del data_filter # optional, if you want to clean up after the try-import
except ImportError:
    from . import tarfile_patch as tarfile
```

Then this file (`tarfile_patch.py`) can just contain the implementation, and the "policy" whether to use the Python-supplied `tarfile` module or `tarfile_patch` is up to the user.

Alternatively:

```python3
import tarfile

if not hasattr(tarfile, 'data_filter'):
    # Python < 3.8.17 compatibility
    import tarfile_patch
    tarfile.open = tarfile_patch.open
```

Or even:

```python3
import sys

if sys.version_info < (3, 12):
    import .tarfile_patch as tarfile
else:
    import tarfile
```

(the versions and names might be slightly different, haven't checked)

thp-canonical (commented):
Some ideas.

agent/tox.ini
```diff
@@ -18,5 +18,5 @@ deps =
 commands =
     {envbindir}/python setup.py develop
     {envbindir}/python -m black --check setup.py testflinger_agent
-    {envbindir}/python -m flake8 setup.py testflinger_agent
+    {envbindir}/python -m flake8 setup.py testflinger_agent --exclude testflinger_agent/tarfile_patch.py
```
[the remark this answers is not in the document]
boukeas: It had an issue with `tarfile.open` being imported but unused. However, moving the conditional import logic in the agent resolved that as well.

agent/testflinger_agent/agent.py
```diff
@@ -143,11 +153,34 @@ def unpack_attachments(self, job_data: dict, cwd: Path):
...
+                try:
+                    del phase_data["attachments"]
+                except KeyError:
+                    pass
+                else:
+                    # it may be the case that attachments were the only data
+                    # included for this phase, so the phase can now be removed
+                    if not phase_data:
+                        del job_data[phase_str]
+
```
nadzyah, line 184: Is it possible to simplify this logic? We can try something like this:

```suggestion
        for phase in ("provision", "firmware_update", "test"):
            phase_str = f"{phase}_data"
            phase_data = job_data.get(phase_str)

            if phase_data:
                # Attempt to delete attachments if they exist
                phase_data.pop("attachments", None)

                # If phase_data is empty, remove it from job_data
                if not phase_data:
                    del job_data[phase_str]
```

agent/testflinger_agent/tarfile_patch.py
boukeas: I moved the conditional import logic into the agent, you are absolutely right that it belongs there.

agent/testflinger_agent/tarfile_patch.py
```diff
@@ -0,0 +1,265 @@
...
+    if for_data:
+        # Ignore ownership for 'data'
+        if member.uid is not None:
+            new_attrs["uid"] = None
+        if member.gid is not None:
+            new_attrs["gid"] = None
+        if member.uname is not None:
+            new_attrs["uname"] = None
+        if member.gname is not None:
+            new_attrs["gname"] = None
```
nadzyah, line 101: I would like also to suggest the simplification here 

```suggestion
        # Ignore ownership for 'data'
        attributes = ["uid", "gid", "uname", "gname"]
        for attr in attributes:
            if getattr(member, attr) is not None:
                new_attrs[attr] = None
```

agent/testflinger_agent/tarfile_patch.py
```diff
@@ -0,0 +1,265 @@
...
+    if linkname is not _KEEP:
+        result.linkname = linkname
+    if uid is not _KEEP:
+        result.uid = uid
+    if gid is not _KEEP:
+        result.gid = gid
+    if uname is not _KEEP:
+        result.uname = uname
+    if gname is not _KEEP:
+        result.gname = gname
```
nadzyah, line 155:
```suggestion
    attributes = {
        "name": name,
        "mtime": mtime,
        "mode": mode,
        "linkname": linkname,
        "uid": uid,
        "gid": gid,
        "uname": uname,
        "gname": gname,
    }

    for key, value in attributes.items():
        if value is not _KEEP:
            setattr(result, key, value)
```

nadzyah (changes requested):
The changes look good in general, I'd just like to suggest the simplification of some parts of the code

agent/testflinger_agent/agent.py
```diff
@@ -17,13 +17,24 @@
...
 from testflinger_agent.config import ATTACHMENTS_DIR
 
+try:
+    # attempt importing a tarfile filter, to check if filtering is supported
+    from tarfile import data_filter
+
+    del data_filter
+except ImportError:
+    # import a patched version of `tarfile` that supports filtering
+    from . import tarfile_patch as tarfile
```
thp-canonical, line 33: It might be useful to add to the comment here something like "This can be removed once the minimum version of Python is X.Y.Z" (3.9? 3.12? 3.8.17?)

thp-canonical (approved):
I didn't check whether `agent/testflinger_agent/tarfile_patch.py` is a proper subset of Python 3.8's `tarfile.py`, but trust that somebody else will verify that :)

Added one more comment about simplifying the nested try..except..else blocks, but other than that, I think this is fine. It might also make sense to add some comment along the lines of "if you read this comment and the minimum supported Python version is >= X.Y.Z, the `tarfile_patch.py` file can be removed and the import changed back to `tarfile`".

agent/testflinger_agent/tarfile_patch.py
boukeas: Almost all code in the patch is copied from the standard library implementation of `tarfile`. I concede that your suggestion is equivalent, shorter and simpler but I am not too keen to make changes there. For one thing, it will remain simpler to compare versions (e.g. to verify that we don't diverge).

agent/testflinger_agent/tarfile_patch.py
boukeas: Same here, this is copied from the standard library implementation of `tarfile`.

agent/testflinger_agent/tarfile_patch.py
nadzyah: Why are we copying this code instead of using it as a dependency?

agent/testflinger_agent/agent.py
boukeas: Done. I actually had one and then removed it :grin:

agent/testflinger_agent/agent.py
boukeas: I think I want to keep the outer try-except but I admit that the nested one makes the logic harder to follow (Thomas made the very same remark as well). Using `pop` is a great idea to resolve that and I'll definitely follow that suggestion.

agent/testflinger_agent/agent.py
```diff
@@ -143,11 +154,34 @@ def unpack_attachments(self, job_data: dict, cwd: Path):
...
+            else:
+                try:
+                    del phase_data["attachments"]
+                except KeyError:
+                    pass
+                else:
+                    # it may be the case that attachments were the only data
+                    # included for this phase, so the phase can now be removed
+                    if not phase_data:
+                        del job_data[phase_str]
```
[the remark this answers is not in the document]
boukeas: Addressed comment following @nadzyah's suggestion.

agent/testflinger_agent/tarfile_patch.py
boukeas: I cover that in the description of the PR: the filtering code is not available in Python 3.8.10 (which is the version the agents are running) so we provide a patch.

nadzyah: Do you think we’ll remove this script and start using the lib itself after dropping python 3.8 support?

.github/workflows/agent-tox.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     runs-on: [self-hosted, linux, jammy, X64]
     strategy:
       matrix:
-        python: ["3.8", "3.10"]
+        python: ["3.8.12", "3.10"]
```
plars, line 21: I think this was supposed to be 3.8.10 to match the version we're targeting for now right?
From focal:
```
$ python3 --version
Python 3.8.10
```

plars (commented):
one very small thing, otherwise looks fine to me.

agent/testflinger_agent/tarfile_patch.py
boukeas: Exactly. See also this [added comment](https://github.com/canonical/testflinger/pull/258/commits/c4c8b914c6bc2dfb73a09c372badcc9590e9b6c7) that clarifies that.

.github/workflows/agent-tox.yml
boukeas: If we make that `3.8.10` we'll get an error because that version is not available on `jammy`. `3.8.12` is the lowest version available on `jammy` _and_ it's still below `3.8.17`, where `filter` became available. So by setting that to `3.8.12`, even though we don't match the version on the agent, we still manage to test on a version that doesn't support filtering (i.e. testing the patch), without going into all the trouble of setting up `pyenv`.

nadzyah (approved):

PR #262, opened 2024-04-28
Testflinger server agent poetry

plars:
## Description
@val500 is working on some changes where it would be really nice to be able to create some kind of common code that is used by multiple subprojects. I experimented a bit with how to accomplish this and created a /common dir wth a testflinger_common module in it, but there doesn't seem to be a way with pip to specify dependencies from a relative path in the same monorepo. However, poetry does allow this quite nicely with something like `+testflinger-common = { path = "../common" }
`

We had already started using poetry with some of our newer projects also, so this seemed like a good time to go ahead and start the process of moving these projects to poetry. The other nice thing about this is that it can preserve the ability to use pip for installation, which means not much needs to change for existing documentation, charms, oci containers, etc.

## Documentation

One small change to the agent readme which still referenced setup.py, but isn't the recommended approach for a while.

## Tests
Running tests through tox still works for me locally, as well as rebuilding the container.
Some changes will be needed later if we add a testflinger_common module, but I think that should be a separate PR


plars:
Also really curious what you think of this @boukeas? Varun said you had discussed using namespaces to address this. I think namespaces might require more structural changes that could have unintended side effects. If there are other possible benefits to that approach, I'd love to discuss, but I'm thinking this might be a good option for now that doesn't necessarily prevent us from doing namespaces later if that's something we still want.

server/pyproject.toml
```diff
@@ -1,9 +1,34 @@
...
+flask = "^3.0.3"
+flask-pymongo = "^2.3.0"
+gunicorn = "^22.0.0"
+gevent = "^24.2.1"
+prometheus-client = "^0.20.0"
+pyyaml = "^6.0.1"
+sentry-sdk = { extras = ["flask"], version = "^2.0.1" }
+
+[tool.poetry.dev-dependencies]
+pytest = "^8.1.2"
```
omar-selo, line 20: should we add `pytest` dev dependency to agent too?

server/pyproject.toml
```diff
@@ -1,9 +1,34 @@
...
 [tool.black]
 line-length = 79
 exclude = "charm/lib"
 
 [tool.pytest.ini_options]
-pythonpath = [
-  "charm/src",
-  "charm/lib"
-]
+pythonpath = ["charm/src", "charm/lib"]
```
omar-selo, line 34: Interesting, does server tox here test even the charm code?

omar-selo (approved):
Looks good to me, just a couple of questions

boukeas:
> Also really curious what you think of this @boukeas? Varun said you had discussed using namespaces to address this. I think namespaces might require more structural changes that could have unintended side effects. If there are other possible benefits to that approach, I'd love to discuss, but I'm thinking this might be a good option for now that doesn't necessarily prevent us from doing namespaces later if that's something we still want.

Poetry is what I have used before for packaging and managing dependencies and I think it's a wonderful tool. I have _not_ used it for this sort of use case though so I will need to dive a little deeper to understand what's being done here.

I have not used namespaces in action. I suggested the approach because reading what they're designed for and what we need to do seemed to coincide completely. But I wouldn't insist on using them. It might be worth doing a spike to investigate what they would look like in the case of Testflinger.

server/pyproject.toml
plars: It shows that I already added them, am I missing something that's different on the similar section of pyproject.toml for the agent?

server/pyproject.toml
plars: True, I thought I had it set up to run from one of the github actions at one point but I don't see it now. I'll try to sort out somewhere to run this if things change with the charm in another PR. For now, the charm rarely changes though so I don't think it needs to rerun the unit tests in the charm for changes to the server code normally.

server/pyproject.toml
omar-selo: you're right, Idk how I missed it

server/pyproject.toml
```diff
@@ -1,9 +1,34 @@
...
+pyyaml = "^6.0.1"
+sentry-sdk = { extras = ["flask"], version = "^2.0.1" }
+
+[tool.poetry.dev-dependencies]
+pytest = "^8.1.2"
+
+[tool.poetry.scripts]
+testflinger-agent = "testflinger_agent.cmd:main"
+
+[build-system]
```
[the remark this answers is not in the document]
plars: yes

server/pyproject.toml
```diff
@@ -1,9 +1,34 @@
...
+gunicorn = "^22.0.0"
+gevent = "^24.2.1"
+prometheus-client = "^0.20.0"
+pyyaml = "^6.0.1"
+sentry-sdk = { extras = ["flask"], version = "^2.0.1" }
+
+[tool.poetry.dev-dependencies]
+pytest = "^8.1.2"
+
+[tool.poetry.scripts]
```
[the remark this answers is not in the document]
plars: fixed, thanks!

boukeas (approved):

PR #263, opened 2024-04-29
Add job status endpoint to server

val500:
## Description

This adds an endpoint to Testflinger server that receives status updates from Testflinger agents. These updates require a webhook that the server will forward the rest of the update message to.

## Documentation

Documentation was added to job_schema.rst to reflect new job_status_webhook option in the job yaml. The server ReadMe was updated with details on the new status endpoint. 

## Web service API changes
- Interface: [POST] /v1/status
- Request Structure: 
    ```json
    {
        "agent_id": "<string>",
        "job_queue": "<string>",
        "job_status_webhook": "<URL as string>",
        "events": [
            {
  	        "event_name": "<string enum of events>",
  	        "timestamp": "<string>",
  	        "detail_msg": "<string>"
            },
            ...
         ]
    }

    ```

- Response Status Codes:
    - HTTP 200 (OK)
    - HTTP 400 (Bad request) - The arguments could not be processed by the server
    - HTTP 504 (Gateway Timeout) - The webhook URL timed out
- Rationale: This endpoint receives status updates from the agent if there is a webhook specified in the job yaml and posts the update to this webhook.

## Tests

For the server, unit tests were added to test_v1.py. For the agents, unit tests were added to both test_client.py and test_agent.py to test both webhook posting behavior and overall status updating.

### DIFF
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -22,7 +22,9 @@ import tempfile
 from testflinger_agent.job import TestflingerJob
 from testflinger_agent.errors import TFServerError
 from testflinger_agent.config import ATTACHMENTS_DIR
-from testflinger_common.enums import JobState, TestPhase
+from testflinger_agent.event_emitter import EventEmitter
+from testflinger_common.enums import JobState, TestPhase, TestEvent
+
 
 try:
     # attempt importing a tarfile filter, to check if filtering is supported
@@ -189,6 +191,7 @@ class TestflingerAgent:
 
     def process_jobs(self):
         """Coordinate checking for new jobs and handling them if they exists"""
+
         TEST_PHASES = [
             TestPhase.SETUP,
             TestPhase.PROVISION,
@@ -202,11 +205,17 @@ class TestflingerAgent:
         self.retry_old_results()
 
         self.check_restart()
-
         job_data = self.client.check_jobs()
         while job_data:
             try:
                 job = TestflingerJob(job_data, self.client)
+                event_emitter = EventEmitter(
+                    job_data.get("job_queue"),
+                    job_data.get("job_status_webhook"),
+                    self.client,
+                )
+                job_end_reason = TestEvent.NORMAL_EXIT
+
                 logger.info("Starting job %s", job.job_id)
                 rundir = os.path.join(
                     self.client.config.get("execution_basedir"), job.job_id
@@ -243,32 +252,48 @@ class TestflingerAgent:
                         == JobState.CANCELLED
                     ):
                         logger.info("Job cancellation was requested, exiting.")
+                        event_emitter.emit_event(TestEvent.CANCELLED)
                         break
+
                     self.client.post_job_state(job.job_id, phase)
                     self.set_agent_state(phase)
-                    exit_code = job.run_test_phase(phase, rundir)
 
-                    self.client.post_influx(phase, exit_code)
+                    event_emitter.emit_event(TestEvent(phase + "_start"))
+                    exitcode, exit_event, exit_reason = job.run_test_phase(
+                        phase, rundir
+                    )
+                    self.client.post_influx(phase, exitcode)
+                    event_emitter.emit_event(exit_event, exit_reason)
 
-                    if phase == "provision":
+                    if exitcode:
                         # exit code 46 is our indication that recovery failed!
                         # In this case, we need to mark the device offline
-                        if exit_code == 46:
+                        if exitcode == 46:
                             self.mark_device_offline()
-                            # Replace with TestEvent enum values once it lands
-                            detail = "recovery_fail"
-                        detail = "provision_fail" if exit_code else ""
-                        self.client.post_provision_log(
-                            job.job_id, exit_code, detail
-                        )
-                    if phase != "test" and exit_code:
-                        logger.debug("Phase %s failed, aborting job" % phase)
-                        break
+                            exit_event = TestEvent.RECOVERY_FAIL
+                        else:
+                            exit_event = TestEvent(phase + "_fail")
+                        event_emitter.emit_event(exit_event)
+                        if phase == "provision":
+                            self.client.post_provision_log(
+                                job.job_id, exit_code, exit_event
+                            )
+                        if phase != "test":
+                            logger.debug(
+                                "Phase %s failed, aborting job" % phase
+                            )
+                            job_end_reason = exit_event
+                            break
+                    else:
+                        event_emitter.emit_event(TestEvent(phase + "_success"))
             except Exception as e:
                 logger.exception(e)
             finally:
                 # Always run the cleanup, even if the job was cancelled
+                event_emitter.emit_event(TestEvent.CLEANUP_START)
                 job.run_test_phase(TestPhase.CLEANUP, rundir)
+                event_emitter.emit_event(TestEvent.CLEANUP_SUCCESS)
+                event_emitter.emit_event(TestEvent.JOB_END, job_end_reason)
                 # clear job id
                 self.client.post_agent_data({"job_id": ""})
 
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -213,10 +213,15 @@ class TestflingerAgent:
                     job_data.get("job_queue"),
                     job_data.get("job_status_webhook"),
                     self.client,
+                    job.job_id,
                 )
                 job_end_reason = TestEvent.NORMAL_EXIT
 
                 logger.info("Starting job %s", job.job_id)
+                event_emitter.emit_event(
+                    TestEvent.JOB_START,
+                    f"{self.client.server}/job/{job.job_id}/events",
+                )
                 rundir = os.path.join(
                     self.client.config.get("execution_basedir"), job.job_id
                 )
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -220,7 +220,7 @@ class TestflingerAgent:
                 logger.info("Starting job %s", job.job_id)
                 event_emitter.emit_event(
                     TestEvent.JOB_START,
-                    f"{self.client.server}/job/{job.job_id}/events",
+                    f"{self.client.server}/jobs/{job.job_id}",
                 )
                 rundir = os.path.join(
                     self.client.config.get("execution_basedir"), job.job_id
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -264,16 +264,16 @@ class TestflingerAgent:
                     self.set_agent_state(phase)
 
                     event_emitter.emit_event(TestEvent(phase + "_start"))
-                    exitcode, exit_event, exit_reason = job.run_test_phase(
+                    exit_code, exit_event, exit_reason = job.run_test_phase(
                         phase, rundir
                     )
-                    self.client.post_influx(phase, exitcode)
+                    self.client.post_influx(phase, exit_code)
                     event_emitter.emit_event(exit_event, exit_reason)
 
-                    if exitcode:
+                    if exit_code:
                         # exit code 46 is our indication that recovery failed!
                         # In this case, we need to mark the device offline
-                        if exitcode == 46:
+                        if exit_code == 46:
                             self.mark_device_offline()
                             exit_event = TestEvent.RECOVERY_FAIL
                         else:
--- a/agent/testflinger_agent/client.py
+++ b/agent/testflinger_agent/client.py
@@ -390,7 +390,11 @@ class TestflingerClient:
             logger.warning("Unable to post provision log to server: %s", exc)
 
     def post_status_update(
-        self, job_queue: str, webhook: str, events: List[Dict[str, str]]
+        self,
+        job_queue: str,
+        webhook: str,
+        events: List[Dict[str, str]],
+        job_id: str,
     ):
         """
         Posts status updates about the running job as long as there is a
@@ -402,6 +406,8 @@ class TestflingerClient:
             String URL to post status update to
         :param events:
             List of accumulated test events
+        :param job_id:
+            id for the job on which we want to post results
 
         """
         if webhook is None:
@@ -413,7 +419,7 @@ class TestflingerClient:
             "job_status_webhook": webhook,
             "events": events,
         }
-        status_update_uri = urljoin(self.server, "/v1/agents/status")
+        status_update_uri = urljoin(self.server, f"/v1/job/{job_id}/events")
         try:
             job_request = self.session.post(
                 status_update_uri, json=status_update_request, timeout=30
--- a/agent/testflinger_agent/client.py
+++ b/agent/testflinger_agent/client.py
@@ -21,6 +21,7 @@ import shutil
 import tempfile
 import time
 
+from typing import List, Dict
 from urllib.parse import urljoin
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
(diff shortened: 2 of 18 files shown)


server/src/api/v1.py
```diff
@@ -489,6 +491,36 @@ def agents_post(agent_name, json_data):
...
+        "phases": [
+            {
+            "phase_name": "<string enum of job phases>",
+            "start": "<datetime> | None",
+            "end": "<datetime> | None",
+            "result": <int>,
+            "end_reason": "<string>",
+            },
+        ]
+    }
```
omar-selo, line 512: Would be great if the structure of the request and response are [schemas](https://github.com/canonical/testflinger/blob/main/server/src/api/schemas.py) rather than a comment. Schemas are exposed on testflinger's [api docs](https://testflinger.canonical.com/docs). And [apiflask](https://apiflask.com/) would provide automatic request validation and serialization

server/src/api/v1.py
```diff
@@ -489,6 +491,36 @@ def agents_post(agent_name, json_data):
...
+    """
+    request_json = request.get_json()
+    webhook_url = request_json.pop("webhook")
+    try:
+        # Times out after 1 seconds
+        response = requests.post(webhook_url, data=request_json, timeout=1)
+        return response.text, response.status_code
+    except requests.exceptions.Timeout:
+        return "", 500
```
omar-selo, line 521: why does it timeout after 1 second? it seems more reliable if it has retries or even better adds it to a queue to send it to TO reliably

omar-selo (changes requested):
Main concern is reliability, as in I wonder if we should add this to a queue rather than forward immediately

server/src/api/v1.py
val500: I think we timeout in order to prevent the server from hanging for too long on any individual request. We include previous phase results in each status update so if any of them fail, you can still get the information in the next one. I think @plars can expand here.

As for the queue, we had discussed using something like celery to maintain a queue of these updates, but I think @mz2 had raised some concerns about adding extra dependencies.

mz2: > I think we timeout in order to prevent the server from hanging for too long on any individual request. 

Sorry, I'm not very clear on this detail. Omar's on point about the good options: either retry, or dequeue and mark the queue event done when it was successfully delivered. In absence of a queue, retry seems saner than dropping with a 1s timeout. The server doesn't have to and therefore shouldn't try to worry about the request / response cycle or conserving itself for concurrency (add more workers if needed).

If we have elsewhere code that optimises for closing a request / response cycle quickly in case of a failure that is retryable, we should similarly not go there anywhere else either.

Let's talk at the sprint about this.

plars (changes requested):
Looks like you're missing the enums.py at the very least. Can you add that here or tell me where to find it if it's external?

agent/testflinger_agent/stop_condition_checkers.py
omar-selo, line 1: This is a tough one. Reading the stop condition checkers makes me feel that changing the whole agent flow to be event driven sounds better. The idea is if something happens like a job is cancelled, then different objects that have different responsibilities can react to this event (without mixing responsibilities together). The runner would cleanup the process. The logger would print some logs. The event emitter would forward this event to the server. But this would be a big refactor that we need to think more carefully about before doing. So I don't recommend trying to do that now. Instead, perhaps it's enough to keep the checkers as they were before. Then at the runner code, if a stop condition returns a string then we emit an appropriate event depending on the type of the checker. e.g. it he type is `JobCancelledChecker` then we `emitJobCancelled` (using the same event emitter class). What I'm not sure about though is whether `emitJobCancelled` should be triggered from the runner or the agent.

omar-selo (changes requested):

server/src/api/v1.py
```diff
@@ -489,6 +491,38 @@ def agents_post(agent_name, json_data):
...
+        ]
+    }
+
+    """
+    request_json = json_data
+    webhook_url = request_json.pop("job_status_webhook")
+    try:
+        # Times out after 1 seconds
+        response = requests.post(webhook_url, json=request_json, timeout=1)
+        return response.text, response.status_code
```
omar-selo, line 521: doesn't sound right to have a 1s timeout, I'd like to push again towards a queue

common/README.rst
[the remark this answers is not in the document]
plars: Yeah, I think the idea was to land that PR and then rebase this on main? Did you cherry-pick it here or something @val500 ?

agent/testflinger_agent/job.py
```diff
@@ -126,7 +126,7 @@ def run_test_phase(self, phase, rundir):
             )
         if phase == "allocate":
             self.allocate_phase(rundir)
-        return exitcode
+        return exitcode, exit_event, exit_reason
```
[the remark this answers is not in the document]
val500: exit_event and exit_reason are the event_name and detail_msg fields respectively. They get used in agent.py when the job gets run.

server/src/api/v1.py
omar-selo: Alternatively, if we can do something similar to [background tasks](https://fastapi.tiangolo.com/reference/background/) in fastapi

agent/testflinger_agent/job.py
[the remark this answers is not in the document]
plars: They were landed as part of a previous PR to make it easier to find this information, so that it could be returned when the phase finishes running. This way we know the reason it stopped and can send that up to the server to be logged, stored, or forwarded to test-observer as needed from one place

server/src/api/schemas.py
```diff
@@ -162,6 +163,23 @@ class Result(Schema):
...
+    detail_msg = fields.String(required=False)
+
+
+class StatusUpdate(Schema):
+    """Status Update schema"""
+
+    agent_id = fields.String(required=False)
+    job_queue = fields.String(required=False)
+    job_status_webhook = fields.URL(required=True)
+    events = fields.List(fields.Nested(JobEvent), required=False)
```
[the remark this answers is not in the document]
plars: it's already in the job data that it received when starting the job. By transmitting it at the time that we tell the server to send this, we save the server from having to do another lookup to find it from the job data.

agent/testflinger_agent/client.py
```diff
@@ -365,3 +365,39 @@ def post_influx(self, phase, result=None):
...
+        :param job_queue:
+            TestFlinger queue the currently running job belongs to
+        :param webhook:
+            String URL to post status update to
+        :param events:
+            List of accumulated test events
+
+        """
+        if webhook is None:
+            return
```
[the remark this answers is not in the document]
plars: I think this is actually ok. While it's true that we DO want to expose some information about recent provision failures on the testflinger UI under the agent detail page, I think it will have to work a little different. Instead of discrete events like this, we are really only concerned with why the provision phase finished, and probably storing it in a way that is easier to retrieve when we retrieve the other agent data. I'm going to be looking into that piece soon, and if it turns out that we can reuse this then I'll certainly try to do so. But the call for sending stuff along to test-observer is already a little strange and probably best to keep it separate for now.

common/README.rst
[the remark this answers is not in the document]
val500: Rebased on main now

agent/testflinger_agent/client.py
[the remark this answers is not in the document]
omar-selo: It does feel strange though that the agent would know about the webhook. wdyt?

server/src/api/v1.py
val500: I think for now we'll do a 3 second timeout with 3 retries. With regards to queues, there was some direction to use Temporal over Celery, but setting up that system might be better to do in a separate issue.

agent/testflinger_agent/agent.py
```diff
@@ -243,25 +254,87 @@ def process_jobs(self):
...
                     self.set_agent_state(phase)
-                    exitcode = job.run_test_phase(phase, rundir)
-
+                    self.add_event_and_post_status_update(
+                        job_queue,
+                        webhook,
+                        events,
+                        TestEvent("started_" + phase),
+                        datetime.now(timezone.utc).isoformat(),
+                        "",
```
nadzyah, line 275: Why is it an empty string? We can probably set the default value to be “” instead of requiring to pass it

val500: Changed to default to ""

agent/testflinger_agent/client.py
```diff
@@ -365,3 +365,39 @@ def post_influx(self, phase, result=None):
...
+            "agent_id": self.config.get("agent_id"),
+            "job_queue": job_queue,
+            "job_status_webhook": webhook,
+            "events": events,
+        }
+        status_update_uri = urljoin(self.server, "/v1/agents/status")
+        try:
+            job_request = self.session.post(
+                status_update_uri, json=status_update_request, timeout=1
+            )
```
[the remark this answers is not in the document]
val500: Changed to 30 seconds

server/src/api/v1.py
omar-selo: sure, as long as we do go back to it. "for now" solutions tend to become "forever" :sweat_smile:

omar-selo (dismissed):
I did have a few comments left, but to avoid blocking anything I'll approve. As I think this is a good first step

plars (changes requested):
I have a few more comments below. I think it would be nice to add a bit more on the unit tests, especially on the possible error/failure conditions. And I would definitely recommend and end-to-end test of this now that you have the test-observer side prototyped.

agent/testflinger_agent/runner.py
```diff
@@ -87,7 +89,7 @@ def check_stop_conditions(self) -> str:
             output = checker()
             if output:
                 return output
-        return ""
+        return None
```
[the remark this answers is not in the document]
val500: That's a good catch - I think it worked because I check if the stop_event is None in the run function.

omar-selo:
It occurred to me that it would be useful for TO if TF informs it about the last event. As in TF can mark an event as `last` to inform listeners that there will be no more events emitted after it. That way TO can say that this test execution has ended. Otherwise TO will need to know in detail which events would be final events. I prefer if TO doesn't actually know what the events mean for now. wdyt @plars @val500 ?

val500:
> It occurred to me that it would be useful for TO if TF informs it about the last event. As in TF can mark an event as `last` to inform listeners that there will be no more events emitted after it. That way TO can say that this test execution has ended. Otherwise TO will need to know in detail which events would be final events. I prefer if TO doesn't actually know what the events mean for now. wdyt @plars @val500 ?
> 
> Update: Just had a meeting with Andrej and Matias about these events. Here are some notes:
> 
>     * Would be great to have traceability, such that we can get to a job from TO through the events. This could be in the form of a url.
> 
>     * TO needs more information about the events, namely it needs to know if an event is the last event, and if an event is a failure/cancellation. We don't want to deduce these on TO by the event name because that couples TO to TF. Instead it would be useful if along with every event perhaps a couple of booleans are set, something like `isFinal` and `isFailure`.
> 
> 
> wdyt @plars @val500

I think every job will end with the cleanup phase regardless of whether the job was cancelled or something threw an error, so the last event will always be ended_cleanup - @plars can confirm

plars:
> @omar-selo I think every job will end with the cleanup phase regardless of whether the job was cancelled or something threw an error, so the last event will always be ended_cleanup - @plars can confirm

That's true for testflinger specifically, but the idea is to make it more generic on both sides. We could set something like a boolean "IsFinal" but for us, it would always be on that cleanup event because that's the last thing that would happen, regardless of whether any part of the job failed.

However, that means that something like an event showing that the job timed out, cancelled, etc... would have already happened. If this is what you are really interested in, then perhaps we should mark those with an "IsTerminal" event for event types that will wind up terminating the whole job? would that be enough for you to pick those out as the reason for the overall job termination?

agent/testflinger_agent/agent.py
```diff
@@ -202,11 +205,17 @@ def process_jobs(self):
         self.retry_old_results()
 
         self.check_restart()
-
+        event_emitter = EventEmitter(None, None, self.client)
```
plars, line 208: This is created too early and could leak into the next job. Delay creation of this event_emitter until we actually start processing a job.

agent/testflinger_agent/client.py
```diff
@@ -365,3 +365,39 @@ def post_influx(self, phase, result=None):
             )
         except InfluxDBClientError as exc:
             logger.error(exc)
+
+    def post_status_update(self, job_queue, webhook, events):
```
plars, line 369: It's nice to add type annotations when we add new code at least. Can we go ahead and do that here like you did for EventEmitter?

plars (changes requested):

agent/testflinger_agent/runner.py
```diff
@@ -78,16 +80,16 @@ def post_output(self, data: str):
...
-    def check_stop_conditions(self) -> str:
+    def check_stop_conditions(self) -> Tuple[Optional[TestEvent], str]:
         """
         Check stop conditions and return the reason if any are met. Otherwise,
         return an empty string if none are met
         """
         for checker in self.stop_condition_checkers:
             output = checker()
-            if output:
+            if output[0] is not None:
```
plars, line 90: If we're going to do it this way, then I think we shouldn't just overload output to make it a tuple. Instead, maybe we could get `event, detail` = checker()` then you could specifically check `if event is not None:`...

agent/testflinger_agent/runner.py
```diff
@@ -78,16 +80,16 @@ def post_output(self, data: str):
...
                 return output
```
plars, line 91: You still need to return 2 values here, not just 1. When we are calling self.check_stop_condition() on line 142, it's expecting both the event and the stop_reason (which should maybe be "detail" or something since now the event also conveys the reason for it)

plars (changes requested):
Thanks for the updates! we should definitely go ahead and implement the things agreed to on the card unless you have further input on it.  A few more comments, but looking better :)

agent/testflinger_agent/agent.py
```diff
@@ -202,11 +205,18 @@ def process_jobs(self):
         self.retry_old_results()
 
         self.check_restart()
-
         job_data = self.client.check_jobs()
         while job_data:
             try:
+                event_emitter = EventEmitter(None, None, self.client)
```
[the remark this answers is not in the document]
val500: I thought I had a case where the Job construction threw an exception which caused the event_emitter to not exist in the catch/finally block, but looking at it again, I'm not sure if Job construction can throw an error.

agent/testflinger_agent/runner.py
val500: I think output here is two values based on the return of the stop condition checkers, but I can make that more explicit.

plars (changes requested):
Meeting notes: need `resource_url` added - we discussed adding a JOB_START event at the beginning like the JOB_END event we have at the end. The detail section for that JOB_START event can contain the `resource_url` which would just be `https://testflinger.canonical.com/jobs/<JOB_ID>`. That isn't going to change throughout the job, so there's no need to send it with every single event.

server/src/api/v1.py
```diff
@@ -489,6 +493,48 @@ def agents_post(agent_name, json_data):
     return "OK"
 
 
+@v1.post("/agents/status")
```
plars, line 496: After our chat today I was looking at how you would add the resource_url and realized that you don't have the job_id here anywhere. Then I got to thinking more about the path for this API and I'm not sure if `/v1/agents/status` is the right place (also the README is wrong but I'll comment on that separately). I think putting it in this path would hint that it's a status update about "agents in general". In the case of this API though, it's that at all. Instead, we're sending events that have happened in the context of a **job**. Sure, it has a specific agent associated with it, and a queue, but it's all data about what happened during the job itself. Also, associating it with a job_id means we can send that information across quite easily.

What do you think about putting it instead, at: `/job/<job_id>/events` or something like that? I'm open to other ideas, but I feel like this is at least a bit more consistent with what we already have, and conveys the meaning a bit better. Also, it's a pretty simple change, and it gets you the job_id that you're also going to want to insert as the `resource_url`

server/README.rst
```diff
@@ -341,3 +341,28 @@ This will find jobs tagged with both "foo" and "bar".
   .. code-block:: console
 
     $ curl -X GET http://localhost:8000/v1/agents/data
+
+**[POST] /v1/status - Receive job status updates from an agent and posts them to the specified webhook.
```
plars, line 345: See my comment about the path for this API in v1.py, and make sure the README is consistent with what we have there.

plars (changes requested):

server/src/api/v1.py
val500: Does the server need the job_id? If we're going to put the resource_url in the JOB_START event, it sounds easier to have the agent create the url with the job_id than have the server parse the events and create the url for the JOB_START event.

plars: @val500 the main thing I was looking for here was a small change to the endpoint which it looks like you took care of. However, I don't see where that resource_url is actually injected?

val500: It's in the detail for the JOB_START event

val500: I think the URL is actually incorrect here - would it be {server_name}/job/{job_id}?

plars: right now, it looks like you have `f"{self.client.server}/job/{job.job_id}/events"` but I think this is wrong, I was under the impression that this should be a link to the url for the job details page, which would be something like `f"{self.client.server}/jobs/{job.job_id}"` (note the "s" in "jobs" also). Let's sync up about this tomorrow as I think Omar also had some related comments on the TO side.

omar-selo: Yes we'd need the job link similar to https://testflinger.canonical.com/jobs/697b8930-3d88-48d0-8458-aafe2820e312 for example

plars (approved):
This looks like it should be good, and thanks for the video showing your testing of it against the TO changes on the other PR.

As discussed yesterday, please wait until the TO changes land so that we can land these PRs at the same time, and test them in staging together.

PR #266, opened 2024-05-01
[CERTTF-303] Handle case of file upload interrupted by the user

boukeas:
## Description

When file upload through the CLI was interrupted by the user, then the job was not automatically cancelled, causing it to appear indefinitely as `waiting` on the server. 

Now `KeyboardInterrupt` is caught and raises an `AttachmentError`, which in turn causes the job to be cancelled.

Also included are a couple of very minor refinements, mostly relating to improved logging.

## Tests

Submitted a job that involved a larger attachment and pressed Ctrl-C during the upload:
```
testflinger-cli --server https://testflinger-staging.canonical.com submit test-job-kernos.yaml 
^CJob 1736d348-32d0-48a3-a212-234060bb201d submitted and cancelled: failed to submit attachments
```
See [result](https://testflinger-staging.canonical.com/jobs/1736d348-32d0-48a3-a212-234060bb201d), verifying the job has been cancelled.

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -382,12 +382,12 @@ class TestflingerCli:
             with tempfile.NamedTemporaryFile(suffix="tar.gz") as archive:
                 archive_path = Path(archive.name)
                 # create attachments archive prior to job submission
-                logger.info(f"Packing attachments into {archive_path}")
+                logger.info("Packing attachments into %s", archive_path)
                 self.pack_attachments(archive_path, attachments_data)
                 # submit job, followed by the submission of the archive
                 job_id = self.submit_job_data(job_dict)
                 try:
-                    logger.info(f"Submitting attachments for {job_id}")
+                    logger.info("Submitting attachments for %s", job_id)
                     self.submit_job_attachments(job_id, path=archive_path)
                 except AttachmentError:
                     self.cancel(job_id)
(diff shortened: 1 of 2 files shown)


plars (commented):
Nice! that should take care of ^C at least, not sure about cases where it might get killed otherwise, but perhaps that's just a theoretical problem.
Would it be possible to add some tests for submit_job_attachment including ones for catching AttachmentError?

boukeas:
> Would it be possible to add some tests for submit_job_attachment including ones for catching AttachmentError?

@plars Were you thinking of something in addition to [`test_submit_attachments_retries`](https://github.com/canonical/testflinger/blob/main/cli/testflinger_cli/tests/test_cli.py#L148), [`test_submit_attachments_no_retries`](https://github.com/canonical/testflinger/blob/main/cli/testflinger_cli/tests/test_cli.py#L200) and [`test_submit_attachments_timeout`](https://github.com/canonical/testflinger/blob/main/cli/testflinger_cli/tests/test_cli.py#L248) that are already included in the tests?

plars:
@boukeas I saw those and they are ok. But I was thinking it would be good to do something that called submit_job_attachments and covered the error paths that can return AttachmentError if possible

plars (approved):
Ok, the tests we have here look like they could cover most of the error conditions we need already. Let's go ahead and land this and we can always add more later if needs be.

PR #238, opened 2024-03-27
Add: reboot control host from poe switch for muxpi devices

nancyc12:
## Description

This PR is based on https://github.com/canonical/testflinger/pull/222, replacing the `shutdown` command with PoE switch snmp command for non-Zapper control hosts (NanoPi). Hoping this can improve the stability of them and ensure a smooth provisioning.

It depends on the changes on [cc-lab-manager](https://github.com/canonical/certification-lab-manager/pull/20) and [cc-tool-box](https://code.launchpad.net/~nancy-chen/cc-tool-box/+git/cc-tool-box/+merge/463194) to generate `control_host_reboot_script` in default.yaml.

## Tests

Tested on several agents for device with NanoPi. 
Manually disable the power supply for `limerick-zcu106001`'s NanoPi before sending the testflinger job. The NanoPi is back and able to continue the provisioning.
```
**************************************************************

* Starting testflinger provision phase on limerick-zcu106001 *

**************************************************************
2024-03-27 10:00:19,943 limerick-zcu106001 INFO: DEVICE CONNECTOR: BEGIN provision
2024-03-27 10:00:19,943 limerick-zcu106001 INFO: DEVICE CONNECTOR: Provisioning device
2024-03-27 10:00:23,003 limerick-zcu106001 ERROR: DEVICE CONNECTOR: Error connecting to serial logging server
2024-03-27 10:00:23,003 limerick-zcu106001 INFO: DEVICE CONNECTOR: Rebooting control host
2024-03-27 10:00:23,004 limerick-zcu106001 INFO: DEVICE CONNECTOR: Running snmpset -c private -v 2c 10.102.x.x 1.3.6.1.2.1.105.1.1.1.3.1.2 i 2
iso.3.6.1.2.1.105.1.1.1.3.1.2 = INTEGER: 2
2024-03-27 10:00:23,012 limerick-zcu106001 INFO: DEVICE CONNECTOR: Running sleep 5
2024-03-27 10:00:28,045 limerick-zcu106001 INFO: DEVICE CONNECTOR: Running snmpset -c private -v 2c 10.102.x.x 1.3.6.1.2.1.105.1.1.1.3.1.2 i 1
iso.3.6.1.2.1.105.1.1.1.3.1.2 = INTEGER: 1
2024-03-27 10:00:56,091 limerick-zcu106001 ERROR: DEVICE CONNECTOR: Error connecting to serial logging server
[...]

PR #267, opened 2024-05-02
Fix sdwire detection issue

nancyc12:
## Description
This is an enhancement patch for https://github.com/canonical/testflinger/pull/238.
We have deployed this change on agents for RPIs machines for weeks and it seems to ease the sdwire detection issue as we power cycling NanoPi and DUT back to back.

## Tests
Have deployed them in real agents and run for weeks.

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
@@ -162,10 +162,6 @@ class MuxPi:
         self._run_control("true")
 
     def provision(self):
-        # If this is not a zapper, reboot before provisioning
-        if "zapper" not in self.config.get("control_switch_local_cmd", ""):
-            self.reboot_sdwire()
-
         # determine where to get the provisioning image from
         source = self.job_data["provision_data"].get("url")
         if source is None:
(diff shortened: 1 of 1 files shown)


kevinyehk (dismissed):
LGTM +1

jocave (commented):
I am pleased that testing seems to indicate this makes the setup more reliable, however it's not clear to me from looking at the code whether this guarantees that the control host and the DUT are off at the same time (something that I think we wanted to achieve to make sure the sdwire resets). The code seems to run through all the commands of the control host reboot script and then does the hardreset - could the control host be back on before the hard reset happens?

plars (dismissed):
Nice, it has definitely seemed a lot more stable over the last week or so. It will be good to get this rolled out to the devices. Thanks!

nancyc12:
@jocave We started with making the control host and the DUT off at the same time, but then we saw some cases that were not 100% fixing the sdwire issue. Have tried different power cycle orders and end up with this one as it is most stable.

jocave:
> @jocave We started with making the control host and the DUT off at the same time, but then we saw some cases that were not 100% fixing the sdwire issue. Have tried different power cycle orders and end up with this one as it is most stable.

Ok, thanks for the explanation

plars (approved):
Thanks for the rebase, +1

Issue #239, opened 2024-03-27
output_timeout not working properly


plars:
After some changes in the way the per-phase execution happens, the output_timeout command wasn't functioning properly. It was unfortunately timing out way too soon because it doesn't have an update mechanism that can be triggered to reset the timer whenever there's new output.

We've disabled it at the moment, because it could cause jobs to stop way too soon, and this is a much bigger problem than not having an output timeout (which may or may not actually be useful anyway).

If we want to continue having a timeout after long periods of no output in the future, then this needs to be reworked before we turn it back on.

plars:
Done and deployed

PR #272, opened 2024-05-07
Re-enable and fix output_timeout

plars:
## Description
When I reworked some of this a while back, I forgot to go back and add something that updates the OutputTimeoutChecker with the latest timestamp every time output occurs so that it can check to see if it's expired. I made this into an event that can be subscribed to. This is the only consumer of it for the moment but I thought it made for a nice way to organize it, and it's easy to extend in case this is useful for other things in the future.

## Resolved issues

#239 
CERTTF-301

## Documentation
N/A

## Tests
Tested on a live test run and also added a unit test

### DIFF
--- a/agent/testflinger_agent/job.py
+++ b/agent/testflinger_agent/job.py
@@ -19,7 +19,7 @@ import os
 import time
 
 from testflinger_agent.errors import TFServerError
-from .runner import CommandRunner
+from .runner import CommandRunner, RunnerEvents
 from .handlers import LiveOutputHandler, LogUpdateHandler
 from .stop_condition_checkers import (
     JobCancelledChecker,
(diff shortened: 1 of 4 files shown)


agent/testflinger_agent/job.py
```diff
@@ -94,6 +95,16 @@ def run_test_phase(self, phase, rundir):
...
 
+        # We only need to check for output timeouts during the test phase
+        if phase == "test":
+            output_timeout_checker = OutputTimeoutChecker(
+                self.get_output_timeout()
+            )
+            runner.register_stop_condition_checker(output_timeout_checker)
+            runner.subscribe_event(
+                "output_received", output_timeout_checker.update
+            )
```
omar-selo, line 106: Would be less error prone to define events somewhere (either as enum or just some global variables) then use them. Something like `RunnerEvents.output_received`

omar-selo (commented):
Looks good, just one suggestion

agent/testflinger_agent/job.py
plars: Good idea, done.

omar-selo (approved):
Looks good!

PR #275, opened 2024-05-14
Add a testflinger_common module with example usage

plars:
## Description
This adds a very small (for now) testflinger_common project for housing things like enums which might be useful across multiple subprojects. I only provided example usage in the agent for now, but we can expand it to also be used by server, etc.

## Resolved issues
N/A

## Documentation
Not much to document at the moment for it, but I made sure to include a README.rst

## Web service API changes
None, but using it in the server could help us to provide better guidance about the acceptable values for things like JobState.

## Tests
Passes Unit testing, and also deployed it locally. Found a problem with specifying a branch to deploy for testing in the charm config where it still would only pull main previously. I fixed that also here as a separate commit.

## Testing
Still not sure I understand why we have this section twice...


common/testflinger_common/enums.py
```diff
@@ -0,0 +1,45 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Job State and Test Phase Enums
+"""
+
+from strenum import StrEnum
```
boukeas, line 20: I haven't used `strenum` before but it might be worth checking if the behaviour of `strenum.StrEnum` is compatible with the `enum.StrEnum` available in Python 3.11 (in case the latter eventually replaces the former in our code). 

Even better, I wonder if it makes sense to derive our own `StrEnum` following the instructions [here](https://docs.python.org/3.8/library/enum.html#others). I don't think the definitions of `JobState` and `TestPhase` would change at all but we would avoid the dependency.

agent/testflinger_agent/agent.py
```diff
@@ -168,7 +169,11 @@ def unpack_attachments(self, job_data: dict, cwd: Path):
         # (so there is no interference with existing processes, especially
         # provisioning or firmware update, which are triggered when these
         # sections are not empty)
-        for phase in ("provision", "firmware_update", "test"):
+        for phase in (
```
boukeas, line 172: I really like the way this looks now.

common/testflinger_common/enums.py
val500: I was looking at this too and I think there was some breaking change when using class(str, Enum) 
https://blog.pecar.me/python-enum

plars: Yeah, good StrEnum support seems to only start around python 3.11 it seems. From what I gather, using the mixin is only possible after 3.10, and there are also [compatibility problems](https://github.com/python/cpython/issues/100458) with that, and with how it works >=3.11.
This strenum module seems to be the most recommended option I could find, and is supposed to still work even in later versions so that we could easily transition from it if there are any differences from using enum.StrEnum. It does mention that it's not a perfect "drop-in" replacement, but it sounds like the biggest difference is that enum.StrEnum does a casefold() to lower case when you use auto() (which I'm not using). So I suspect we'll be able to migrate pretty easily. Just to test this, I also tried switching it for enum.StrEnum and at least in my current python (3.12.3) it seems to work.

boukeas (approved):

val500 (approved):

PR #277, opened 2024-05-20
[CERTTF-327] Introduce `submit` GitHub action

boukeas:
Add the `submit` action for submitting Testflinger jobs to a Testflinger server.

The main input that the action expects is the job definition for the job that will be submitted to Testflinger (see the [Testflinger documentation](https://canonical-testflinger.readthedocs-hosted.com/en/latest/reference/job-schema/) to understand how this job definition is structured).

There are two ways to provide that job definition to the `submit` action: using a file through the `job-path` input or inline through the `job` input. 
```
description: Submit a job to a Testflinger server
inputs:
  job:
    description: Inline contents of a job file
    required: false
  job-path:
    description: Path to a job file
    required: false
```
The action will fail if neither one of those inputs are provided.

There are additional (optional) inputs to the action:
```
  poll:
    description: Specify if the submitted job should be tracked to completion
    required: false
    default: false
  dry-run:
    description: Specify if the job should really be submitted
    required: false
    default: false
  server:
    description: The Testflinger server to use
    required: false
    default: testflinger.canonical.com
```
Here are some example workflows (the Checkbox "canary test") that make use of this action:
- submit a job [by creating a job file and providing its path](https://github.com/canonical/checkbox/blob/test-action-fling-test/.github/workflows/checkbox-canary-test.yaml) to the `submit` action
- submit a job [by providing it inline as an argument](https://github.com/canonical/checkbox/blob/test-action-fling-test/.github/workflows/checkbox-canary-test-inline.yaml) to the `submit` action
- submit a job [without providing a job definition](https://github.com/canonical/checkbox/blob/test-action-fling-test/.github/workflows/checkbox-canary-test-empty.yaml), either through a file or inline, resulting in an error

### DIFF
--- a/.github/actions/submit/action.yaml
+++ b/.github/actions/submit/action.yaml
@@ -1,0 +1,52 @@
+name: Test-flinging
+description: Submit a job YAML file to a Testflinger server
+inputs:
+  job:
+    description: "Path to a job file (see https://canonical-testflinger.readthedocs-hosted.com/en/latest/reference/job-schema/ for more info)"
+    required: true
+  poll:
+    description: Specify if the submitted job should be tracked to completion
+    required: false
+    default: false
+  dry-run:
+    description: Specify if the job should really be submitted
+    required: false
+    default: false
+  server:
+    description: The Testflinger server to use
+    required: false
+    default: testflinger.canonical.com
+runs:
+  using: composite
+  steps:
+    - name: Install prerequisites
+      shell: bash
+      run: sudo snap install testflinger-cli jq
+
+    - name: Test connection to Testflinger server
+      shell: bash
+      run: nc -vz ${{ inputs.server }} 443
+
+    - name: Display contents of job file (for verification)
+      shell: bash
+      run: cat ${{ inputs.job }}
+
+    - name: Submit job to the Testflinger server
+      if: inputs.dry-run != 'true'
+      id: submit
+      shell: bash
+      run: |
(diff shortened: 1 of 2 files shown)


boukeas:
Note to @boukeas: Add `v1` tag to the (squashed) commit, after it is merged.

plars (commented):
This looks good, and thanks for the links to the canary runs where you tested it. But unless I'm missing something, it looks like they were all dry runs, right? Is there anything blocking us from setting us a test with dry-run set to false? It would be great to see a test where it actually runs and waits for the results if possible.

boukeas:
@plars The dry runs were just part of the testing but there were other tests as well. I have trouble navigating through the workflow runs myself as they are linked to commits.

Here's an [example of a run](https://github.com/canonical/checkbox/commits/test-action-fling-test/.github/workflows/checkbox-canary-test.yaml). In the "Submit job" section you can see the actual job ID and check [the results on Testflinger](https://testflinger.canonical.com/jobs/11ab8513-ab9f-459f-9e54-c5260bfadeaa).

What I _haven't_ tested to completion is using `poll: True`, which I cancelled half-way to completion (after the job was submitted). I can do that as well though and report back.

boukeas:
@plars Here's a [fresh run with polling](https://github.com/canonical/checkbox/actions/runs/9210331140/job/25337088184).

plars (approved):
Nice, thanks! Looks fine to me then. I'd say let's land it so that we can start trialing it for some things and iterate on it if we find anything it still needs.

mz2 (approved):
Looks great!

PR #281, opened 2024-05-28
Multi improve submit error reporting

plars:
## Description
The job_id won't be valid here, so don't try to use it. Also, try to capture more info in an error message if something goes wrong during the submission

## Resolved issues
N/A

## Documentation
N/A

## Web service API changes
N/A

## Tests
See below

## Testing
Hoping to test this in the server lab if possible, where we use this type of connector


nadzyah (approved):
Don't see anything to complain about. I like your improvements with enums!

PR #282, opened 2024-05-30
Report a nicer error if the user looks for a non-existant agent

plars:
## Description
It shouldn't come from a link but if someone tries to access an agent that doesn't exist directly through /agents/foo, for example, the will get an unhelpful error page with `Unhandled Exception: 'None' has no attribute 'updated_at'` and we'll get a similar error in the logs. Let's return something more informative that says this agent wasn't found.

## Resolved issues
N/A

## Documentation
N/A

## Web service API changes
Not an API change, just better visual feedback to the user on the rendered html

## Tests
Added a unit test

## Testing
Also tried it locally? Still not sure why we have two sections for this.

### DIFF
--- a/server/src/views.py
+++ b/server/src/views.py
@@ -47,6 +47,8 @@ def agents():
 def agent_detail(agent_id):
     """Agent detail view"""
     agent_info = mongo.db.agents.find_one({"name": agent_id})
+    if not agent_info:
+        return render_template("agent_not_found.html", agent_id=agent_id)
     return render_template("agent_detail.html", agent=agent_info)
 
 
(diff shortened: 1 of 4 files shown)


server/src/views.py
```diff
@@ -47,6 +47,8 @@ def agents():
 def agent_detail(agent_id):
     """Agent detail view"""
     agent_info = mongo.db.agents.find_one({"name": agent_id})
+    if not agent_info:
+        return render_template("agent_not_found.html", agent_id=agent_id)
```
nadzyah, line 51: Does it return a 404 HTTP code?

plars: I was considering that also, was kinda on the fence about it because it wouldn't be seen by the user, but it seems more complete to do it that way. I've added that too.

nadzyah (approved):

PR #290, opened 2024-06-20
Refine submit action check

boukeas:
## Description

The `submit` composite action used `nc` (netcat) to check if the Testflinger server was listening at the appropriate port. This was a simple test that often failed to detect connectivity issues that would later cause the `submit` action to fail (e.g. proxy problems), partly because `nc` operates at a lower protocol level (TCP) than the actual HTTPS Testflinger requests.

In this PR, the `nc` test is replaced by a more refined one that uses `curl` to check if an HTTPS connection can be established with the Testflinger server.

## Documentation

No changes to the documentation are required.

## Tests

The action was tested:
- in a workflow executing on a self-hosted Github runner with proxy-related connectivity issues ([sample run](https://github.com/canonical/hwcert-jenkins-tools/actions/runs/9600407702/job/26476886011)).
- in a workflow executing a self-hosted Github runner with no connectivity issues ([sample run](https://github.com/canonical/hwcert-jenkins-tools/actions/runs/9600407702/job/26603311087)).

### DIFF
--- a/.github/actions/submit/action.yaml
+++ b/.github/actions/submit/action.yaml
@@ -38,7 +38,7 @@ runs:
           cat error.log
           exit $STATUS
         else
-          echo "Successful server ping at $SERVER, status: $status"
+          echo "Successful server ping at $SERVER, status: $STATUS"
         fi
     - name: Create job file, if required
       id: create-job-file
--- a/.github/actions/submit/action.yaml
+++ b/.github/actions/submit/action.yaml
@@ -27,13 +27,15 @@ runs:
       env:
         SERVER: https://${{ inputs.server }}
       run: |
-        STATUS=$(curl -I -w "%{http_code}\n" -o /dev/null -s $SERVER/jobs)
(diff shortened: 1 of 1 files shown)


plars (dismissed):
This looks fine to me and we've discussed this already too. One thing I was thinking, though I'm not sure how useful it would be in practice, is whether we want verbose output from the curl just in case something goes wrong? I quick test locally shows that if you use `-Ivw ...` for the args, then you'll get verbose output to STDERR. So setting `STATUS=` still works fine and you would just see the debug output under the twistie if you care to do so, but the STATUS var would still hold the HTTP code.

Anyway, if you think it's useful then it might be worth adding. Just something to think about.  Otherwise +1

boukeas:
@plars Thanks for the review and the suggestion.

I updated the PR, adding the verbose option and also capturing standard error to a file, so that the diagnostics are only displayed in case of an issue.

plars (approved):
Even better, very nice!

PR #291, opened 2024-06-21
Auto deploy tf-cmd-scripts in agent-host charm

nancyc12:
## Description
The [tf-cmd-scripts](https://git.launchpad.net/~canonical-hw-cert/hwcert-jenkins-tools/+git/tf-cmd-scripts/tree/) are a set of scripts deployed on agent hosts in HWCert labs. These scripts are used by the `testflinger-agent` to trigger the `testflinger-device-connector` at each phase. Previously, whenever the scripts were updated, a manual pull from the Launchpad Git repository to each agent host was required. This process has become inefficient as the number of labs has increased.

This PR aims to integrate tf-cmd-scripts into the mono repository and automate their deployment in the agent host environment.

## Resolved issues

https://warthogs.atlassian.net/browse/LM-1616?atlOrigin=eyJpIjoiNGE5YTkxMzY5MjkzNDllZTkzZTUyNzYyZmNkZTc4MDgiLCJwIjoiaiJ9

## Documentation

Update README for testflinger-agent-host-charm to introduce tf-cmd-scripts. Also, add a description for Building and Configuration.
Update README for testflinger-agent-charm to reflect the current Configuration.

## Tests

`charmcraft pack` to build the testfligner-agent-host charm. Juju deploy the charm on my laptop and validate all scripts are copied and executable. Also tested the `upgrade-charm` with another local charm packed with modified tf-cmd-scripts and verified the changes were landed in the application.

### DIFF
--- a/agent/charms/testflinger-agent-host-charm/src/charm.py
+++ b/agent/charms/testflinger-agent-host-charm/src/charm.py
@@ -35,6 +35,7 @@ class TestflingerAgentHostCharm(CharmBase):
         self.framework.observe(self.on.install, self.on_install)
         self.framework.observe(self.on.start, self.on_start)
         self.framework.observe(self.on.config_changed, self.on_config_changed)
+        self.framework.observe(self.on.upgrade_charm, self.on_upgrade_charm)
         self._stored.set_default(
             ssh_priv="",
             ssh_pub="",
@@ -49,7 +50,7 @@ class TestflingerAgentHostCharm(CharmBase):
         # maas cli comes from maas snap now
         self.run_with_logged_errors(["snap", "install", "maas"])
         self.setup_docker()
-        self.copy_tf_cmd_scripts()
+        self.update_tf_cmd_scripts()
 
     def setup_docker(self):
         self.run_with_logged_errors(["groupadd", "docker"])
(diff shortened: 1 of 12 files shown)


agent/charms/testflinger-agent-host-charm/src/tf-cmd-scripts/tf-allocate
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+AGENT=$(echo $agent_id | sed 's/-\([0-9]*\)$/\1/g')
+PROVISION_TYPE="$provision_type"
+
+if [ -f /srv/testflinger-agent/$AGENT/testflinger-agent.conf ]; then
+        . /srv/testflinger-agent/$AGENT/env/bin/activate && PYTHONUNBUFFERED=1 testflinger-device-connector $PROVISION_TYPE allocate -c /srv/testflinger-agent/$AGENT/default.yaml testflinger.json
```
plars, line 7: These scripts were changed a while back to handle the transition to monorepo and putting the config files under `/srv/testflinger-agent/$AGENT`. So we can make them much simpler now, and just keep the first part that runs if that config file exists, and drop the conditional.

plars (changes requested):

agent/charms/testflinger-agent-host-charm/src/tf-cmd-scripts/tf-allocate
nancyc12: Thanks! It's been updated.

nancyc12:
@plars I'm wondering if there might be a chance to modify these scripts in the future. What do you think? Should an action be added to cope with it?

plars:
> @plars I'm wondering if there might be a chance to modify these scripts in the future. What do you think? Should an action be added to cope with it?

I would not recommend doing it as an action. Instead, I think we should always copy the scripts whenever it's installed or when the charm itself is upgraded.

plars (changes requested):
Looking good, a couple of small things:
1. It's definitely good to copy them during on_install but let's also copy them during the [upgrade-charm](https://juju.is/docs/sdk/upgrade-charm-event) event. You'll need to add a new event handler for that. This way if we ever change the scripts, then we can just upgrade the charm, since we'll need to do that anyway to get the new version of the scripts
2. Can you confirm in testing that they get the execution bit set/copied correctly when it copies them to the destination dir? It might even be good to set this, just to protect ourselves from the possibility that we didn't have it set in the original file that it copies from.

agent/charms/testflinger-agent-host-charm/src/charm.py
```diff
@@ -81,13 +82,24 @@ def copy_ssh_keys(self):
...
             self.write_file("/home/ubuntu/.ssh/id_rsa.pub", pub_key)
 
-    def copy_tf_cmd_scripts(self):
+    def update_tf_cmd_scripts(self):
+        """Update tf_cmd_scfipts"""
         tf_cmd_dir = "src/tf-cmd-scripts/"
+        usr_local_bin = "/usr/local/bin/"
+        for file in os.listdir(usr_local_bin):
+            if "tf" in file:
+                os.remove(os.path.join(usr_local_bin, file))
```
plars, line 91: I don't think there's a need to remove these. shutil.copy() will overwrite them. Also, although it's unlikely, you could accidentally remove some unrelated binary that happens to have "tf" in the filename.

plars (changes requested):
one very minor thing, then I think it's ok. Great work!

plars (approved):
+1

PR #295, opened 2024-06-28
Nicer handling of job not found on the job details view

plars:
## Description
You won't get there accidentally by clicking, but if you try to go to the job details page for an invalid job_id, you'll get a confusing error response of `Unhandled Exception: 'None' has no attribute 'job_data' ` and it'll hit the server logs as an error because we aren't handling it. This just makes a nicer 404 page for it like we did for the similar situation with agent details

## Resolved issues
N/A

## Documentation
N/A

## Web service API changes
N/A

## Tests
Added unit test and also tested locally

### DIFF
--- a/server/src/templates/job_not_found.html
+++ b/server/src/templates/job_not_found.html
@@ -1,0 +1,11 @@
+{% extends 'base.html' %}
+{% set active_page = 'jobs' %}
+{% block content %}
+<div class="p-strip is-shallow">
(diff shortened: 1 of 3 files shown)


nancyc12 (dismissed):
Thanks for the changes! It looks good to me. Just added a small suggestion.

nancyc12 (approved):
LGTM+1

PR #300, opened 2024-07-02
Fix missing exit code if we fall through to the finally block

plars:
## Description
Unlikely and hard to trigger, but if we hit an error here and then drop to the finally block, then we will be missing some data that it needs when it tries to update the phase results.

## Resolved issues

Described above. There's a small flake8 thing that I worked around for now because I think it's useful to include the exception in the details if we hit this. When the PR from @val500 lands, then we can remove the `noqa` comment.

## Documentation

N/A

## Web service API changes
N/A

## Tests
Should be obvious from inspection, but also added a unit test


nadzyah (approved):
Nice change, nothing questionable here

PR #302, opened 2024-07-03
Fix an issue where we might not post the provision log unless it fails

plars:
## Description
When testing on staging, I noticed that the provision log might only get updated if the provision fails. That's not quite what we want, we also want to record the successful attempts. I updated this in staging also and confirmed it was showing both successful and unsuccessful provision attempts:
image

## Resolved issues
see above

## Documentation
see above

## Web service API changes
None

## Tests
See above


asbalderson (approved):
+1

PR #308, opened 2024-07-11
support assign timeout to device system boot checking

kiya956:

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/oemrecovery/oemrecovery.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/oemrecovery/oemrecovery.py
@@ -184,9 +184,19 @@ class OemRecovery:
     def check_device_booted(self):
         """Check to see if the device is booted and reachable with ssh"""
         logger.info("Checking to see if the device is available.")
+        try:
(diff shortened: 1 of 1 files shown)


device-connectors/src/testflinger_device_connectors/devices/oemrecovery/oemrecovery.py
```diff
@@ -184,9 +184,19 @@ def copy_ssh_id(self):
     def check_device_booted(self):
         """Check to see if the device is booted and reachable with ssh"""
         logger.info("Checking to see if the device is available.")
+        try:
+            boot_timeout = self.job_data.get("test_data", {}).get(
```
plars, line 188: This is not something generally associated with the test phase, so the test_data section is probably the wrong place for it. We don't currently support an option for this, but if we did, it would probably go in either provision_data or up at the top level.

plars (changes requested):
I don't think this is how we would want to achieve this. It also lacks documentation and none of the fields on the PR were filled in. Having the full information about it would really help us in the review so that we can give it better consideration with regards to your goals. Thanks!

PR #310, opened 2024-07-23
Provision error logs

val500:
## Description

This adds logging of provisioning errors in the device connector to be read by the Testflinger agent and propagated to the event log. The device connector logs errors for each phase in a file, device-connector-error.json, which is read by the agent. This JSON takes the following format:
```
{
    {phase}_exception_info: {
        exception_name: String
        exception_message: String,
        exception_cause: String
    } 
    ...
}
```
## Resolved issues
Resolves https://warthogs.atlassian.net/browse/CERTTF-359

## Tests
Adds new tests to the device connector which mocks the MAAS CLI to throw an exception and adds tests to the agent which mocks the running of the provisioning command.


agent/testflinger_agent/client.py
```diff
@@ -429,6 +429,5 @@ def post_status_update(
             job_request = None
         if not job_request:
             logger.error(
-                "Unable to post status updates to: %s (error: %s)"
-                % (status_update_uri, job_request.status_code)
+                f"Unable to post status updates to: {status_update_uri}"
```
plars, line 432: There are some exceptions, but in general, it's usually a bad habit to use f-strings in logging messages. Also, this dropped the detail about the status code from the error - any reason for that?

device-connectors/src/testflinger_device_connectors/devices/__init__.py
```diff
@@ -54,11 +54,27 @@
...
-    pass
+    def __init__(self, message: str = None):
+        if message is not None:
+            with open(
+                "provision_error.log", "w", encoding="utf-8"
+            ) as provision_error_file:
+                provision_error_file.write(message)
+                super().__init__(message)
+        else:
+            super().__init__()
```
plars, line 65: I don't think I would handle writing the error out in the exception class itself. Instead, catch the error up at a higher level and handle the write once, along with the type of error, etc. It might make better sense to save it as a json file so that you can be flexible and capture more than just a single detailed err - then you could specify the type of error, and other details as needed? Just a thought.

plars (changes requested):
I know it's just a draft and it'll be easier to review when the previous stuff lands, but I wanted to go ahead and start taking a look. Just a few comments for now but I'll take a deeper look later as it progresses.

agent/testflinger_agent/client.py
val500: I think this should be in a different PR - but I caught a bug here where if there is a RequestException and job_request is None, that status_code doesn't exist.

device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
```diff
@@ -70,19 +71,22 @@ def recover(self):
...
+            if self.config.get("clear_tpm"):
+                self.clear_tpm()
+            provision_data = self.job_data.get("provision_data")
+            # Default to a safe LTS if no distro is specified
+            distro = provision_data.get("distro", "xenial")
+            kernel = provision_data.get("kernel")
+            user_data = provision_data.get("user_data")
+            storage_data = provision_data.get("disks")
+            self.deploy_node(distro, kernel, user_data, storage_data)
+        except Exception as err:
```
plars, line 87: I wonder if we could catch this at an even higher level? We'd like to do something similar for other device connectors eventually, so it would be really nice if we didn't have to repeat this part of the work. If it fails, then we may have some steps here where we *want* to catch the exception, do some cleanup, and re-raise it. But if we catch it up at the point where we run the subcommand, or the command as a whole, then we can catch it for everything - possibly even errors that we haven't thought of.

agent/testflinger_agent/tests/test_agent.py
```diff
@@ -536,3 +536,71 @@ def test_post_agent_status_update_output_timeout(
         event_list = status_update_requests[-1].json()["events"]
         event_name_list = [event["event_name"] for event in event_list]
         assert "output_timeout" in event_name_list
+
+    def test_provision_error_in_event_detail(self, agent, requests_mock):
```
plars, line 540: Same issue here - make sure to include docstrings, especially for tests to describe what it's intending to test.

agent/testflinger_agent/agent.py
```diff
@@ -278,11 +306,15 @@ def process_jobs(self):
...
-                        event_emitter.emit_event(exit_event)
+                        detail = ""
                         if phase == "provision":
                             self.client.post_provision_log(
                                 job.job_id, exit_code, exit_event
                             )
+                            detail = parse_provision_error_logs(
+                                provision_error_log_path
+                            )
+                        event_emitter.emit_event(exit_event, detail)
```
plars, line 317: I don't see any reason we should limit this to *only* happen during the provision phase. Yes, we're focusing on the provision phase of maas for now, because that's a big target for us. But there's a chance that we could capture other kinds of unexpected errors at runtime from other phases if we just do this as a generic step for each phase and add it to the detail if we find some exception that gets propagated back to the agent in this way. It's certainly much MORE likely to happen during the provision phase, but anything is possible, and it might help us later if we unrestrict it.

device-connectors/src/testflinger_device_connectors/devices/__init__.py
```diff
@@ -61,6 +61,20 @@ class RecoveryError(Exception):
     pass
 
 
+def log_provision_error(exception: Exception):
+    exception_info = {
+        "exception_info": {
+            "exception_name": type(exception).__name__,
+            "exception_message": str(exception),
+            "exception_cause": repr(exception.__cause__),
```
plars, line 69: Just out of curiosity, what kinds of things have you managed to catch in the __cause__ field when you tried this with exceptions from maas? From what I understand, in other circumstances (chaining exceptions) it might be more beneficial to check `__context__` if `__cause__` isn't present. Also, I wonder if `__doc__` might actually prove more useful? It *should* normally be more human readable, but might lack details that cause or context might have if they exist at all.

plars (changes requested):
Don't forget to fill in the PR details, especially doc/test sections

device-connectors/src/testflinger_device_connectors/devices/__init__.py
val500: I used cause here in order to capture exceptions like MaasStorageError which seemed to be raised as `raise ProvisioningError from MaasStorageError`

I'll take a look at doc and context.

agent/testflinger_agent/client.py
plars: Interesting, do you know how to reproduce it? I thought it would always return a response object, even on error conditions, but perhaps RequestException is not like that. Feel free to submit a separate PR for this if you have a quick way to test that.

plars (changes requested):
In the process of trying to do some testing with a real maas device using this change, I encountered something strange. I decided to see if giving it a bad image name would be enough to make it fail properly. The provisioning step definitely failed... but then it kept on going through the job as if nothing had happened, when it should have stopped at that point. I switched back to main and confirmed that main still has the correct behavior.

I haven't had a chance to run down what happened exactly, but I think we need to do some more testing and figure that out

plars:
@val500 Ok, I looked into this a bit more and I see the problem. In __init__.py of the maas2 device connector (and possibly others, you will want to look at those also!), we seem to specifically intercept the ProvisioningError so that we can add a nice log message about it, but then instead of re-raising it, we log an error. IIRC this was likely intended to avoid getting a lot of extra debug output for common issues, but in this case we need to expose that back up to the  exception handler that will write the output so that it gets propagated to the agent.

We don't want to show the full traceback in the user logs if we can avoid it - an nicer error message there would suffice. But we still need to write that output file with the full details.

agent/testflinger_agent/agent.py
```diff
@@ -67,6 +67,29 @@ def secure_filter(member, path):
     return tarfile.data_filter(member, path)
 
 
+def parse_error_logs(error_log_path: str, phase: str):
+    with open(error_log_path, "r") as error_file:
```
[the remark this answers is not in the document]
val500: I create/clear the the error file before all of the test phases run in order to prevent this/avoid old error data in the log.

agent/testflinger_agent/agent.py
```diff
@@ -67,6 +67,29 @@ def secure_filter(member, path):
...
+                    exception_info["exception_message"],
+                )
+            else:
+                detail = "%s: %s caused by %s" % (
+                    exception_info["exception_name"],
+                    exception_info["exception_message"],
+                    exception_info["exception_cause"],
+                )
+            return detail
+        except ValueError:
```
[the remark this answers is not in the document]
val500: If the JSON is malformed in some way, ie. we get a key error or json decode error, what would make sense to report in the detail? It should still be a ProvisionFail event(or other phase fail event), but with no extra detail in the event.

plars (approved):
Thanks for the updates, this seems to be working for me now. I've tried it with the maas connector and also with some others like muxpi since those had to be altered. I also tried forcing a recovery error to ensure that this still works properly so I think it's doing the right things now. The one thing I did notice is that if the error output is REALLY long, then the UI shows up in kind of a strange way on the test-observer side. We might be able to iterate on distilling it down to the more usable message or alternatively, making it display nicer when this happens, but I think that will be for another time. It'll be more clear once we see what happens in real failure situations.
+1 thanks!

PR #311, opened 2024-07-25
Documentation Link in Dashboard Header

val500:
## Description
For CERTTF-334. This change adds a link to the testflinger documentation([https://canonical-testflinger.readthedocs-hosted.com](https://canonical-testflinger.canonical.com/)) in the dashboard.

### DIFF
--- a/server/src/templates/base.html
+++ b/server/src/templates/base.html
@@ -18,6 +18,7 @@
     {'name': 'Agents', 'url': '/agents'},
     {'name': 'Jobs', 'url': '/jobs'},
     {'name': 'Queues', 'url': '/queues'},
+    {'name': 'Documentation', 'url': 'https://canonical-testflinger.readthedocs-hosted.com'},
     ] -%}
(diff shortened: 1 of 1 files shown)


plars (approved):
I loaded this up locally and it looks ok and redirects fine. I'm not sure if this is the preferred url or if https://testflinger.readthedocs.io/en/latest/ is though. Might be worth double-checking with @tang-mm before landing, but otherwise, looks fine to me. +1

tang-mm:
> I loaded this up locally and it looks ok and redirects fine. I'm not sure if this is the preferred url or if https://testflinger.readthedocs.io/en/latest/ is though. Might be worth double-checking with @tang-mm before landing, but otherwise, looks fine to me. +1

The implemented one `readthedocs-hosted.com` is preferred as it's hosted on RTD commercial version, providing more configurations in the project dashboard. Both URLs are built from the same source though. Thanks for checking!

PR #312, opened 2024-07-26
Exit when we know we're unable to get the list of queues from the server

plars:
## Description
Fix up a corner case where you could get a traceback if there's a problem connecting to the server on list-queues

Also some opportunistic cleanups to move to sys.exit() (which is preferred) instead of raising SystemExit. Still tried to maintain consistency with prints and sys.exit(str) used for messages aimed at the user, and logging used for things that are typically going to be things like connection problems and things like that, so that there's more debug information in those.

## Resolved issues
See above

## Documentation
N/A

## Web service API changes
N/A

## Tests
Added unit tests


omar-selo (approved):
Looks good

PR #319, opened 2024-07-31
Add api and view for queue wait time percentiles

plars:
## Description

Store wait times for queues and add an API as well as a view on the queue_detail page that shows the percentiles for these wait times.  The API can be used to retrieve metrics on individual queue(s) or for all queues at once. For the API, the wait times are returned in seconds, but on the view, we convert this to a more human readable form (see screenshot)

image

## Resolved issues

CERTTF-333

## Documentation

Added README section about the new API

## Web service API changes
```
**[GET] /v1/queues/wait_times** - get wait time metrics - optionally take a list of queues

- Parameters:

  - queue (array): list of queues to get wait time metrics for

- Returns:

  JSON mapping of queue names to wait time metrics

- Example:

  .. code-block:: console

    $ curl http://localhost:8000/v1/queues/wait_times?queue=foo\&queue=bar%     
```
This adds a new database collection for queue_wait_times. For displaying a single queue, I expect this to be quite fast and have no noticeable effect on page load time for queue_details. I did some testing of the code that does the calculation of the percentiles and it was quite fast on my system - potentially it could be slower across thousands of queues if we frequently call the API with no queues specified to get all queues, but we could consider capping the number of data points per queue if this becomes a problem. The space consumed is small enough and the performance is good enough that I've left it uncapped for now.

## Tests

Tested locally and unit tests added. The wait time data doesn't exist until this is deployed on the server though (no agent changes needed), so we won't start collecting the data until then and it won't display until it has data for a given queue.

### DIFF
--- a/server/src/database.py
+++ b/server/src/database.py
@@ -246,21 +246,34 @@ def get_queue_wait_times(queues: list[str] | None = None) -> list[dict]:
 
 
 def calculate_percentiles(data: list) -> dict:
-    """Calculate the percentiles of the wait times for each queue"""
+    """
+    Calculate the percentiles of the wait times for each queue
+
+    This uses the nearest rank with interpolation method, which can help in
+    cases where the data is not evenly distributed, such as when we might
+    have big gaps between the best and worst case scenarios.
+    """
     if not data:
         return {}
     percentiles = [5, 10, 50, 90, 95]
     data.sort()
     percentile_results = {}
     for percentile in percentiles:
+        # Index is the position in our sorted data that goes with this
+        # percentile
         index = (len(data) - 1) * (percentile / 100.0)
+        # Lower and upper index are the indexes of two closest data points
+        # that are below and above the percentile
         lower_index = int(index)
         upper_index = lower_index + 1
         if upper_index < len(data):
+            # Interpolate the value based on how far the lower and upper
+            # data points are from the percentile
             lower_value = data[lower_index] * (upper_index - index)
             upper_value = data[upper_index] * (index - lower_index)
             percentile_results[percentile] = lower_value + upper_value
         else:
+            # If the upper index is out of bounds, just use the lower value
             percentile_results[percentile] = data[lower_index]
     return percentile_results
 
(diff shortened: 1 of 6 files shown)


server/src/database.py
```diff
@@ -196,9 +201,70 @@ def pop_job(queue_list):
...
+    for percentile in percentiles:
+        index = (len(data) - 1) * (percentile / 100.0)
+        lower_index = int(index)
+        upper_index = lower_index + 1
+        if upper_index < len(data):
+            lower_value = data[lower_index] * (upper_index - index)
+            upper_value = data[upper_index] * (index - lower_index)
+            percentile_results[percentile] = lower_value + upper_value
+        else:
+            percentile_results[percentile] = data[lower_index]
```
nadzyah, line 264: I guess it would be nice to explain how you calculate the values in the doc-string. This maths is not pretty obvious

server/src/views.py
```diff
@@ -196,6 +196,30 @@ def queue_detail(queue_id):
...
+        queue_percentile_data=queue_percentile_data,
     )
+
+
+def seconds_to_hms(seconds: int) -> str:
+    """Convert seconds to a human-readable string"""
+    seconds = int(seconds)
+    hours = seconds // 3600
+    minutes = (seconds % 3600) // 60
+    seconds = seconds % 60
```
nadzyah, line 224: Also here, but not necessary

server/src/database.py
plars: Yeah, it might be helpful to have this. I added some more comments around this.

server/src/views.py
plars: I think this one is less important and pretty easy to understand from the existing function name and comments? we're just taking something like: 60.0 seconds and converting it to "00h 01m 00s" so that it's easier to read. Is there something that would make it better?

nadzyah (approved):
Thanks for adding the explanation!

server/src/views.py
nadzyah: I thought about explaining the minutes value, but it is not that hard to understand, so we can keep it as is

PR #322, opened 2024-08-05
fix incorrect formatting for file not found error in submit

andersson1234:
I noticed this when doing some testing with testflinger. It seems what is meant to be an fstring isn't actually formatted. Not sure if this is intended but I thought I'd submit a PR anyway.

This is the current behaviour:
```
File not found: {self.args.filename}
```

## Description
Changes what is seemingly meant to be an fstring, into an fstring

## Resolved issues
None

## Documentation
None necessary

## Web service API changes
Nope!

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -371,7 +371,7 @@ class TestflingerCli:
                 ) as job_file:
                     data = job_file.read()
             except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


plars (approved):
Good catch, thanks! +1

plars:
/canonical/self-hosted-runners/run-workflows e5994316e00b7f78b637d35cda6f7e768d24e58a

PR #324, opened 2024-08-06
Change: additional arguments/assertions for UC24 and OEM 24.04 via Zapper KVM

p-gentili:
## Description

This PR addresses a few problems detected while provisioning a device with UC24-subiquity and Desktop 24.04 OEM.

About UC24:
- being a non-dangerous grade image, the installed user will be given by `ubuntu_sso_email`. This PR forwards the `ubuntu_sso_email` argument to the zapper.provision() API and assert that `test_username` matches the local part of the SSO email. The DUT pulls TF SSH key from the account, so it's accessible and reservation works fine.

About 24.04 OEM:
- `user_data_base` was intended as a plain string, making it difficult to include in a TF job yaml. Taking inspiration from the MAAS connector I made it a base64 encoded field. Added tests and assertions.
- being a common case, instead of asking the user to pass the `user_data_base` argument, I added a dedicated `autoinstall_oem` boolean flag. It will be handled on the Zapper side as described in the autoinstall docs. See https://github.com/canonical/zapper/pull/311 for reference on what it actually does on the Zapper side.

## Resolved issues

Resolves [ZAP-873](https://warthogs.atlassian.net/browse/ZAP-873)
Part of [ZAP-776](https://warthogs.atlassian.net/browse/ZAP-776)

## Documentation

Updated README and `docs/`.

## Web service API changes

N/A

## Tests

About **UC24**: couldn't really test it e2e but, once manually provisioned via Zapper API, the device is accessible as usual by TF with:
```
~ cat uc24.yaml       
job_queue: 202101-28611
**test_data:
  test_username: ce-certification-qa**
reserve_data:
  ssh_keys:
    - lp:pgentili
  ```

About 24.04 OEM: tested on the Zapper side as part of https://github.com/canonical/zapper/pull/311 passing `oem=True`.

[ZAP-873]: https://warthogs.atlassian.net/browse/ZAP-873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
@@ -62,7 +62,11 @@ class DeviceConnector(ZapperConnector):
         if "storage_layout" not in provision:
             return None
 
-        autoinstall_conf = {"storage_layout": provision["storage_layout"]}
+        autoinstall_conf = {
+            "storage_layout": provision["storage_layout"],
+            "oem": provision.get("autoinstall_oem", False),
+        }
+
         if "base_user_data" in provision:
             self._validate_user_data(provision["base_user_data"])
             autoinstall_conf["base_user_data"] = provision["base_user_data"]
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
@@ -101,17 +101,6 @@ class DeviceConnector(ZapperConnector):
             )
             retries = self.job_data["provision_data"].get("robot_retries", 1)
 
-        # If a SSO email is specified, e.g. UC, username must match
-        # the local-part because that would be the only user available
-        # on the DUT after provisioning.
-        if "ubuntu_sso_email" in self.job_data["provision_data"]:
-            email = self.job_data["provision_data"]["ubuntu_sso_email"]
-            if username != email.split("@")[0]:
-                raise ProvisioningError(
-                    "Test username doesn't match the provided "
-                    "ubuntu_sso_email."
-                )
-
         provisioning_data = {
             "url": url,
             "username": username,
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
@@ -101,6 +101,17 @@ class DeviceConnector(ZapperConnector):
             )
             retries = self.job_data["provision_data"].get("robot_retries", 1)
 
+        # If a SSO email is specified, e.g. UC, username must match
+        # the local-part because that would be the only user available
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py [hunk 1]
```diff
@@ -77,6 +101,17 @@ def _validate_configuration(
             )
             retries = self.job_data["provision_data"].get("robot_retries", 1)
 
+        # If a SSO email is specified, e.g. UC, username must match
+        # the local-part because that would be the only user available
+        # on the DUT after provisioning.
```
plars, line 106: I don't think this is correct. The username is the launchpad user associated with that sso login and isn't necessarily the same as what comes before the "@" in the email. For example, Mine is `paul.larson@...` for the email, but the resulting user created in Ubuntu Core is `pwlars`

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
```diff
@@ -36,15 +39,36 @@ class DeviceConnector(ZapperConnector):
...
+            raise ProvisioningError(
+                "Provided `base_user_data` is not a valid YAML."
+            ) from exc
+
     def _get_autoinstall_conf(self) -> Optional[Dict[str, Any]]:
         """Prepare autoinstall-related configuration."""
         provision = self.job_data["provision_data"]
 
         if "storage_layout" not in provision:
             return None
```
plars, line 63: Just to confirm, if no storage_layout is specified, then this means that autoinstall is not supported? If that's not the case, then I think this check needs fixing. Otherwise, it might be nice to make that more clear in the documentation. For example, is there some default storage layout you can accept and then still be able to specify params like `oem`?

plars (changes requested):
Overall, looks really great! Just a couple of comments

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
p-gentili: oh, TIL. I think I learned from the wrong example then (ce-certification-qa). So we can't do much except for making it clear in the documentation.

device-connectors/src/testflinger_device_connectors/devices/zapper_kvm/__init__.py
p-gentili: You're perfectly right, it doesn't make much sense to rely on that single argument. In https://github.com/canonical/testflinger/pull/324/commits/581686fed35a1dbfda9c9a38a0534ebd3c0d2550 I made it more generic: any `autoinstall`-related argument is taken in consideration and possibly validated. Yes, a couple of arguments name have changed, but I think it makes them clearer. There are no job on jenkins yet, it shouldn't be a problem having this breaking change in naming.

PR #326, opened 2024-08-11
Small fixes for the agent-host charm tests

plars:
## Description
It looks like at one time there was an intent to remove the extraneous files in /usr/local/bin when installing the tf-* scripts, but that is no longer the case (understandable). Also, we need a slightly different environment to set up the tests for the charm, so let's add that here. Especially since we're in the process of expanding this charm and making it the only one for the agent, it would be good to make sure things are working correctly.

## Resolved issues
See above

## Documentation
N/A

## Web service API changes
N/A

## Tests
That's the point :)


nancyc12 (approved):
Thanks for fixing this! The tests should have been changed along with [this commit](https://github.com/canonical/testflinger/pull/291/commits/e5e3e48fbc09ca6ec83651cdd5efa83442f1f1e4), but I missed to update it.

PR #327, opened 2024-08-12
Add github token secret to the workflow for uploading agent charms

plars:
## Description
It looks like the upload-charm needs the github token even if you don't pull an oci container from ghcr.io. It seems to use it for tagging. I'm not sure how the current set of tags it seems to have used really line up with the charm releases, but there also seems to be a tag prefix you can use that helps differentiate them here so I added that also.

## Resolved issues
Charm fails to upload without the token at least

## Documentation
N/A

## Web service API changes
N/A


boukeas (approved):
+1

PR #328, opened 2024-08-12
Use config options instead of resource files for ssh keys

plars:
NOTE: Trying something new here... this is the first of several changes to implement CERTTF-374 so that we can unify the charms into one charm that configures, deploys, and updates all the agents. We *could* just do ALL the changes, and push them in one PR, but then you'll have a big PR that changes everything to review all at once. Instead, I thought it might be nice to try this as a feature branch. This charm doesn't change much otherwise, and this keeps the PRs small, focused, easy to review. Once we're done, we can merge the whole thing, because it will almost certainly break the way we do things now. So it doesn't make sense to do it in pieces against main.

## Description
The existing testflinger-agent-host-charm uses a juju “resource” for the ssh keys. This seemed like a good fit when we first implemented it because it allows sending whole files. However, the it looks like juju now intends for this to be used more for globally shared resources like OCI images rather than per-application configuration, because it REQUIRES uploading a version of the resource when you update the charm. The more idiomatic way of managing configs and keys that we see in other charms it to use config items, so we should use that instead for these keys.

## Resolved issues
CERTTF-375

## Documentation
The README was already saying that these were config options instead of resource files, which was not quite right. However I did update the name slightly to avoid abbreviations which might sometimes be confusing.

## Web service API changes
N/A

## Tests
Tests updated


agent/charms/testflinger-agent-host-charm/metadata.yaml
omar-selo, line 1: Shouldn't `ssh_private_key` and `ssh_public_key` be added to either `metadata.yaml` or `config.yaml` as charm configuration variables similar to how we have TF server [charm config](https://github.com/canonical/testflinger/blob/main/server/charm/config.yaml)

plars: Thanks, yes! I forgot to git add the config.yaml. It's there now :)

omar-selo (approved):
Looks good!

PR #329, opened 2024-08-15
Shared agent code location

plars:
Like this previous PR, this one is being proposed against the feature branch for `unified-agent-host-charm`. Because it's against a feature branch, our rules don't strictly require reviews. However I would appreciate reviews against these smaller branches to make it simpler than reviewing one large PR at the end.

## Description

We currently use the agent-charm to create a new checkout of the testflinger agent/device-connector code and install it separately for each agent in a path for each agent. This has always been useful for debugging on a real environment, but going forward, it would be much nicer to have a real dev/staging environment with real agents connected.  The down side of doing it the way we’ve been doing it, is that it adds a lot of additional time required for updating ALL the agents to the latest version of testflinger code, because we need to pull the repo for each one of them. Having it in a shared location would speed up installation and updates considerably.

## Resolved issues
CERTTF-376

## Documentation

Updated README

## Web service API changes
N/A

## Tests

Integration tests added for building the charm, deploying it, and running the action. NOTE: juju must be enabled on the system where you run these tests.


agent/charms/testflinger-agent-host-charm/tests/integration/test_charm_integration.py
```diff
@@ -0,0 +1,29 @@
...
+import pytest
+from pytest_operator.plugin import OpsTest
+
+# Root of the charm we need to build is two dirs up
+CHARM_PATH = Path(__file__).parent.parent.parent
+APP_NAME = "testflinger-agent-host"
+
+
+#@pytest.mark.skip_if_deployed
+#@pytest.mark.setup
```
omar-selo, line 11: is there a need for these commented out decorators?

agent/charms/testflinger-agent-host-charm/tests/integration/test_charm_integration.py
```diff
@@ -0,0 +1,29 @@
...
+    app = await ops_test.model.deploy(charm)
+
+    await ops_test.model.wait_for_idle(status="active",  timeout=600)
+
+
+#@pytest.mark.skip_if_deployed
+#@pytest.mark.abort_on_fail
+async def test_action_update_testflinger(ops_test: OpsTest):
+    #charm = await ops_test.build_charm(CHARM_PATH)
+    #app = await ops_test.model.deploy(charm)
```
omar-selo, line 24: same here with some declarations too

omar-selo (approved):
Looks good, just some commented out code

agent/charms/testflinger-agent-host-charm/tests/integration/test_charm_integration.py
plars: fixed, thanks!

agent/charms/testflinger-agent-host-charm/tests/integration/test_charm_integration.py
plars: Also fixed, thanks. It turns out the environment only needs to be deployed once. Which is nice, because it can take a long time in the integration tests.

PR #332, opened 2024-08-16
refactor(agent): unify treatment of output and other events

boukeas:
## Description

Even though there is currently only a single event (output) that can be subscribed to on `CommandRunner`, there are two different methods for subscribing:
- a generic `subscribe_event` method for registering handlers for events
- a specific `register_output_handler` method for registering handlers for output events

Likewise, there are two different methods for notifying the handlers that an event has taken place:
- a generic `post_event` method for events
- a specific `post_output` method for output events

This PR unifies the treatment of output events and general events, i.e. output events are treated no differently than any other event. This makes the `CommandRunner` interface slimmer and more consistent but also allows for extensibility when additional event types might be added. Event handlers are still callables but now expect the event as their argument (carrying all necessary information to handle the event, e.g. the output, within the passed argument).

## Resolved issues

This PR is not linked to a specific issue.

## Documentation

No changes required to the documentation.

## Web service API changes

No changes to the Web service API

## Tests

All existing tests are still passing (following a couple of minor modifications to reflect the PR changes). No additional functionality has been added and hence no additional tests have been introduced.

### DIFF
--- a/agent/testflinger_agent/runner.py
+++ b/agent/testflinger_agent/runner.py
@@ -22,60 +22,72 @@ import threading
 import time
 
 from collections import defaultdict
-from enum import Enum
-from typing import Callable, List, Optional, Tuple
+from typing import Callable, List, Optional, Tuple, Type
 
 from testflinger_common.enums import TestEvent
 
 logger = logging.getLogger(__name__)
 
-OutputHandlerType = Callable[[str], None]
 StopConditionType = Callable[[], Optional[str]]
 
 
-class RunnerEvents(Enum):
+class RunnerEvent:
     """
     Runner events that can be subscribed to.
     """
 
-    OUTPUT_RECEIVED = "output_received"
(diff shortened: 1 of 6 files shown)


agent/testflinger_agent/runner.py
```diff
@@ -22,60 +22,72 @@
...
+    There are also events that can be subscribed to for notifications.
     """
 
     def __init__(self, cwd: Optional[str], env: Optional[dict]):
-        self.output_handlers: List[OutputHandlerType] = []
         self.stop_condition_checkers: List[StopConditionType] = []
         self.process: Optional[subprocess.Popen] = None
         self.cwd = cwd
         self.env = os.environ.copy()
+        # a mapping of event names to lists of registered event handlers
```
nadzyah, line 74:
```suggestion
        # A mapping of event names to lists of registered event handlers
```

PR #333, opened 2024-08-16
Fix provision_log reporting

plars:
## Description

Not sure where exactly, but one of the recent changes seems to have broken provision log reporting. So it was working for a while, but after deploying updated agents in the last few days, there are some cases where it doesn't manage to report the provision_logs correctly. (this is not the TO events, this is the reporting of provision outcome for calculating pass/fail streak for provisioning on the testflinger UI).

This fixes it.

## Resolved issues

See above

## Documentation

N/A

## Web service API changes
N/A

## Tests

Added unit tests to ensure it's called properly for both cases of successful and unsuccessful provision attempts.


jocave (approved):
LGTM

PR #335, opened 2024-08-19
Create new endpoint in Testflinger Server to authenticate clients and distribute tokens for submitting jobs with priority

val500:
## Description

In order to support authorized clients sending jobs to Testflinger with priority, we need some way to identify clients and their permissions. Every authorized client will have client_id and client-key(stored as a hash in the database) with a list of permissions.
This PR adds a new endpoint which takes in a client_id and client-key and returns an encoded list of their permissions in the form of a JWT:
```
{
    "exp": <datetime> # time that the token expires
    "iat": <datetime> # time that the token was generated
    "sub": <str> # description of token("access_token")
    "max_priority": {
        "queue_name": <int> # queue name to priority map
        ...
    }
}
```
The JWT is then signed with a secret key, provided by the environment variable JWT_SIGNING_KEY, which is set in the charm config. The client can then use this token to submit privileged jobs.

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/CERTTF-370 

## Documentation

Documentation for the new endpoint was added in the server README and documentation regarding the new environment variable was added to the server configuration docs.

## Web service API changes

GET /v1/authenticate/token/<client_id>
- Authorization
    - client_id must be provided in the url
    - client-key must be provided in a header with title "client-key"
    - Invalid client_id and client-key tested in test_v1.py
- Returns status code 200 with correct credentials
- Returns status code 401 with invalid credentials

## Tests

Unit Tests were added to test_v1.py. These tests test JWT generation and the new endpoint, testing both valid and invalid credentials. A new test fixture was added to conftest.py to set up the mongo_app with permissions pre-populated in the database. In support of this, the "mongo_app" fixture was renamed to "mongo_app_fixture" to fix a linting issue where the "mongo_app" name was shadowed.

### DIFF
--- a/docs/reference/testflinger-server-conf.rst
+++ b/docs/reference/testflinger-server-conf.rst
@@ -23,6 +23,8 @@ The configuration values of Testflinger servers are read from environment variab
      - MongoDB port to connect to (Default: 27017)
    * - ``MONGODB_URI``
      - URI for connecting to MongoDB (used instead of the above config options). For example: ``mongodb://user:pass@host:27017/dbname``
+   * - ``JWT_SIGNING_KEY``
+     - Secret key used for signing tokens with permissions for authenticated clients
 
 
 Example configuration
@@ -38,3 +40,4 @@ Example configuration
   MONGODB_DATABASE="testflinger_db"
   MONGODB_HOST="mongo"
   MONGODB_URI="mongodb://mongo:27017/testflinger_db"
+  JWT_SIGNING_KEY="my_secret_key"
--- a/server/README.rst
+++ b/server/README.rst
@@ -399,15 +399,12 @@ The job_status_webhook parameter is required for this endpoint. Other parameters
 
     $ curl http://localhost:8000/v1/queues/wait_times?queue=foo\&queue=bar
 
-**[GET] /v1/authenticate/token/<client_id>** - Authenticate client key and return JWT with permissions
-
-- Parameters:
-
-  - client_id (string): Client identifier
+**[POST] /v1/oauth2/token** - Authenticate client key and return JWT with permissions
 
 - Headers:
 
-  - client-key (string): unique secret key for client
+  - Basic Authorization: client_id:client_key (Base64 Encoded)
+    
 
 - Status Codes:
 
@@ -422,5 +419,5 @@ The job_status_webhook parameter is required for this endpoint. Other parameters
 
   .. code-block:: console
 
-    $ curl http://localhost:8000/v1/authenticate/token/101 \
-           -X GET --header "client-key: ABCDEF12345"
+    $ curl http://localhost:8000/v1/oauth2/token \
+           -X GET --header "Authorization: Basic ABCDEF12345"
--- a/server/README.rst
+++ b/server/README.rst
@@ -398,3 +398,29 @@ The job_status_webhook parameter is required for this endpoint. Other parameters
   .. code-block:: console
 
     $ curl http://localhost:8000/v1/queues/wait_times?queue=foo\&queue=bar
+
+**[GET] /v1/authenticate/token/<client_id>** - Authenticate client key and return JWT with permissions
+
+- Parameters:
+
+  - client_id (string): Client identifier
+
+- Headers:
+
+  - client-key (string): unique secret key for client
+
+- Status Codes:
+
+  - HTTP 200 (OK)
+  - HTTP 401 (Unauthorized) - Invalid client_id or client-key
+
+- Returns:
+
+  Signed JWT with permissions for client
+
+- Example:
+
+  .. code-block:: console
+
+    $ curl http://localhost:8000/v1/authenticate/token/101 \
+           -X GET --header "client-key: ABCDEF12345"
--- a/server/src/api/v1.py
+++ b/server/src/api/v1.py
@@ -33,7 +33,7 @@ from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
 
 import jwt
-
+import bcrypt
 
 from src import database
 from . import schemas
@@ -675,15 +675,19 @@ def validate_client_key_pair(client_id: str, client_key: str):
     """
     Checks client_id and key pair for validity and returns their permissions
     """
+    if client_key is None:
+        return None
+    client_key_bytes = client_key.encode("utf-8")
     client_permissions_entry = database.mongo.db.client_permissions.find_one(
         {
             "client_id": client_id,
-            "client_secret_hash": client_key,
         }
     )
-    if client_permissions_entry is None:
-        return {}
 
+    if client_permissions_entry is None or not bcrypt.checkpw(
+        client_key_bytes, client_permissions_entry["client_secret_hash"]
+    ):
+        return None
     permissions = client_permissions_entry["permissions"]
     return permissions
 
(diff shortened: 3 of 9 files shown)


server/src/api/v1.py
```diff
@@ -652,3 +656,45 @@ def queue_wait_time_percentiles_get():
...
+        return {}
+
+    permissions = client_permissions_entry["permissions"]
+    return permissions
+
+
+SECRET_KEY = secrets.token_hex(20)
+
+
+@v1.get("/authenticate/token/<client_id>")
```
boukeas, line 694: If there's a 1:1 correspondence between the `client_id` and the `client_key` then you don't really need both; the key in the header should be enough to retrieve the permissions entry.

boukeas:
General naming question: the use of the word "client" suggests (to me at least) that we are authenticating _machines_ rather than _users_ (I am using the term "user" here broadly, it could refer to teams as well). It's not too important and I might be wrong, I am just querying what the intention is, i.e. what we are trying to convey with the term.

server/src/api/v1.py
```diff
@@ -652,3 +656,45 @@ def queue_wait_time_percentiles_get():
...
+    }
+    token = jwt.encode(token_payload, secret_key, algorithm="HS256")
+    return token
+
+
+def validate_client_key_pair(client_id: str, client_key: str):
+    """
+    Checks client_id and key pair for validity and returns their permissions
+    """
+    client_permissions_entry = database.mongo.db.client_permissions.find_one(
```
boukeas, line 678: I hope this is not a silly question but why do the permissions need to be retrieved here and encoded in the token? I feel like the token could simply contain the `client_id`. When we actually _submit_ a job with the token in the header, the `client_id` in the token should be sufficient to retrieve the permissions from the database (just as we do here).

Also, since this is an authentication endpoint, isn't it possible that we might be using it for other purposes other than submitting a job? What I am saying is that it feels to me a bit incongruous to include permissions in response to an authentication request.

server/src/api/v1.py
boukeas, line 694: What is the idea behind the use of "token" in `authenticate/token`? I might be missing something but if we are authenticating a client or a user, this could be `/authenticate/client` or `authenticate/user`. Which would also be compatible with adding something like `/authenticate/agent` later.

plars: In this case, I think having both the id and a key/token/secret (the private part of the credentials) is useful so that you don't have to search through all the known IDs to see if it matches one of them. I would imagine there could be another use here where it's not 1:1 at all, but rather a user being able to generate multiple API keys, giving them names, expirations, etc. In that case, you would definitely want some sort of ID to compare against.

server/src/api/v1.py
```diff
@@ -652,3 +656,45 @@ def queue_wait_time_percentiles_get():
...
+        }
+    )
+    if client_permissions_entry is None:
+        return {}
+
+    permissions = client_permissions_entry["permissions"]
+    return permissions
+
+
+SECRET_KEY = secrets.token_hex(20)
```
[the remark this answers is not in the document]
boukeas: https://docs.python.org/3/library/secrets.html

server/src/api/v1.py
val500: I was just copying the oauth standard endpoint of `/oauth/token` - https://auth0.com/docs/authenticate/protocols/oauth

boukeas:
> General naming question: the use of the word "client" suggests (to me at least) that we are authenticating _machines_ rather than _users_ (I am using the term "user" here broadly, it could refer to teams as well). It's not too important and I might be wrong, I am just querying what the intention is, i.e. what we are trying to convey with the term.

Alright, after reading around a little bit I retract that comment. We could be authenticating users, machines or applications and the general term used for all these is indeed "client". And, in our case, we could really be authenticating users (submitting jobs) or agents (requesting jobs) so a general term is what we need.

server/src/api/v1.py
```diff
@@ -652,3 +656,45 @@ def queue_wait_time_percentiles_get():
...
+
+
+@v1.get("/authenticate/token/<client_id>")
+def authenticate_client_get(client_id: str):
+    """Get JWT with priority and queue permissions"""
+    client_key = request.headers.get("client-key")
+    allowed_resources = validate_client_key_pair(client_id, client_key)
```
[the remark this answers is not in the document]
val500: client_key is hashed in new patch set

server/README.rst
```diff
@@ -398,3 +398,29 @@ The job_status_webhook parameter is required for this endpoint. Other parameters
   .. code-block:: console
 
     $ curl http://localhost:8000/v1/queues/wait_times?queue=foo\&queue=bar
+
+**[GET] /v1/authenticate/token/<client_id>** - Authenticate client key and return JWT with permissions
```
plars, line 402: Let's change this slightly. 
- Instead of `/v1/authenticate/token/...`, let's just use `/v1/oauth2/token` to be a bit more consistent with things like hexr (we may also eventually want to add data like grate-type and scope, but it's not clear how we would use that just yet in our current spec).
- I think this would usually be a POST request, not GET
- Instead of specifying the client_id here, let's use something more standard like basic auth. We should *ALWAYS* use this with https, but instead of inventing our own header format here, we can use [RFC3986](https://datatracker.ietf.org/doc/html/rfc3986) compliant basic auth headers which are base64 encoded. So that means the header you should be looking for in the API looks like this:
```
Authorization: Basic BASE_64_ENCODE_OF_USERNAME:PASSWORD
```
When you b64decode that, then you get the original data back as username:password. Along with https of course, this is generally regarded as much better than transmitting it without some sort of encoding. Fun fact, the other thing specified in that RFC, which is adopted by pretty much every client today I think, means that if you do `curl https://myuser:mypassword@testflinger.canonical.com/v1/login` then it will automatically b64encode it for you on the client side! :)

plars (changes requested):
Looking good so far, I'll keep testing but I spotted a few things I think we should go ahead and change.

server/src/api/v1.py
```diff
@@ -652,3 +656,60 @@ def queue_wait_time_percentiles_get():
...
+        return None
+    permissions = client_permissions_entry["permissions"]
+    return permissions
+
+
+SECRET_KEY = os.environ.get("JWT_SIGNING_KEY")
+
+
+@v1.post("/oauth2/token")
+def authenticate_client_post():
```
plars, line 698: Can we rename this to `retrieve_token()`? I think that's a more descriptive name for what we're doing.

docs/reference/testflinger-server-conf.rst
```diff
@@ -37,4 +39,5 @@ Example configuration
   MONGODB_PASSWORD="testflinger"
   MONGODB_DATABASE="testflinger_db"
   MONGODB_HOST="mongo"
-  MONGODB_URI="mongodb://mongo:27017/testflinger_db"
\ No newline at end of file
+  MONGODB_URI="mongodb://mongo:27017/testflinger_db"
+  JWT_SIGNING_KEY="my_secret_key"
```
plars, line 43: Let's also add this to the example override file in `devel/docker-compose-override.yml`

server/README.rst
plars: Don't forget to update the README here!

plars (changes requested):
Thanks for fixing it - github seemed to be choking on the need for a merge. I was able to successfully run it and get a jwt back. I just have a few more minor things here which should be real quick I think, and one thing you missed from the previous review. I'm looking forward to testing it out with the bits that make use of it!

plars (approved):
+1 thanks!

PR #337, opened 2024-08-21
Update configs action

plars:
## Description

Moving the code for deploying configs to the agent-host-charm from the agent-charm means that we can update all the configs at once. This should be a lot faster and easier to manage. The charm needs a place to pull them from. Right now, we store them in git, so it makes sense to add a config option for the git repo and path for the part of that repo that contains the configs this agent host should look at. NOTE: this does NOT tie us to using git forever. If a better option emerges, then we could support that in addition to, or instead of this.

The configs are not static. Agents will be added, removed, and changed over time. There should be a juju action that allows us to trigger an update to the agent configs.

## Resolved issues

CERTTF-377

## Documentation

Updated README

## Web service API changes

N/A

## Tests

Updated unit and integration tests

NOTE: For the integration tests, I added some data to THIS branch, and used config settings we provide here to force it to get that test data from this branch. Eventually, when this lands, we'll want to update this to point to main instead, but that's not something we can change until this lands.

### DIFF
--- a/agent/charms/testflinger-agent-host-charm/src/charm.py
+++ b/agent/charms/testflinger-agent-host-charm/src/charm.py
@@ -6,10 +6,13 @@
 
 
 import logging
-import shutil
 import os
-from pathlib import PosixPath
+import shutil
+import sys
+from pathlib import Path
 from common import run_with_logged_errors
+from git import Repo, GitCommandError
+from jinja2 import Template
 
 from charms.operator_libs_linux.v0 import apt
 from ops.charm import CharmBase
@@ -22,7 +25,7 @@ from ops.model import (
     ModelError,
 )
 import testflinger_source
-from defaults import LOCAL_TESTFLINGER_PATH
+from defaults import AGENT_CONFIGS_PATH, LOCAL_TESTFLINGER_PATH
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


plars:
Oops, just realized I forgot to set the base branch. This one should also go against the feature branch for now, not main

agent/charms/testflinger-agent-host-charm/src/charm.py
```diff
@@ -79,6 +93,39 @@ def update_testflinger_repo(self):
         self.unit.status = MaintenanceStatus("Cloning testflinger repo")
         testflinger_source.clone_repo(LOCAL_TESTFLINGER_PATH)
 
+    def update_config_files(self):
+        """
+        Clone the config files from the repo and swap it in for whatever is
+        in AGENT_CONFIGS_PATH
+        """
+        config_repo = self.config.get("config-repo")
+        config_dir = self.config.get("config-dir")
```
nancyc12, line 102: `config_dir` is not used in Repo.clone_from. Is it missed?

nancyc12 (changes requested):
This is great to put all config updates in one single action!
I just have one small question.

agent/charms/testflinger-agent-host-charm/src/charm.py
plars: Correct. That's something we're going to use later, but it was helpful to go ahead and include the config option for it here. I'm happy to consider a different name if you have suggestions for something you think might be more clear. The purpose of `config-dir` is the location inside the repo where you will find the configs for agents this host needs to care about. For example, we might point it at a repo that has configs for multiple labs, but each needs a separate agent-host deployment. So with `config-dir`, you can tell it to only look at configs in the path `ce-cert/data-tel-l3` for example.

An upcoming change will then take all those configs, and generate service files to run the agent for each one of them. But this way it won't look at the configs in other parts of the repo that might be out-of-scope for this lab.

nancyc12: Thanks for the clarification. If it is something that we will utilize in upcoming changes then it looks good to keep it as now. The name is very clear for me as my understanding was the same as your definition.  :)

nancyc12 (approved):

PR #338, opened 2024-08-23
device-connectors: Fix handling attachments with absolute path in oem-autoinstall

Artur-at-work:
## Description
`oem_autoinstall` fails when user gives absolute path in job.yaml (i.e `/home/artur/files/jenkins-token.txt`), but if relative path (i.e `files/jenkins-token.txt`) is given then it works.
Missed this bug at first, because was testing on local machine and when absolute path was overwriting ATTACHMENTS_DIR it still pointed to the right file. But when deployed in charm it failed to find the attachments correctly.

## Resolved issues
Closes https://warthogs.atlassian.net/browse/OEX86-416

## Documentation
n/a

## Web service API changes
n/a

## Tests
Run agent charm and test provisioning on HP machine.

Before fix:
https://testflinger.canonical.com/jobs/67001261-d590-41a1-bb3d-43a3e5db153b

After fix:
https://testflinger.canonical.com/jobs/f86f6684-45be-407b-8db6-ca724c828606

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/data/muxpi/oem_autoinstall/provision-image.sh
+++ b/device-connectors/src/testflinger_device_connectors/data/muxpi/oem_autoinstall/provision-image.sh
@@ -268,7 +268,7 @@ do
     sleep 180
     currentTime=$(date +%s)
     if [[ $((currentTime - startTime)) -gt $TIMEOUT ]]; then
-        echo "Timeout is reached, deployment are not finished"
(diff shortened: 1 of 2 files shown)


Artur-at-work:
@plars could you help to review please?
or should I ask oem team to review when smaller changes?

plars (approved):
I'm happy to review it! And yes I think this would work, although you could also say that the config is wrong if they specify the path including /. Handling it gracefully like this is probably a better option than just rejecting it though. So +1

PR #342, opened 2024-08-26
chore(muxpi): Move storage plugging to context manager

thp-canonical:
## Description

This moves the code for temporarily mounting the storage device on the host running the device connector to a single function, centralizing the `stm` and `zapper sdwire` / `zapper typecmux` commands for easy discovery. Ideally in the future, this feature should be provided to the test code in a generic way, so that test code can work without having to know whether SD, USB or the "stm" method is used for deployment.

## Resolved issues

This is a first step to making it possible for test scripts to mount the SD card / USB storage device back to the agent for inspection in a test case.

## Documentation

This should keep the behavior the same, with the exception that due to the `finally:` blocks in the context manager, the storage device is returned to the DUT even if an exception is thrown in the code that flashes the image to the device. This is different to what happened before (exception is thrown, and `stm -dut`, `zapper sdwire set DUT`, `zapper typecmux set DUT` were never called). It might be desireable that we don't re-attach the storage back to the DUT in these cases. In this case, just the `try..finally` would need to be removed.

## Web service API changes

None.

## Tests

Not tested.

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/muxpi/muxpi.py
@@ -161,6 +161,58 @@ class MuxPi:
         # One final check to ensure the control host is alive, or fail
         self._run_control("true")
 
+    @contextmanager
+    def _storage_plug_to_self(self):
+        """Temporarily connect storage to provision an image.
+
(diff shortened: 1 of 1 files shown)


plars (approved):
Looks good, and I also ran some provisioning on some devices with this branch applied and they worked fine. +1

PR #346, opened 2024-08-26
Trigger an agent restart if USR1 signal is received

plars:
So I'm proposing this one against main rather than the feature branch.

## Description

We already have a “safe restart” mechanism for the testflinger agent.  This works by waiting until the agent is not running a job, then checking for a marker file that signals the agent to restart.  This way, individual agents can be safely updated, then told to restart whenever they are done with their current jobs, so that we avoid a hard-restart that interrupts a job in progress.

Since supervisord gives us a nice mechanism for sending signals to all processes, we should take advantage of that by changing testflinger-agent to also respond to a signal as a trigger that it needs to perform a safe-restart. It’s common to use HUP to signal that something should reload it’s configuration files, but I’m not sure that really fits here. That’s normally just a “reread” of the configs, but in this case we’re telling it to completely exit so that it not only rereads the configs but also reloads the new version of the code that was potentially installed. So I would suggest using USR1 for the signal in this case.

## Resolved issues

CERTTF-379

## Documentation

Added documentation in the explanation section about this, and other methods for safely shutting down or restarting an agent

## Web service API changes

N/A

## Tests

Tested locally and also added a new test that executes the real agent, triggers a reset with the signal, and ensure it takes all the right steps.

### DIFF
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -344,6 +344,6 @@ class TestflingerAgent:
         If we receive the restart signal, tell the agent to restart safely when
         it is not running a job
         """
-        logger.info("Triggering agent restart")
+        logger.info("Marked agent for restart")
         restart_file = self.get_restart_files()[0]
         open(restart_file, "w").close()
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2017-2020 Canonical
+# Copyright (C) 2017 Canonical
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -17,6 +17,8 @@ import logging
 import os
 from pathlib import Path
 import shutil
+import signal
+import sys
 import tempfile
 
 from testflinger_agent.job import TestflingerJob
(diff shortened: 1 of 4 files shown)


agent/testflinger_agent/agent.py
```diff
@@ -335,3 +338,12 @@ def retry_old_results(self):
             except TFServerError:
                 # Problems still, better luck next time?
                 pass
+
+    def restart_signal_handler(self, _, __):
+        """
+        If we receive the restart signal, tell the agent to restart safely when
+        it is not running a job
+        """
+        logger.info("Triggering agent restart")
```
omar-selo, line 347: If I understand this correctly, the agent isn't restarted immediately. It's restarted whenever `check_restart` gets called after this restart file was created. If so would it make sense to log `Marked agent for restart` instead?

omar-selo (approved):
Looks good

agent/testflinger_agent/agent.py
plars: That's why I said "Triggering" instead of just "restarting agent" but if that's more clear to call it "Marked" then I'm happy to do that. Changing it now, will push an update in just a minute.

omar-selo (approved):
Looks good, thanks for the change!

PR #347, opened 2024-08-28
Mount value needs to be converted to a string from Path object

plars:
## Description

Quick fix for a problem Kevin reported:
```
packages/testflinger_device_connectors/devices/muxpi/muxpi.py", line 383, in remote_mount
    self._run_control("sudo umount {}".format(shlex.quote(mount)))
  File "/usr/lib/python3.10/shlex.py", line 329, in quote
    if _find_unsafe(s) is None:
TypeError: expected string or bytes-like object
```

## Resolved issues

See above

## Documentation

N/A

## Web service API changes

N/A

## Tests

Tested deployment with this to ensure nothing else was needed.


kevinyehk (approved):
LGTM +1, thanks for the quick fix.

PR #348, opened 2024-08-28
fix(oem_autoinstall): Fix download ISO from webdav URL with password

Artur-at-work:
## Description
- When DUT needs to download ISO from the oem-share webdav behind OpenID, then wget fails.
So we have to use rclone in this case.
 Make provision-image.sh script to install rclone and scp rclone.conf to DUT, then run `rclone copy` on DUT to download the ISO

- When download from tel cache server need to add the tls cert. So provision-image.sh will copy the agent's cert to DUT

## Resolved issues
Closes https://warthogs.atlassian.net/browse/OEX86-425

## Documentation
- Update oem_autoinstall docs to specify the rclone config file format

## Web service API changes
n/a

## Tests
- Run charm agent and provision the laptop
- Use webdav [url](https://oem-share.canonical.com/partners/somerville/share/releases/noble/oem-24.04a/20240822-73/somerville-noble-oem-24.04a-20240822-73.iso) for ISO
Job pass:
https://testflinger.canonical.com/jobs/43a8a102-8e5a-4db5-8911-f0d858b0b1d0

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/data/muxpi/oem_autoinstall/provision-image.sh
+++ b/device-connectors/src/testflinger_device_connectors/data/muxpi/oem_autoinstall/provision-image.sh
@@ -96,10 +96,10 @@ wget_iso_on_dut() {
     URL_TOKEN="$CONFIG_REPO_PATH"/url_token
 
     echo "Downloading ISO on DUT..."
-    if [[ "$URL_DUT" =~ "oem-share" ]]; then
-        # use rclone for oem-share webdav storage, coz wget fails OpenID
+    if [[ "$URL_DUT" =~ "oem-share.canonical.com" ]]; then
+        # use rclone for webdav storage
         if [ ! -f "$URL_TOKEN" ]; then
-            echo "oem-share URL requires webdav authentication. Please provide url_token file"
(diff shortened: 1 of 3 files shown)


plars (approved):
I think this is fine if it's needed, but looks like it's only needed for oem-share. So it might be worth considering using the proxy we set up in the lab so that things like this are not needed, if that's the only server where you need webdav. I'm also kinda surprised that http basicauth doesn't work in this case. But if this works for you, it doesn't seem to affect other cases, so +1

I'll leave it unmerged for now, because I think you can also merge it and I wanted you to consider the comment above. But if you need me to land it, let me know. Also, if you do want to land it, then you'll need to sign the commits first. Thanks!

Artur-at-work:
Yes, it's only needed for oem-share. Proxy could work for internal teams, but we use that oem-share URL to auto-release dev images to external customers and QA need to verify via those urls. So I think it may not work for them. Later, if QA can confirm proxy works for everyone, I'm happy to retire this dependency on rclone.

medicalwei (approved):
There is no change from @plars 's review, the last 2 force-pushes was an attempt to sign commit the first was with a one-line change to suppress wget logs, the second force push reverted it.

PR #351, opened 2024-08-29
Write out the supervisord configs for all agents

plars:
## Description

To use supervisord for managing the agents, there will need to be a service file created for each agent.  This is pretty simple to generate and looks basically like this:
[program:<agent name>]
command=<command to start the agent>

Whenever the location of the configs is changed, or the action to update the configs is triggered, the charm should remove the existing configs under /etc/supervisord/conf.d, and then generate a new set of configs for all known agents.

The reason for this, is so that if any agents were removed, we ensure that the old agent is gone. This will be used for ensuring that we no longer run the service for a deleted agent.

Also, this branch now includes the parts that `update` the supervisord services when we rewrite them - this will launch any new agents that show up, and shut down any old ones that are removed. This will ONLY happen if there's an actual change to that supervisor conf file. i.e. it will NOT re-execute (update) the running service just because the identical file has been rewritten. This is important, because if it did, then it would shut it down immediately even if it was in the middle of a job.

So that means that we also have to handle the case of agents that have had possible updates to the config files for testflinger (not supervisor). Since there's a chance that could have happened to any of them, the safest course of action is to mark them all as needing a restart. This will use the `signal` mechanism in supervisor to tell all the agents to safely restart as soon as they are not running a test job.

## Resolved issues

CERTTF-378
CERTTF-380

## Documentation

This also added some features to make it easier to run and debug integration tests. Updated documentation for the agent host charm to include instructions for running the charm tests and debugging them.

## Web service API changes

N/A

## Tests

Added integration tests for this and fixed affected unit tests


agent/charms/testflinger-agent-host-charm/src/charm.py
```diff
@@ -126,6 +131,54 @@ def update_config_files(self):
...
+            logger.error("No agent directories found in config-dirs")
+            self.unit.status = BlockedStatus(
+                "No agent directories found in config-dirs"
+            )
+            sys.exit(1)
+
+        # Remove all the old service files in case agents have been removed
+        for conf_file in os.listdir("/etc/supervisor/conf.d"):
+            if conf_file.endswith(".conf"):
+                os.unlink(f"/etc/supervisor/conf.d/{conf_file}")
```
jocave, line 163: Clearing all the service files and re-writing seems reliable, but I'm wondering if you considered performing a comparison with the new intended configuration and only deleting or updating where necessary? I would see the small potential benefit being that someone inspecting the agent host could see from file creation times whether an agent is new or updated.

plars: @jocave This is just for the supervisor service files, which all follow the same template. What changes are you anticipating here that we need to look at the timestamp on them? There will be new ones that show up if we add agents, but you can also see when they were added to the repo.

I suppose we could do something like this, but I'd like to better understand what we're trying to achieve. I saw this as a much simpler approach which also ensures that we remove any agents that are no longer valid.

jocave: Just trying to think of ideas that might help in managing an agent host that has been running for some period of time. I can't really say what problems might arise, but just thought there could be a situation in which it would be helpful to know which agents are new after an update and which have had instance that (according to the config produced by the charm) should have been running for longer.

plars:
Updated to include CERTTF-380 - which is closely related to what was already in this branch anyway.

agent/charms/testflinger-agent-host-charm/src/charm.py
plars: Not sure if this helps, but you can see how long an agent has been running:
```
$ supervisorctl status
agent-001                        RUNNING   pid 7218, uptime 3 days, 18:05:33
agent-002                        RUNNING   pid 7219, uptime 3 days, 18:05:33
agent-003                        RUNNING   pid 7220, uptime 3 days, 18:05:33
agent-004                        RUNNING   pid 7221, uptime 3 days, 18:05:33
agent-005                        RUNNING   pid 7222, uptime 3 days, 18:05:33
agent-006                        RUNNING   pid 7223, uptime 3 days, 18:05:33
agent-007                        RUNNING   pid 7224, uptime 3 days, 18:05:32
agent-008                        RUNNING   pid 7225, uptime 3 days, 18:05:32
agent-009                        RUNNING   pid 7226, uptime 3 days, 18:05:32
...
```

But there are also certain situations, such as when we update the testflinger code, where we will also trigger ALL the agents to restart when they are not running a job. We also do this for all agents when the configs have been updated. Otherwise, we would have to keep track of any possible change to any config  within the testflinger and device connector configs across all agents. I think it's less error prone to let them all safely restart and ensure they are running the latest config though.

plars:
A few more updates, like renaming the config options to keep them consistent with uses of - vs _. Also updated the README for the charm with additional details and operational tips.

agent/charms/testflinger-agent-host-charm/src/charm.py
plars: resolving so I can land this, per our mm conversaton

PR #354, opened 2024-09-03
Add agent_name in TF environment

rmartin013:
## Description
It is currently impossible to guess the name of the device agent from a testflinger job running from a device agent queue. This is unfortunate since we have a few Riverside tools that abstract the complexity of a setup, given the device name as an argument. Adding it as an environment variable would reduce the development effort on those tools and can probably benefit as well to other use cases.

We feel the need for this change in the context of https://warthogs.atlassian.net/browse/PERI-572.

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/__init__.py
@@ -195,6 +195,12 @@ class DefaultDevice:
         serial_port = config.get("serial_port")
         serial_proc = SerialLogger(serial_host, serial_port, "test-serial.log")
         serial_proc.start()
+
+        extra_env = {}
+        extra_env["agent_name"] = self.config.get("agent_name")
+        if "env" not in self.config:
+            self.config["env"] = {}
+        self.config["env"].update(extra_env)
         try:
             exitcode = testflinger_device_connectors.run_test_cmds(
                 test_cmds, config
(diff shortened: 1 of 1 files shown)


rmartin013:
@plars I would like your review on this PR. I tried to test it using "snappy-device-agent muxpi provision -c default-agx-c30841.yaml test.json", but I am constantly running into timeout issue, and I feel there is probably a better way to test it. Since in that case, I don't need to provision a device to test it, but simply run a "env" command in the test phase.

We already discussed about this need, and I was feeling like it would probably be easier for you if I propose a patch

device-connectors/src/testflinger_device_connectors/devices/__init__.py
```diff
@@ -195,6 +195,12 @@ def runtest(self, args):
         serial_port = config.get("serial_port")
         serial_proc = SerialLogger(serial_host, serial_port, "test-serial.log")
         serial_proc.start()
+
+        extra_env = {}
+        extra_env["agent_name"] = self.config.get("agent_name")
```
plars, line 200:
```suggestion
        extra_env["AGENT_NAME"] = self.config.get("agent_name", "")
```

Just two things I would recommend here (proposed above):
1. make the env var all caps
2. make sure to get "" as a default value in case (unlikely) agent_name is missing, that way we avoid TypeError

plars (changes requested):
This looks ok I think, just a couple of small recommendations

rmartin013:
@plars thanks for your review, I fixed that and a syntax error for getting the configuration. I finally managed to find the test command allowing me to check my change, so I am now pretty confident with it:
`snappy-device-agent muxpi runtest -c default-nano.yaml test.json`

plars (approved):
+1

plars:
/canonical/self-hosted-runners/run-workflows 3ae5b1ff8c1669dca25bb5be9a495637b75de9c2

PR #357, opened 2024-09-10
feat: add custom timeout option to zapper provisioning stage

andersson1234:
The reason I've done this, is that the default timeout of 5400 seconds is too short for my use case.

I'm downloading desktop ISO's, from a cdimage mirror, which needs to cache the image on the first download request.

Because of this, it can take longer than 90 mins to download the target image, especially if the first download attempt fails (it can attempt up to 5 times I believe).

This change fixes the issue by allowing a custom timeout for the provisioning stage when using the zapper kvm device connector.

## Description

Adds the option to have a custom timeout for zapper provisioning staage.

## Resolved issues

No associated Jira cards.

## Documentation

There is new documentation under docs/device-connector-types.rst

## Web service API changes

None

## Tests

I just tested my changes with Tox. If there's further testing needed please let me know.


andersson1234:
Amended based on comment

PR #359, opened 2024-09-12
Reset storage to flat layout if no default is specified

plars:
## Description

The maas device connector supports specifying a storage configuration in the job, and allows a default to be set in the device-connector config for it to reset back to if it's not specified. However, if this isn't set, there's no way for it to know how to set it back to something "safe". This forces it back into a default "flat" layout which should match what it had when the system was originally enrolled in maas.

## Resolved issues

N/A

## Documentation
N/A

## Web service API changes
N/A

## Tests
Tested on a maas agent by sending a job with a storage layout to see that it used the one in the job, then sending one without a storage layout in the job to see that it was reset back to a flat layout. This system did not have a default layout configured in the device connector config.


val500 (approved):
LGTM

PR #360, opened 2024-09-12
Add authentication to CLI when submitting jobs with priority

val500:
## Description

Requires #335 and #361. This updates the submit command in the Testflinger CLI to authenticate clients and authorize jobs that specify a job_priority in the yaml. For these jobs, the CLI will first get a token from the server, which requires a client_id and client-key. These can be specified as an environment variable, in the command line options, or in the config. The CLI will submit this token with the job.

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/CERTTF-372

## Documentation

Documentation was added to the CLI README and submit-job documentation in  docs/ 

## Web service API changes
N/A

## Tests
Unit tests were added to test_cli.py. These tests test that the JWT received from the authentication endpoint is passed with the job for jobs with priority, and jobs with priority are rejected when client credentials aren't specified in the CLI.

### DIFF
--- a/docs/how-to/job-priority.rst
+++ b/docs/how-to/job-priority.rst
@@ -1,0 +1,15 @@
+Submit a test job with priority
+===============================
+
+You can add the :doc:`job_priority <../explanation/job-priority>` field to your
+job YAML like this:
+
+.. code-block:: yaml
+
+        job_priority: 100
+
+This field requires an integer value with a default value of 0. The maximum
+priority you can set depends on the permissions that you have for the queue
+you are submtting to.
+
+In order to use this field, you need to be :doc:`authenticated <./authentication>` with the server.
--- a/docs/how-to/job-priority.rst
+++ b/docs/how-to/job-priority.rst
@@ -10,6 +10,6 @@ job YAML like this:
 
 This field requires an integer value with a default value of 0. The maximum
 priority you can set depends on the permissions that you have for the queue
-you are submtting to.
+you are submitting to.
 
 In order to use this field, you need to be :doc:`authenticated <./authentication>` with the server.
--- a/docs/how-to/job-priority.rst
+++ b/docs/how-to/job-priority.rst
@@ -12,4 +12,4 @@ This field requires an integer value with a default value of 0. The maximum
 priority you can set depends on the permissions that you have for the queue
 you are submitting to.
 
-In order to use this field, you need to be :doc:`authenticated <./authentication>` with the server.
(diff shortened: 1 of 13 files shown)


plars (changes requested):
Haven't looked at all of this yet, but I noticed the tests section is completely empty. Can we get more details on how you tested this? I think it's also good to have some sort of unit tests that cover the new code here too.

val500:
> Haven't looked at all of this yet, but I noticed the tests section is completely empty. Can we get more details on how you tested this? I think it's also good to have some sort of unit tests that cover the new code here too.

Just added 2 unit tests - is there any other functionality you think it would be beneficial to test?

docs/how-to/job-priority.rst
```diff
@@ -0,0 +1,15 @@
...
+
+.. code-block:: yaml
+
+        job_priority: 100
+
+This field requires an integer value with a default value of 0. The maximum
+priority you can set depends on the permissions that you have for the queue
+you are submitting to.
+
+In order to use this field, you need to be :doc:`authenticated <./authentication>` with the server.
```
plars, line 15: Small rephrase - "... you need an authorisation token from the server"

docs/reference/job-schema.rst
```diff
@@ -54,7 +54,7 @@ The following table lists the key elements that a job definition file should con
     - | (Optional) URL to send job status updates to. These updates originate from the agent and get posted to the server which then posts the update to the webhook. If no webhook is specified, these updates will not be generated.
   * - ``job_priority``
     - integer
-    - | (Optional) Integer specifying how much priority this job has. Jobs with higher job_priority will be selected by agents before other jobs. Specifying a job priority requires authorization in the form of a JWT obtained by sending a POST request to /v1/oauth2/token with a client id and client key specified in an Authorization header.
+    - | (Optional) Integer specifying how much priority this job has. Jobs with higher job_priority will be selected by agents before other jobs. Specifying a job priority requires authorisation in the form of a JWT obtained by sending a POST request to /v1/oauth2/token with a client id and client key specified in an Authorisation header.
```
plars, line 57: This is where spelling differences get tricky. To follow guidelines, we should use the UK spelling. But AFAIK, "Authorization" is the spelling used in the RFC. To be more clear, when you're actually referring to the header, it should be the spelling with the 'z'.

plars (changes requested):
Just a couple of very minor things

plars (approved):
checkmate, +1 :)

PR #361, opened 2024-09-18
Add logic to validate token and check priority permissions

val500:
## Description

Requires #335. This PR updates the job schema include a new job_priority field. In order for a job to specify a job_priority, the client needs to include a signed JWT that gives them the permission to push a job to the specified queue at their specified priority level. This PR includes validation of the attached JWT with jobs that specify priority. After token validation, job priority is added to the job structure in the database.

## Resolved issues

Resolves https://warthogs.atlassian.net/browse/CERTTF-371

## Documentation
 
Job schema docs were updated to reflect addition of job_priority field.

## Web service API changes

POST /v1/job endpoint updated to include token validation  

## Tests
Unit tests were added to test_v1.py.

### DIFF
--- a/server/src/api/v1.py
+++ b/server/src/api/v1.py
@@ -100,7 +100,7 @@ def has_attachments(data: dict) -> bool:
     )
 
 
-def check_token_permission(
+def check_token_priority_permission(
     auth_token: str, secret_key: str, priority: int, queue: str
 ) -> bool:
     """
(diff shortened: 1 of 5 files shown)


server/src/api/v1.py
```diff
@@ -101,7 +100,33 @@ def has_attachments(data: dict) -> bool:
...
-def job_builder(data):
+def check_token_permission(
+    auth_token: str, secret_key: str, priority: int, queue: str
+) -> bool:
+    """
+    Validates token received from client and checks if it can
+    push a job to the queue with the requested priority
+    """
+    if auth_token is None:
+        abort(401, "No authentication token specified")
```
plars, line 111:
```suggestion
        abort(401, "Unauthorized")
```

server/src/api/v1.py
```diff
@@ -101,7 +100,33 @@ def has_attachments(data: dict) -> bool:
...
+    else:
```
plars, line 112: Eliminate this entire `else` block. You did the right thing by checking the error condition first and aborting. But once you've done that, then we're on the happy path. No need for the `else` here :)

server/src/api/v1.py
```diff
@@ -101,7 +100,33 @@ def has_attachments(data: dict) -> bool:
...
+        try:
+            decoded_jwt = jwt.decode(
+                auth_token,
+                secret_key,
```
plars, line 116: I'm not sure this is in scope?

server/src/api/v1.py
```diff
@@ -101,7 +100,33 @@ def has_attachments(data: dict) -> bool:
...
+                algorithms="HS256",
+                options={"require": ["exp", "iat", "sub", "max_priority"]},
```
plars, line 118: what about client permissions that don't have a max_priority? We already know of at least one other use we want to add - for restricted queues. It's entirely possible that they have `allowed_queues` or some other future key, but not `max_priority`.

plars (approved):
Looks good now, thanks for the fixes!

PR #362, opened 2024-09-19
Add agent-host terraform module

plars:
## Description

This is the terraform module I've used to deploy the agent host in the new server lab. Wanted to make sure this was merged so that we have it in this branch with the rest of the new stuff for the unified agent host charms.

## Resolved issues

N/A

## Documentation

added README

## Web service API changes

None

## Tests

I've deployed this in a test environment at home, as well as in the new server lab.

### DIFF
--- a/agent/terraform/README.md
+++ b/agent/terraform/README.md
@@ -1,0 +1,70 @@
+# Deploying with Terraform + Juju
+
+This Terraform module can be used for deploying the Testflinger agent
+host on a physical or virtual machine model using Juju.
+
+## Deploying an Environment
+
+This terraform module assume that the Juju model you want to deploy
+to has already been created.
+
+1. First, create the model
+```
+    $ juju add-model agent-host-1
+```
+
+2. Create ssh keys to use on the agent host
+```
+    $ ssh-keygen -t rsa -f mykey
+```
+
+3. Create a git repo with the Testflinger configs
+
+If you have more than one agent host to deploy, create a separate directory
+for each of them in this repo. Then create a separate directory for each
+agent in the agent host directory where they reside. For example:
+
+```
+/agent-host-1
+-/agent-101
+  -testflinger-agent.yaml
+  -default.conf
+-/agent-102
+  -testflinger-agent.yaml
+  -default.conf
+/agent-host-2
+...
+```
+
+4. Create a main.tf which specifies the required parameters for this module
+
+```
+terraform {
+  required_providers {
+    juju = {
+      version = "~> 0.13.0"
+      source  = "juju/juju"
+    }
+  }
+}
+
+provider "juju" {}
+
+module "lab1" {
+  source = "/path/to/this/module"
+  agent_host_name = "agent-host-1"
+  juju_model = "agent-host-1"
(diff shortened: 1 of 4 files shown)


agent/terraform/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+```
+terraform {
+  required_providers {
+    juju = {
+      version = "~> 0.13.0"
+      source  = "juju/juju"
+    }
+  }
+}
```
omar-selo, line 49: perhaps there's no need to add this to `main.tf` given that it's already in `versions.tf`

agent/terraform/main.tf
omar-selo, line 1: if this file is named `main.tf` wouldn't it clash with the `main.tf` described in `README.md`?

omar-selo (changes requested):
Looks good, just a couple of comments around `main.tf`

agent/terraform/main.tf
plars: No, we also have this same pattern in the testflinger server terraform. It's fine for them to both be called main.tf and seems to be a pretty common pattern for modules that you import.

agent/terraform/README.md
plars: I thought so too, but if I leave it out then I get an error:
```
╷
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider hashicorp/juju: provider registry registry.terraform.io does not have a provider
│ named registry.terraform.io/hashicorp/juju
│ 
│ Did you intend to use juju/juju? If so, you must specify that source address in each module which requires that provider. To see which modules
│ are currently depending on hashicorp/juju, run the following command:
│     terraform providers
```
Not sure if this is something new in the latest versions of the terraform provider or what, because I didn't think I previously had to do this.

agent/terraform/main.tf
asbalderson: For context here, terraform should be combining all the .tf files into a big chunk of terraform for all the files in the directory. since the example in the readme is a different terraform module, it wont cause a problem.

agent/terraform/main.tf
```diff
@@ -0,0 +1,57 @@
+
+variable "juju_model" {
```
asbalderson, line 2: Even though this is small
it's probably worth following the normal terraform structure where you put 
variables in variables.tf
resources in main.tf
outputs in outputs.tf
and providers in versions.tf

it isnt that the current structure is wrong, but its good to move toward consistency.

agent/terraform/main.tf
omar-selo: I see, thanks for the clarification!

agent/terraform/README.md
omar-selo: That's strange, I also recall this working before. I guess it doesn't hurt to have it if it doesn't work otherwise

omar-selo (approved):
Thanks for addressing my concerns

agent/terraform/main.tf
plars: Good idea, updated!

asbalderson (approved):
noice

that looks great! thanks

PR #365, opened 2024-09-25
[CERTTF-412][CERTTF-413] refactor: treat attachment unpacking as a separate phase

boukeas:
## Description

Retrieving and unpacking the attachments of a Testflinger job is currently handled within the agent code, _before_ the agent starts going through the phases of that job. This imposes certain limitations with regards to how attachment-related failures are handled:
- There are no attachment-related start, success or failure events transmitted. In fact, attachment unpacking currently takes place even before the `job_start` event is emitted. If an error occurs during attachment unpacking, there is currently no way to convey that through events.
- There is no attachment-related result included in the results dict associated with a job. If an error occurs during attachment unpakcing, this is currently no way to convey that through the job results.
- If an error occurs during attachment unpacking it can be seen and investigated through the agent logs. However, there is no output to the user submitting the job (as would be required for resolving e.g. [CERTTF-412](https://warthogs.atlassian.net/browse/CERTTF-412) and subsequently in the job results. Anyone polling a Testflinger job or monitoring the output of a Jenkins job would not be able to see any sort of error message like the one recorded in the agent logs.

One way to lift all these restrictions simultaneously is to treat the retrieval and unpacking of the job attachments as a _separate phase_.

### Changelog

This PR:

- Introduces a new `TestflingerJobPhase` abstract base class, with an interface that captures how all phases are supposed to work procedurally.
- Introduces a new `ExternalCommandPhase` abstract base class, derived from `TestflingerJobPhase`, that captures the workings of phases that run a pre-configured external command. All previously existing phases fall into this category.
- Refactors the code previously in `TestflingerJob.run_test_phase` into separate classes derived from `ExternalCommandPhase`, each corresponding to a different phase.
- Introduces a new `UnpackPhase` derived from  `TestflingerJobPhase`.
- Fixes a minor issue in how `tarfile` is patched for Python 3.8 that allows directories to be included as attachments.

Some points to note while reviewing:

- It makes little sense to review the "diff" for `agent/testflinger_agent/job.py` as the refactoring is considerable. It is best to view the file in its entirety.
- There is now absolutely no attachment-related code under `TestflingerAgent`. It has all been moved into the newly introduced `UnpackPhase`. Functions/methods like `unpack_attachments` and `secure_filter` are now implemented as methods of the `UnpackPhase`, since this is the only phase they are relevant to.
- In a similar vein, phase-specific functionality is restricted to the class representing the corresponding phase. See, for example, the `wait_for_completion` method or the `post_core` implementation for the allocate phase: these used to be methods of `TestflingerJob`, whereas now they are methods of `AllocatePhase`, which is the only phase they are relevant to.
- Both jobs and job phases require access to the same bundle of parameters. This PR introduces the `TestflingerJobParameters` named tuple to hold this bundle. For each job, both the job and its phases store a reference to a single object of this class. This avoids duplication of this information across jobs and phases and, most importantly, allows jobs and phases to be loosely coupled, instead of one circularly referencing the other.
- The `rundir` is no longer provided as an argument to `TestflingerJob.run_test_phase`. This directory is always the same for each job and can be determined when instantiating the job.

## Resolved issues

Resolves [CERTTF-412](https://warthogs.atlassian.net/browse/CERTTF-412) and [CERTTF-413](https://warthogs.atlassian.net/browse/CERTTF-413).

## Documentation

N/A

## Web service API changes

N/A

## Tests

[CERTTF-412]: https://warthogs.atlassian.net/browse/CERTTF-412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CERTTF-412]: https://warthogs.atlassian.net/browse/CERTTF-412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/testflinger_agent/agent.py
+++ b/agent/testflinger_agent/agent.py
@@ -15,56 +15,17 @@
 import json
 import logging
 import os
-from pathlib import Path
 import shutil
 import signal
 import sys
-import tempfile
 
 from testflinger_agent.job import TestflingerJob
 from testflinger_agent.errors import TFServerError
-from testflinger_agent.config import ATTACHMENTS_DIR
 from testflinger_agent.event_emitter import EventEmitter
 from testflinger_common.enums import JobState, TestPhase, TestEvent
 
 
-try:
-    # attempt importing a tarfile filter, to check if filtering is supported
-    from tarfile import data_filter
-
-    del data_filter
-except ImportError:
-    # import a patched version of `tarfile` that supports filtering;
-    # this conditional import can be removed when all agents run
-    # versions of Python that support filtering, i.e. at least:
-    # 3.8.17, 3.9.17, 3.10.12, 3.11.4, 3.12
-    from . import tarfile_patch as tarfile
-else:
-    import tarfile
-
-
 logger = logging.getLogger(__name__)
-
-
-def secure_filter(member, path):
-    """Combine the `data` filter with custom attachment filtering
-
-    Makes sure that the starting folder for all attachments coincides
-    with one of the supported phases, i.e. that the attachment archive
-    has been created properly and no attachment will be extracted to an
-    unexpected location.
-    """
-    try:
-        resolved = Path(member.name).resolve().relative_to(Path.cwd())
-    except ValueError as error:
-        # essentially trying to extract higher than the attachments folder
-        raise tarfile.OutsideDestinationError(member, path) from error
-    if not str(resolved).startswith(
-        ("provision/", "firmware_update/", "test/")
-    ):
-        # trying to extract in an invalid folder, under the attachments folder
-        raise tarfile.OutsideDestinationError(member, path)
-    return tarfile.data_filter(member, path)
 
 
 def parse_error_logs(error_log_path: str, phase: str):
@@ -94,6 +55,7 @@ class TestflingerAgent:
     def __init__(self, client):
         self.client = client
         signal.signal(signal.SIGUSR1, self.restart_signal_handler)
+        # [TODO] What is the relation between the agent state and the job state?
         self.set_agent_state("waiting")
         self._post_initial_agent_data()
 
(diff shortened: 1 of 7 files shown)


boukeas:
@plars @val500 I have noticed that `<phase>_start` and `<phase>_success` events are emitted (and recorded) even for phases that are subsequently skipped. Is this behaviour intentional, i.e. are there any advantages to doing this? I would argue that these events should only be emitted if the corresponding phase is actually executed.

plars (commented):
Before you go any further with this, a couple of questions and observations. tl;dr - what if we could support directories as a valid attachment type instead of yielding an error here?

1. I tried to reproduce this by just making a dir with a single file in it and attaching that. I got the error which looked like this in journalctl:
```
Traceback (most recent call last):                                                     File "/srv/testflinger-agent/rpi4b1g001/env/lib/python3.8/site-packages/testflinger
_agent/agent.py", line 195, in unpack_attachments
    tar.extractall(cwd / ATTACHMENTS_DIR, filter=secure_filter)
  File "/srv/testflinger-agent/rpi4b1g001/env/lib/python3.8/site-packages/testflinger_agent/tarfile_patch.py", line 257, in extractall
    self.chmod(tarinfo, dirpath)
  File "/usr/lib/python3.8/tarfile.py", line 2312, in chmod                              os.chmod(targetpath, tarinfo.mode)
TypeError: an integer is required (got type NoneType)
```
2.  The job actually worked in spite of this! I had a test phase command which just did an `ls -lR` on the attachments dir and it looks like everything actually extracted ok:
```
+ ls -lR attachments
attachments:
total 4
drwxr-xr-x 3 ubuntu ubuntu 4096 Sep 30 17:15 test

attachments/test:
total 4
drwxrwxr-x 2 ubuntu ubuntu 4096 Sep 30 16:55 foof

attachments/test/foof:
total 4
-rwxrwxr-x 1 ubuntu ubuntu 50 Sep 30 16:55 foo.sh
```
3. I tried taking out the secure_filter and it no longer gives this error. Not a solution obviously, but it does help narrow down where to start looking for the problem.

4. Even if we get this working properly, it still might be useful to catch a hard error during this attachment unpack and fail the job at that point. However I'm still confused why it means that we need a whole new phase for unpacking the attachments. I'm sure I'm missing something that you've already explored, but why couldn't we just call emit_event() with a new event type for `ATTACHMENT_ERROR` or something from agent.py?

boukeas:
@plars The attachment error you reproduced is solved with [this commit](https://github.com/canonical/testflinger/pull/365/commits/d9b5660aebdb7464b8a9ad4ad2c73bba32a06799) (part of this PR) and mentioned in the description:

> Fixes a minor issue in how tarfile is patched for Python 3.8 that allows directories to be included as attachments.

So attaching directories is supported and the issue is fixed but what this PR aims to do is also provide support for surfacing any other attachment errors, in a manner consistent with how this is handled in other phases. More in a separate comment.

boukeas:
@plars When you tried to reproduce the error, you had no indication as a Testflinger user that there was indeed an attachment error. If you polled the Testflinger output, there was nothing there. If you requested the job results, there was no `<phase>_status` exit code or any other field to reflect that something failed. And if you were monitoring the events emitted, again there would be no relevant events. These are all mechanisms that a Testflinger user can rely on in order to determine the outcome of a job and its phases but they don't apply to attachment unpacking: you had to go check the agent's log in order to see the error message, which is something that a user cannot do. This is all outlined in the PR description as well.

Of course we could handle attachment unpacking as a special case and do all that (add a result field, emit events and create a runner to generate error output) specifically for attachments but why would we when this is _all_ done for each phase anyway? And, as a bonus, you get what I believe is a very sensible refactoring of the existing phases as well.

boukeas:
Closing this PR as it attempts to resolve multiple issues at once:
- Support for folder attachments has been resolved separately with a minor patch through #368 
- The refactoring proposed in this PR will be examined separately through #371 
- The issue with attachment failures not being included in the job results will be addressed in the context of [CERTTF-423](https://warthogs.atlassian.net/browse/CERTTF-423).

Issue #369, opened 2024-10-03
New agent host charm needs to establish a default ssh config


plars:
The new agent host charm doesn't set a sane default for the agent host config. As a result, it was stuck waiting for user input when trying to run commands over ssh because it couldn't verify the host key on the other end. Since we reprovision systems all the time, we EXPECT that the host key is going to be changing quite frequently.  This is something that should be ignored by default on the agent host.

plars:
This was fixed in https://github.com/canonical/testflinger/pull/370

PR #370, opened 2024-10-03
Add a config option with a sane default for the agent host ssh config

plars:
## Description

In order to avoid ssh failures due to unknown host keys, we need some sane defaults in the ssh config on the agent host. I've made it configurable so that it's easy to make changes if we ever want to, but this sets a default that we already use successfully in the Taipei lab.

## Resolved issues
#369 

## Documentation

Added the new config option to the README

## Web service API changes
N/A

## Tests

Tested locally through deployment


plars:
Merging it to the updated testflinger-agent-host charm branch as Pedro also tested the update applied by this and confirmed it's working for him too.

PR #371, opened 2024-10-04
Agent job refactor

boukeas:


agent/testflinger_agent/agent.py
```diff
@@ -179,168 +116,71 @@ def mark_device_offline(self):
...
-
-                logger.info("Starting job %s", job.job_id)
-                event_emitter.emit_event(
-                    TestEvent.JOB_START,
-                    f"{self.client.server}/jobs/{job.job_id}",
-                )
-                rundir = os.path.join(
-                    self.client.config.get("execution_basedir"), job.job_id
-                )
-                os.makedirs(rundir)
```
plars, line 254: Not really done testing this, but wanted to get started on it... I did see this so far. Because it never forces creation of this dir, if it's missing, then it'll result in problems later:
```
[24-10-08 19:01:02]   ERROR: (cmd.py:32)| cannot access local variable 'job' where it is not associated with a value
  File "/home/plars/work/cert/testflinger/agent/testflinger_agent/cmd.py", line 27, in main
    start_agent()
  File "/home/plars/work/cert/testflinger/agent/testflinger_agent/__init__.py", line 139, in start_agent
    agent.process_jobs()
  File "/home/plars/work/cert/testflinger/agent/testflinger_agent/agent.py", line 159, in process_jobs
    if job.go(phase):
       ^^^
UnboundLocalError: cannot access local variable 'job' where it is not associated with a value
```

agent/testflinger_agent/tests/test_agent.py
```diff
@@ -679,28 +678,24 @@ def test_provision_error_no_cause(self, agent, requests_mock):
             }
         }
 
+        # patch
```
plars, line 681: probably extraneous?

agent/testflinger_agent/job.py
```diff
@@ -27,239 +35,712 @@
...
+                    self.serial_log, "r+", encoding="utf-8", errors="ignore"
+                ) as logfile:
+                    set_truncate(logfile)
+                    outcome_data[f"{self.phase_id}_serial"] = logfile.read()
+            except FileNotFoundError:
+                pass
+            results.seek(0)
+            json.dump(outcome_data, results)
+
+    def banner(self, line):
```
plars, line 192: I've always kinda hated the way this works, but still not sure how to best fix it. I see the reason for it, but tricky to replace the functionality. In the meantime though, I think we don't need it to be a method of the class since it never refers to self. Maybe we could move it outside the class as a helper for now instead? (ok to keep in the same module if you like though - still related)

agent/testflinger_agent/job.py
```diff
@@ -27,239 +35,712 @@
...
+                        exception_info["exception_message"],
+                        exception_info["exception_cause"],
+                    )
+                return detail
+            except (json.JSONDecodeError, KeyError):
+                # [TODO] When do these errors arise?
+                return ""
+
+
+'''
```
plars, line 286: I'm sure you're coming back to clean this... just marking it. I know it's still draft :)

plars (changes requested):
This is one spot that's been begging for a refactor for a while now. I really like your idea and where you are taking it, thanks for taking this on!
I know it's still draft and some obvious things. I was actually able to get it running through some phases in some local testing. I just set up some fake commands for it to run and had it attach to a server I was running in docker locally, then submitted a job to that server for it to pick up. I did find one corner case because it was a new agent and didn't have the directory structure in place for (results, logs, run...). There's an os.makedirs() you removed that I think we'll either need to keep, or handle somewhere to ensure that run dir exists before we try to drop the job file into it.
Looking good so far!

agent/testflinger_agent/agent.py
boukeas: > Because it never forces creation of this dir, if it's missing

The rundir is created [here](https://github.com/canonical/testflinger/blob/41f8a6d2281eda3e505f803f3768cbd024c5ee5b/agent/testflinger_agent/job.py#L505), when the `job` object is `__init__`ed.

That error is raised in the `finally` clause because the `job` object was never properly created so there should be an indication in the agent log about what the actual error was. I just don't think it's the `rundir` not being created.

boukeas: Alright, I think I figured this out. `Path.mkdir` needs the `parents` argument set to `True` in order to mirror how `os.mkdirs` creates intermediate directories recursively. This should be fixed now

agent/testflinger_agent/agent.py
```diff
@@ -179,168 +116,68 @@ def mark_device_offline(self):
...
+            provision_result = job.phases[TestPhase.PROVISION].result
+            if (
+                provision_result is not None
+                and provision_result.event == TestEvent.RECOVERY_FAIL
+            ):
+                self.mark_device_offline()
 
             try:
-                self.client.transmit_job_outcome(rundir)
+                self.client.transmit_job_outcome(str(job.params.rundir))
```
plars, line 172: So I forgot that I had the edge revision of serve from Varun's branch deployed when I was testing this today, and predictably hit a 422 error when it tried to transmit the outcome of a job containing attachments. But then I realized the job was never marked complete. This is because that information is conveyed through transmit_job_outcome. The result is that the CLI is just left in a "stuck" state waiting for the job to complete.

Of course, this isn't really from your patch, but it gave me an "aha" moment for this pathological case that could absolutely occur. I wonder if instead of just failing here, if we could go ahead an try to at least do a client.post_job_state() with the state that we intend to transmit (cancelled or complete) if we fall into the exception block here? That would help with things like this in the future.

common/testflinger_common/enums.py
```diff
@@ -22,6 +22,7 @@
 
 class JobState(StrEnum):
     WAITING = "waiting"
+    UNPACK = "unpack"
```
plars, line 25: should we call this "unpack" or some variation of "attachments"? I think the latter sounds more awkward but also makes it more obvious what's happening in this phase.

Either way, we should make sure the documentation is updated to reflect this new phase. In particular, in docs/reference/test-phases.rst and perhaps others that may describe results schema

plars (changes requested):
A few minor comments. I did a bit of testing with this and it mostly worked for me except for one thing that was totally my fault (forgot the current server revision deployed is the one from Varun's branch).

The main thing to fix here is documentation and possibly a renaming of "unpack" but see what you think. I'm not tied to a specific word there as long as it's documented.

But there's an opportunity to fix another bug here too. I think it might be an easy opportunistic fix, and related enough to include it here. But if not, then feel free to kick it back and we can file it as a bug and handle it in a separate PR.

Also, please go ahead and take this out of draft mode so that everyone knows it's something you think is ready for review and merging as soon as it's good :)

bladernr:
Hi, what is needed to get this taken care of so we can close the Jira issue it's holding up? Is this in a state where it can be made a PR for official review, or should we work on  pulling this elsewhere and superceding this branch?

PR #373, opened 2024-10-09
device_connectors: Add default-user-data to oem_autoinstall

Artur-at-work:
## Description
Add default-user-data file in oem_autoinstall for convenience. 
When user didn't provide user-data file, it will use default one.

## Resolved issues
Closes https://warthogs.atlassian.net/browse/OEX86-473

## Documentation
Updated docs for oem_autoinstall

## Web service API changes
N/A

## Tests
1. Submit job without user_data field and without user-data attachment
2. Wait till provisioning completed on Stella with the default user-data
https://testflinger.canonical.com/jobs/940e48fb-70df-45a3-8179-4687f352e105


kevinyehk (commented):
@Artur-at-work Did this change also include a default authorize_keys?

Artur-at-work:
> @Artur-at-work Did this change also include a default authorize_keys?

No, because I'd have to add the public key into the default-user-data and it will be exposed.
To add the keys to provisioned device, client have to add the authorized_keys file as an attachment.

device-connectors/src/testflinger_device_connectors/devices/oem_autoinstall/oem_autoinstall.py
```diff
@@ -44,6 +44,9 @@ def __init__(self, config, job_data):
             self.config = yaml.safe_load(configfile)
         with open(job_data, encoding="utf-8") as job_json:
             self.job_data = json.load(job_json)
+        self.data_path = (
+            Path(__file__).parent / "../../data/muxpi/oem_autoinstall"
```
plars, line 48: I just realized that this has been using files under data/muxpi, shouldn't it organize them under data/oem_autoinstall instead? That seems like a better place for it, since they are not used by the muxpi device connector.

plars (changes requested):
one comment

PR #381, opened 2024-10-15
Add REBOOT_SCRIPT in TF agent environment

rmartin013:
This patch allows export the reboot_script commands to a REBOOT_SCRIPT env variable that can later be used within the TF agent script in order to reboot the DUT. That's sometimes necessary when dealing with alternate provisioning methods (calling 'eval $REBOOT_SCRIPT').

## Resolved issues

This patch will also allow fixing issues related to https://warthogs.atlassian.net/browse/PERI-556 (Nano or NX being stuck)


plars (approved):
I'm ok with landing this for now, but we should look into other ways of doing this in the future.

PR #382, opened 2024-10-16
Automatically release and power-off maas nodes at the end of a job

plars:
## Description

Some maas-deployed systems use a lot of power. Rather than leaving them on between jobs, we should power them off automatically when the job is complete. The cleanup step is currently only used to ensure the agent container is removed, but can also be used for operations specific to the device connector.

## Resolved issues
CERTTF-427

## Documentation

There are some new log messages saying that the node was released. But there's nothing the user needs to do to activate this, it will happen automatically once the job is complete or cancelled.

## Web service API changes

N/A

## Tests

Tested on a maas deployed node. Attached images on jira card show the job being run, and then the maas screen showing that the device was released and powered off.


val500 (approved):
LGTM

PR #384, opened 2024-10-19
device_connectors: zapper_iot: provide non blocking console command

kiya956:
## Description
Add one more section for support non-blocking send command

## Resolved issues
Prevent some commands like "reboot" stocks our tool due to it could never back to enterable state

## Tests
send "reboot" command through testflinger and zapper to DUT, and make sure it would 
not wait for enterable


tsunghanliu (approved):
LGTM

p-gentili (approved):
+1

PR #386, opened 2024-10-27
Agent host terraform module

plars:
## Description

The existing terraform module doesn't allow specifying constraints for the virtual machine that should be deployed, if you are deploying this to something like LXD. This adds some reasonable defaults and also allows you to override them.

## Resolved issues
N/A

## Documentation
Covered the new options in the README.md

## Web service API changes
N/A

## Tests
Confirmed these options can be used to deploy a new instance, and also confirmed that overriding works properly if you have an old instance with some previous default like "arch=amd64" and need to keep it without terraform trying to remove/re-add it.

### DIFF
--- a/agent/terraform/README.md
+++ b/agent/terraform/README.md
@@ -1,0 +1,70 @@
+# Deploying with Terraform + Juju
+
+This Terraform module can be used for deploying the Testflinger agent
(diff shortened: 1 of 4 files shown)


plars:
Ignore this, it was a similarly named branch to the one I meant to push

PR #388, opened 2024-10-28
Create new script to create client credentials in MongoDB

val500:
## Description
This creates a new script to create, edit, or delete client credentials from MongoDB. MongoDB login credentials must be specified using environment variables. User gets prompted for the new client_id and secret_key and must input the max_priority and allowed_queues data.

## Resolved issues
Resolves https://warthogs.atlassian.net/browse/CERTTF-437

## Web service API changes
N/A

## Tests
Tested locally using docker compose to spin up MongoDB. Creation, editing and deletion were all tested.


server/tools/client_credentials_admin.py
```diff
@@ -0,0 +1,179 @@
...
+"""
+This script administers client credentials in a MongoDB database. Database login
+information must be specified with environment variables. The script prompts user
+to create a new entry, or edit/delete an existing entry. The user will then be prompted to enter new max_priority and allowed_queues data.
+
+"""
+
+import sys
+import json
+import bcrypt
```
plars, line 26: we should go ahead and add this as a requirement on the server side since it's needed for this

server/tools/client_credentials_admin.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def create_client_credential(db):
+    """
+    Creates new client_id, client_secret pair with max_priority
+    and allowed_queues
+    """
+
+    client_id = input("Enter the new client_id: ")
+    client_secret = input("Enter the new client_secret: ")
```
plars, line 109: Lets use getpass for this

server/tools/client_credentials_admin.py
```diff
@@ -0,0 +1,179 @@
...
+
+def handle_max_priority_input() -> dict:
+    """Gets a user inputted max_priority dict"""
+    max_priority = {}
+    while True:
+        max_priority_input = input(
+            "Do you want to specify a max priority for one or more queues? (y/n): "
+        )
+        if max_priority_input == "y":
+            queue_name = input("Enter the name of the queue: ")
```
plars, line 74:
```suggestion
            queue_name = input("Enter the name of the queue(* to apply max priority for all queues): "): ")
```

server/tools/client_credentials_admin.py
```diff
@@ -0,0 +1,179 @@
...
+
+    return client[mongo_db]
+
+
+def handle_max_priority_input() -> dict:
+    """Gets a user inputted max_priority dict"""
+    max_priority = {}
+    while True:
+        max_priority_input = input(
+            "Do you want to specify a max priority for one or more queues? (y/n): "
```
plars, line 71: I feel like the wording of this is unclear... maybe something more like "Specify a queue to set max priority on?" Unless you can think of something better?

plars (changes requested):
A few comments below, but a few more general comments:
1. please run flake8/pylint/black -n79 on this
2. There's a big problem I found - you can add the same client ID more than once. We shouldn't allow that
3. There should be some confirmation step at the end where it shows you what it's going to add (except the secret) and ask to confirm whether to add it or not

server/tools/client_credentials_admin.py
```diff
@@ -0,0 +1,179 @@
...
+            "Do you want to specify a list of queues that you are allowed to use? (y/n): "
+        )
+        if allowed_queues_input == "y":
+            queue_name = input("Enter the name of the queue: ")
+            allowed_queues.append(queue_name)
+        elif allowed_queues_input == "n":
+            break
+        else:
+            print("Invalid input. Please enter 'y' or 'n'.")
+    return allowed_queues
```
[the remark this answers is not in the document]
val500: Yeah not supported yet. I can change it so it also adds these listed queues to the restricted_queues collection.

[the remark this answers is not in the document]
val500: I added a separate option to add/remove entries from that restricted_queues collection

server/tools/client_credentials_admin.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def confirm_dialogue(entry: dict) -> bool:
+    """Prompts the user to confirm their changes"""
+    print(entry)
+    while True:
+        confirm = input("Confirm(y/n) that this entry looks correct: ")
```
plars, line 94: Put the (y/n) at the end to be consistent with the other prompts

server/tools/client_credentials_admin.py
```diff
@@ -0,0 +1,232 @@
...
+    allowed_queues = handle_allowed_queues_input()
+    confirm_output = confirm_dialogue(
+        {
+            "client_id": client_id,
+            "max_priority": max_priority,
+            "allowed_queues": allowed_queues,
+        }
+    )
+    if confirm_output:
+        db.client_permissions.find_one_and_update(
```
plars, line 156: There's no check to see if this actually updated something, and it didn't check that the client_id existed, so if you edit one that doesn't exist, then nothing happens but it also doesn't give you an error saying that it didn't exist

plars (changes requested):
Found a couple more small issues, and I think there were still some earlier comments.

Additionally, I had trouble making it work if you use pip install. I know we use poetry, but we still try to import pip install for the charm, previous way of doing things, etc. And both *should* work fine

I did solve this with a few minor changes. I can just submit a patch against your branch if you prefer... just let me know, but might be easier for you to make the other changes first. But here are the things I needed to change to get it working:
- move tools to src/tools
- add a src/tools/__init__.py
- pyproject.toml: add this under tool.poetry section:
```
+packages = [
+    { include = "src" },
+]
```
- set `client_credentials_admin = "src.tools.client_credentials_admin:main"` (added "src" to the path)

plars (approved):
Ok, I think this works well enough for our short term needs. It's something we'll replace with an API pretty soon I hope, but it'll get us though until then. Thanks!

PR #389, opened 2024-10-31
Add priority test case across multiple queues

val500:
## Description

Adds an extra test case to ensure that job priority works when querying for jobs across multiple queues.

## Resolved issues
N/A

## Documentation
N/A

## Web service API changes
N/A

## Tests
Adds an extra test to test_v1.py


plars (approved):
Just like that! :)

PR #390, opened 2024-11-05
Configure home directory for snaps 

hartmutobendorf:
## Description

testflinger_cli did throw an error when trying to write to ~/.local/share/testflinger-cli-history.json (see below)

I changed the 

```
  File "/snap/testflinger-cli/x1/bin/testflinger-cli", line 8, in <module>
    sys.exit(cli())
  File "/snap/testflinger-cli/x1/lib/python3.10/site-packages/testflinger_cli/__init__.py", line 53, in cli
    tfcli.run()
  File "/snap/testflinger-cli/x1/lib/python3.10/site-packages/testflinger_cli/__init__.py", line 182, in run
    sys.exit(self.args.func())
  File "/snap/testflinger-cli/x1/lib/python3.10/site-packages/testflinger_cli/__init__.py", line 427, in submit
    self.history.new(job_id, queue)
  File "/snap/testflinger-cli/x1/lib/python3.10/site-packages/testflinger_cli/history.py", line 54, in new
    self.save()
  File "/snap/testflinger-cli/x1/lib/python3.10/site-packages/testflinger_cli/history.py", line 74, in save
    with open(
PermissionError: [Errno 13] Permission denied: '/home/hartmut/.local/share/testflinger-cli-history.json'
```

## Documentation

No Tests are included for the changed functionality in this PR - I would need help in setting them up.

## Tests

Manual test:
```
snap install tf_...snap --dangerous
testflinger-cli
testflinger-cli submit example.yaml
```


plars:
@hartmutobendorf Mine seems to automatically pick this up when installing from the snap. Do you know if this is something that might have changed in snapd that we need to do this now? I'm just curious why it wasn't needed previously but now seems to be needed for your system

plars (approved):
Ok, this makes sense now. We found that he had a local profile setting:
```
export XDG_DATA_HOME=$HOME/.local/share
export XDG_CONFIG_HOME=$HOME/.config
```
This was passed through to the snap environment, but the confined snap doesn't have access to those dirs. I think this fix is fine, and could help anyone else who has a similar problem

PR #399, opened 2024-11-09
Generate prometheus metrics for flask and add metrics endpoint

plars:
## Description

Adds a /metrics route for prometheus to pull lots of data about what flask is doing, and also a few testflinger things like job submissions. This can be extended if we want to add more things.

## Resolved issues
N/A

## Documentation
N/A (the more interesting bits will come when we hook it up to grafana)

## Web service API changes
Technically, yes it adds `/metrics` but it's only meant to be consumed by prometheus and is a well-known interface for anything that uses prometheus metrics. It's not a Testflinger API

## Tests
I've hooked this up to a local prometheus/grafana instance and was able to get data from it

### DIFF
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -24,6 +24,7 @@ urllib3 = "^2.2.1"
 pymongo = "<4.9.0"
 pyjwt = "^2.8.0"
 bcrypt = "^4.2.0"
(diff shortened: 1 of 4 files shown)


nadzyah (approved):
A pretty simple change, +1 from me

PR #402, opened 2024-11-12
Refactor muxpi connector

talhaHavadar:
## Description
I wanted to refactor the existing code in muxpi to have room for Partner Engineering specific actions to be placed in the future. I also saw muxpi was lacking tests so I thought it would be nice to increase coverage a bit it is still low but better than before.

I tried to keep changes small didnt really divide functions to keep things similar to ease the review.

## Documentation
No interface has been changed so no need for doc update it is just an internal refactoring. New tests are also added to increase the coverage and tests are written based on the old implementation to make sure no business logic change introduced.

## Tests
Only run `tox` I need help to test this in the field.

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/muxpi/image.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/muxpi/image.py
@@ -1,0 +1,47 @@
+from enum import Enum
+
+
+class ImageType(Enum):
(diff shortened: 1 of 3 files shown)


talhaHavadar:
@LiaoU3 can you also please check the changed to make sure I am not missing something crucial

PR #404, opened 2024-11-13
[CERTTF-388] [CERTTF-443] Implement masking on the agent

boukeas:
## Description

As described in [CR080 - Handling secrets and sensitive information in Testflinger](https://docs.google.com/document/d/1ATsqtCAytZJrWFenELCsY9LK80iyZspXalH4Dlvscd4/edit?usp=sharing), one of the tasks involved in supporting secrets in Testflinger is implementing functionality for masking sensitive information in the output generated by the Testflinger agent.

This PR:
- Introduces a standalone generic `Masker` class for masking sensitive information in text. What is considered sensitive is specified through a list of regular expressions.
- Introduces the `MaskingCommandRunner` class, derived from the regular `CommandRunner`. Their only difference lies in the `post_output` method, where the masking variant of the runner employs a `Masker` to the output before posting it. The `Masker` to be used is actually provided to the runner upon its construction, so the runner itself is not aware of the specifics, i.e. it is not aware of the kind of information that is considered sensitive.
- Extends the `TestflingerJob` class to check if "sensitive patterns" are specified in the agent's configuration file or if secrets are included in the job data. In any of these cases, it employs a `MaskingCommandRunner` instead of a regular `CommandRunner`, to make sure that sensitive information is masked in the output. In the case of secrets, it also injects them into the corresponding environment variables of the runner ([CERTTF-443](https://warthogs.atlassian.net/browse/CERTTF-443)).

## Resolved issues

Resolves [CERTTF-388](https://warthogs.atlassian.net/browse/CERTTF-388) and [CERTTF-443](https://warthogs.atlassian.net/browse/CERTTF-443).

Some (reasonable) assumptions have been made regarding the way the server will pass secrets onto the agent along with other job data. This is separate work to be undertaken within [CERTTF-441](https://warthogs.atlassian.net/browse/CERTTF-441) and possibly [CERTTF-444](https://warthogs.atlassian.net/browse/CERTTF-444), so some minor modifications might be necessary when these tasks are also completed. 

## Tests

Tests have been included for all the newly introduced classes and functionality, including integration-style tests that use the `fake_connector` device connector to run `test_cmds` that include sensitive information and/or secrets.

### DIFF
--- a/agent/testflinger_agent/tests/test_job.py
+++ b/agent/testflinger_agent/tests/test_job.py
@@ -1,5 +1,7 @@
+import json
 import os
 import pytest
+import re
 import shutil
 import tempfile
 
@@ -15,7 +17,7 @@ from testflinger_agent.stop_condition_checkers import (
     GlobalTimeoutChecker,
     OutputTimeoutChecker,
 )
-from testflinger_common.enums import TestEvent
+from testflinger_common.enums import TestEvent, TestPhase
 
 
 class TestJob:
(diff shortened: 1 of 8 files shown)


agent/testflinger_agent/tests/test_job.py
```diff
@@ -173,6 +175,111 @@ def test_run_test_phase_with_run_exception(
...
+        # this will result in the use of a masking command runner
+        self.config["sensitive_patterns"] = ["[Cc][a-z]*n"]
+
+        # create the job data
+        job_data = {
+            "test_data": {"test_cmds": "echo -n Message: Captain Major"}
+        }
+
+        log_data = self.run_testcmds(job_data, client, tmp_path)
+        print(log_data)
```
nadzyah, line 219: are you supposed to print log data to stdout, not stderr?

boukeas: This is a test, so neither `stdout` nor `stderr` are printed when the test is successful. They are however both captured and printed (separately) when the test fails, so this `print` is meant as a diagnostic, i.e. you only see it when the test fails and might help you understand what went wrong. I'd happily change this to print to `stderr` if you feel it makes any difference.

plars:
@boukeas I have a really stupid question to ask first... when/why do we want to use regexp to specify patterns that need to be masked, and why would they be in the agent config? I guess I just assumed that the only masking would be secrets that were sent to us from the server - which would be literal string matching in the output and not defined in the agent config. The specific example cited here is masking the IP address, but I don't think we ever want to mask that information do we? That could make interpreting the job really hard in many cases.

PR #406, opened 2024-11-14
Restricted Queues

val500:


docs/explanation/index.rst
```diff
@@ -9,4 +9,5 @@ This section covers conceptual questions about Testflinger.
     agents
     queues
     job-priority
+    restricted-queues
```
plars, line 12: This is referenced in the index.rst, but should have a file for it in docs/explanation/restricted-queues.rst but I don't see it.

plars (changes requested):

docs/explanation/authentication.rst
```diff
@@ -3,7 +3,8 @@ Authentication and Authorisation
 
 Authentication requires a client_id and a secret_key. These credentials can be
 obtained by contacting the server administrator with the queues you want priority
-access for as well as the maximum priority level to set for each queue. The
-expectation is that these credentials are shared between users on a team. 
+access for, the maximum priority level to set for each queue, and any restricted
+queues that you need access to. The expectation is that these credentials are
+shared between users on a team.
```
plars, line 8: I think you can eliminate the last sentence about "The expectation is that these credentials are shared between users on a team." That's a bit of a detail with how we're using it, but in reality it could be a single user, a bot account, or whatever. There's no need to narrow the scope of how it's used here.

server/src/api/v1.py
```diff
@@ -100,13 +100,17 @@ def has_attachments(data: dict) -> bool:
     )
 
 
-def check_token_priority_permission(
+def check_token_permissions(
```
plars, line 103: The logic here is not too bad yet, but it's getting harder to follow. It's an important piece to get right, so we really want to make sure that it's testable and conceptually obvious.
Can we find a way to break this up so that it goes through a series of independent checks? For instance, one method check that either priority is 0 or it has permission to set the higher priority. The next method checks to see if it is either using an unrestricted queue, or it has permission to use the queue it asked for. There will be other checks in the future, such as whether it's using vault paths that it has access to, etc.
If it passes all of them, then the token is ok. Otherwise we reject it as soon as we see something is wrong.

plars (changes requested):

server/src/api/v1.py
```diff
@@ -156,27 +194,25 @@ def job_builder(data: dict, auth_token: str):
...
-                    f"Not enough permissions to push to {job_queue}",
-                    f"with priority {priority_level}",
-                ),
-            )
-        job["job_priority"] = priority_level
-        data.pop("job_priority")
-    else:
-        job["job_priority"] = 0
+    job["job_priority"] = priority_level
+    data.pop("job_priority", None)
```
plars, line 214: Why do we pop here?

plars (changes requested):
One very minor question, but I think once that's sorted then we can land this - then do some further testing in staging before deployment to production. There's some policy documentation that needs to be written, but I think that's more specific to our own policies of how we use it, so we'll need to sort out where that should go.

server/src/api/v1.py
val500: I don't think it's strictly necessary - I pop only so that job_priority isn't duplicated in both the job dict and the data dict. Is there a reason to include it in the data dictionary as well?

plars: but the data dict is just the data that's passed in from the original json job submission right? so afaict it's not used anywhere after that, so modifying it is just removing a key from data that you're about to throw away... unless I'm not seeing something?

val500: The data gets added to the job dict right? In line 217

plars (approved):
Looks good now, let's land and do some further testing in staging.

PR #391, opened 2024-11-05
Configure XDG_*_HOME directories based on SNAP_USER_DATA

hartmutobendorf:
## Description

testflinger_cli did throw an error when trying to write to ~/.local/share/testflinger-cli-history.json (see below)

I changed the 

```
Traceback (most recent call last):
  File "/snap/testflinger-cli/x1/bin/testflinger-cli", line 8, in <module>
    sys.exit(cli())
  File "/snap/testflinger-cli/x1/lib/python3.10/site-packages/testflinger_cli/__init__.py", line 53, in cli
    tfcli.run()
  File "/snap/testflinger-cli/x1/lib/python3.10/site-packages/testflinger_cli/__init__.py", line 182, in run
    sys.exit(self.args.func())
  File "/snap/testflinger-cli/x1/lib/python3.10/site-packages/testflinger_cli/__init__.py", line 427, in submit
    self.history.new(job_id, queue)
  File "/snap/testflinger-cli/x1/lib/python3.10/site-packages/testflinger_cli/history.py", line 54, in new
    self.save()
  File "/snap/testflinger-cli/x1/lib/python3.10/site-packages/testflinger_cli/history.py", line 74, in save
    with open(
PermissionError: [Errno 13] Permission denied: '/home/hartmut/.local/share/testflinger-cli-history.json'
```

## Documentation

No Tests are included for the changed functionality in this PR - I would need help in setting them up.

## Tests

Manual test:
```
snap install tf_...snap --dangerous
testflinger-cli
testflinger-cli submit example.yaml
```

PR #407, opened 2024-11-15
Fix snap environment variables

pedro-avalos:
## Description

- Similar to #391, which is blocked since the commit is unsigned.
- The `testflinger-cli.wrapper` was just setting environment variables, so it was not really necessary.
  - This can be done within the `snapcraft.yaml` file.

## Resolved issues

- Resolves issues where users have XDG directories defined in their bashrc.

## Tests

- Tested by building and installing locally


plars (approved):
nice, +1

PR #408, opened 2024-11-16
rename tplan to more appropriate name "provision_plan"

kiya956:
## Description
rename tplan to provision_plan to avoid confusion

## Tests
Update task.yaml for x8high-pdk and med-pdk and provision successfully


plars (changes requested):
Will this break any existing users of it? It's definitely a more descriptive name, but I also don't see any documentation about it. It needs to have documentation in the /docs dir for this option and how to use it.

kiya956:
> Will this break any existing users of it? It's definitely a more descriptive name, but I also don't see any documentation about it. It needs to have documentation in the /docs dir for this option and how to use it.

Hi Paul. We currently don't any existing users but me. And about the documentation, the PR was closed by Paolo and asked me to move the doc to zapper repo.

plars (approved):
Ah, that makes sense in this case, just kind of different from what we usually need to do. +1 on the name change

PR #417, opened 2024-11-21
Add testflinger-testenv container

plars:
## Description

When running the test phase for agents, we currently have it run under a docker image pulled from dockerhub at plars/testflinger-testenv-focal (don't be too bothered that it's focal based, it's just to give some working env on the agent-host and this one is well tested).

Now that we're on GH for everything, we should just use ghcr.io, so that nobody has to switch it over to a new location when that bus hits me some day :)

This is the first step in resolving this. I intentionally didn't change the path pointed to by the tf-test script that the agent calls **yet**, because we need to make sure firewall rules are all adjusted and everything before rolling this out in production.

## Resolved issues
N/A

## Documentation
There's a small README there just saying what it is, but this doesn't really do anything different, just moves the location

## Web service API changes
N/A

## Tests
Successful action run: https://github.com/canonical/testflinger/actions/runs/11961073449
Published container image: https://github.com/canonical/testflinger/pkgs/container/testflinger%2Ftestflinger-testenv

I also tried pulling this container image and running it locally.

### DIFF
--- a/.github/workflows/build-testflinger-testenv-container.yml
+++ b/.github/workflows/build-testflinger-testenv-container.yml
@@ -19,6 +19,10 @@ jobs:
       contents: read
       packages: write
 
+    strategy:
+      matrix:
+        version: [focal, jammy, noble]
+
     steps:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
@@ -42,6 +46,8 @@ jobs:
         uses: docker/metadata-action@v5
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ matrix.base_version }}
 
       - name: Build and push backend Docker image
         uses: docker/build-push-action@v6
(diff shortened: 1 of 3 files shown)


plars:
A few more changes:
- Added some packages that George requested
- matrix build so that it makes a tagged version of this for focal, jammy, and noble. This way we could try out newer versions if we want.

.github/workflows/build-testflinger-testenv-container.yml
```diff
@@ -0,0 +1,61 @@
+name: Build testflinger-testenv container image
+
+on:
+  push:
+    branches:
+      - main
+      - testflinger-testenv
```
nancyc12, line 7: should we remove `testflinger-testenv`?

nancyc12 (dismissed):
Thanks for making this workflow! I only have a small question :)

.github/workflows/build-testflinger-testenv-container.yml
plars: fixed

nancyc12 (approved):
LGTM+1

nancyc12 (approved):

PR #420, opened 2024-11-26
Update device-connector-types.rst

jocave:
## Description

Correct some contradictory text and state the requirement on image type for oemscript derived connector types

## Documentation

Only doc changes

## Web service API changes

None

## Tests

None


plars (approved):
hah, thanks for catching that! It looks like I was the one that did it. Best guess is that I copy/pasted the noprovision one then forgot to go back and fix that line.

PR #423, opened 2024-12-01
Check for online agents when submitting a job

plars:
## Description

This implements a new server API to get a list of the agents listening on a queue. That's used by the CLI to figure out if any of them are online before submitting a job. If there are no agents listening on the specified queue, or if all of them are offline, then it uses a new parameter to the CLI to determine what to do: `--wait-for-available-agents`

If that flag is used, then it will warn the user that no agents seem to be listening, and wait for the job to run anyway. Otherwise, the default behavior is to print an error and exit, failing to submit the job.

## Resolved issues

CERTTF-338

## Documentation

Added API documentation to the server README.rst and also extended the schema so that the /docs API reference is updated.

## Web service API changes

This is detailed in the API reference (/docs path on the server) as well as in the README.rst for the server:

** [GET] /v1/queues/<queue_name>/agents** - Get the list of agents listening to a specified queue
- Parameters:
  - queue_name (string): name of the queue for which to get the agents that are listening to it
- Returns: 
  - JSON array of agents listening to the specified queue
- Example:
```
$ curl http://localhost:8000/v1/queues/foo/agents
```

## Tests

Tested locally and with added unit tests

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -343,6 +343,7 @@ class TestflingerCli:
         parser.set_defaults(func=self.submit)
         parser.add_argument("--poll", "-p", action="store_true")
         parser.add_argument("--quiet", "-q", action="store_true")
+        parser.add_argument("--wait-for-available-agents", action="store_true")
         parser.add_argument("filename").completer = (
             argcomplete.completers.FilesCompleter(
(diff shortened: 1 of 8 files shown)


plars:
I'm aware of the pylint issue, but Pedro already has a nice fix for this in a branch he's working on, so I'll merge it with that once his lands

pedro-avalos (approved):
Looks good

PR #424, opened 2024-12-04
Add better messages for 401 and 403 errors from job submissions

val500:
## Description

Simple change to add some extra messages in Testflinger CLI when submitting a job that requires authorization.

## Resolved issues
N/A

## Documentation
N/A

## Web service API changes

N/A
## Tests
N/A

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -552,7 +552,8 @@ class TestflingerCli:
             if exc.status == 403:
                 sys.exit(
                     "Received 403 error from server with reason "
-                    f"{exc.msg}The specified client credentials do "
+                    f"{exc.msg}"
+                    "The specified client credentials do "
                     "not have sufficient permissions for the resource(s) "
                     "you are trying to access."
                 )
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -541,6 +541,22 @@ class TestflingerCli:
                     "Received 404 error from server. Are you "
                     "sure this is a testflinger server?"
                 )
+            if exc.status == 401:
+                sys.exit(
+                    "Received 401 error from server. You are "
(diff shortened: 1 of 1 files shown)


plars (changes requested):
A few adjustments here...
We already know that we're going to use client credentials for a lot of other things and we don't want to come back here and fix the errors to specify them every time we change it, so let's be a little more generic with the error. I'm not trying to be prescriptive, so feel free to change it, but something more like this perhaps:
```
You are attempting to use a feature that requires client authorisation without using client credentials. See <documentation link> for more details
```

and for the other one, maybe something like:
```
The specified client credentials do not have sufficient permissions for the resource(s) you are trying to access.
```

cli/testflinger_cli/__init__.py
```diff
@@ -541,6 +541,22 @@ def submit_job_data(self, data: dict, headers: dict = None):
...
+                    "Received 401 error from server. You are "
+                    "attempting to use a feature that requires "
+                    "client authorisation without using client "
+                    "credentials. See https://testflinger.readthedocs"
+                    ".io/en/latest/how-to/authentication/ for more details"
+                )
+            if exc.status == 403:
+                sys.exit(
+                    "Received 403 error from server with reason "
+                    f"{exc.msg}The specified client credentials do "
```
plars, line 555:
```suggestion
                    f"{exc.msg}"
                    "The specified client credentials do "
```

plars (changes requested):
I was going to say that you are missing a space after `{exc.msg}` which could make it run together with the following text, but I think it might be more readable just to continue on a new line after the message. What do you think?

cli/testflinger_cli/__init__.py
val500: Yeah I did that since exc.msg already had a newline character at the end for some reason. I can do that change for readability.

plars (approved):
+1

PR #426, opened 2024-12-09
testflinger cli promote snap stable

plars:
- **Add a github workflow to promote testflinger-cli snap from beta to stable**
- **When originally publishing the snap, push it to both edge and beta**

## Description

This introduces a workflow that allows promoting the testflinger-cli snap to stable.

We already have a workflow that publishes the snap to edge, but this modifies it slightly to also publish to beta. This is because the `snapcraft promote` command doesn't allow you to start on edge for some reason. So the only way to promote it, is to start on some channel other than edge.

I made this flexible, so that you *can* specify the channels if you want, or just in case we ever want to reuse it.

## Resolved issues
N/A

## Documentation
N/A

## Web service API changes
N/A

## Tests
https://github.com/canonical/testflinger/actions/runs/12242201005

### DIFF
--- a/.github/workflows/cli-publish-snap.yml
+++ b/.github/workflows/cli-publish-snap.yml
@@ -1,4 +1,4 @@
-name: Publish Testflinger CLI snap to edge
+name: Publish Testflinger CLI snap to edge and beta
 on:
   push:
     branches: ["main"]
@@ -21,3 +21,9 @@ jobs:
       with:
         snap: ${{ steps.build.outputs.snap }}
         release: edge
+    - uses: snapcore/action-publish@v1
+      env:
+        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAP_STORE_LOGIN }}
+      with:
+        snap: ${{ steps.build.outputs.snap }}
+        release: beta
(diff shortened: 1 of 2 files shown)


pedro-avalos (approved):
+1

.github/workflows/cli-publish-snap.yml
```diff
@@ -21,3 +21,9 @@ jobs:
       with:
         snap: ${{ steps.build.outputs.snap }}
         release: edge
+    - uses: snapcore/action-publish@v1
+      env:
+        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAP_STORE_LOGIN }}
+      with:
+        snap: ${{ steps.build.outputs.snap }}
+        release: beta
```
pedro-avalos, line 29: Why not push to beta and just close edge, then edge should follow the same version as beta, no?

pedro-avalos (commented):
Why push to both edge and beta?

pedro-avalos (approved):
I think it's fine to push to both, was just curious about it. The only benefit to push to beta and close edge is to avoid having both commands in the workflow, but that is a very fractional benefit (if one at all)

PR #427, opened 2024-12-10
Fix for Testflinger `submit` action

boukeas:
## Description

Following recent changes to the `testflinger-cli` snap, a problem with the `submit` action surfaced: a pair of quotes were erroneously used around the `$RELATIVE` variable. 

During testing, it was also determined that the pre-requisite snaps for the workflow should be installed separately, because a possible failure to install for one of them (e.g. because it has already been installed) would prevent the rest from installing as well.

This PR implements these two fixes.

## Resolved issues

This is a hotfix, not tracked with a JIRA issue.

## Documentation

N/A

## Tests

Multiple jobs were deployed from this branch, e.g. see [this run](https://github.com/canonical/sqa-cloud-deployment-pipeline/actions/runs/12259828532/job/34202939932#step:13:123).

### DIFF
--- a/.github/actions/submit/action.yaml
+++ b/.github/actions/submit/action.yaml
@@ -101,7 +101,8 @@ runs:
       shell: bash
       run: |
         echo "::group::Install prerequisites"
-        sudo snap install testflinger-cli jq
+        sudo snap install testflinger-cli
+        sudo snap install jq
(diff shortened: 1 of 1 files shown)


pedro-avalos (approved):
Looks good, that was a very weird bug, for sure

PR #428, opened 2024-12-10
Add extended reservation checks in the server

val500:
## Description
This PR adds support for extended reservation times in Testflinger Server. Certain users need longer reservation times than the maximum 6 hours, so this provides this ability for authorized clients. The server rejects jobs with higher than 6 hours of requested reservation time if they do not provide credentials or if their credentials do not provide them the authorization. The authentication flow is similar to that of job priority and restricted queues.

## Resolved issues
Resolves https://warthogs.atlassian.net/browse/CERTTF-458

## Documentation
Documentation was added to the docs/ folder.

## Web service API changes
No new endpoints, but jobs that have longer than 6 hours requested of reservation times will be rejected without prior authorization.

## Tests
Unit tests were added to test_v1_authorization.py

### DIFF
--- a/server/src/api/v1.py
+++ b/server/src/api/v1.py
@@ -170,7 +170,9 @@ def check_token_reservation_timeout(
         return True
     decoded_jwt = decode_jwt_token(auth_token, secret_key)
     max_reservation_time_dict = decoded_jwt.get("max_reservation_time", {})
-    max_reservation_time = max_reservation_time_dict.get(queue, 0)
+    queue_reservation_time = max_reservation_time_dict.get(queue, 0)
+    star_reservation_time = max_reservation_time_dict.get("*", 0)
+    max_reservation_time = max(queue_reservation_time, star_reservation_time)
     return reservation_timeout <= max_reservation_time
 
 
--- a/server/src/api/v1.py
+++ b/server/src/api/v1.py
@@ -806,9 +806,8 @@ def generate_token(allowed_resources, secret_key):
         "exp": expiration_time,
         "iat": datetime.now(timezone.utc),  # Issued at time
         "sub": "access_token",
-        "permissions": {},
+        "permissions": allowed_resources,
     }
-    token_payload.get("permissions").update(allowed_resources)
     token = jwt.encode(token_payload, secret_key, algorithm="HS256")
     return token
 
--- a/server/src/api/v1.py
+++ b/server/src/api/v1.py
@@ -784,7 +784,11 @@ def get_agents_on_queue(queue_name):
 
 
 def generate_token(allowed_resources, secret_key):
-    """Generates JWT token with queue permission given a secret key"""
+    """
+    Generates JWT token with queue permission given a secret key
+    See retrieve_token for more information on the contents of
+    the token payload
+    """
     expiration_time = datetime.utcnow() + timedelta(seconds=2)
     token_payload = {
         "exp": expiration_time,
--- a/server/src/api/v1.py
+++ b/server/src/api/v1.py
@@ -832,6 +832,7 @@ def validate_client_key_pair(client_id: str, client_key: str):
     ):
         return None
     client_permissions_entry.pop("_id", None)
+    client_permissions_entry.pop("client_secret_hash", None)
     return client_permissions_entry
 
 
--- a/server/src/api/v1.py
+++ b/server/src/api/v1.py
@@ -179,20 +179,24 @@ def check_token_reservation_timeout(
 def check_token_permissions(
     auth_token: str,
     secret_key: str,
-    priority: int,
-    queue: str,
-    reservation_timeout: int,
+    job_data: dict,
 ) -> bool:
     """
     Validates token received from client and checks if it can
     push a job to the queue with the requested priority
     """
+    priority_level = job_data.get("job_priority", 0)
+    job_queue = job_data["job_queue"]
     priority_allowed = check_token_priority(
-        auth_token, secret_key, queue, priority
+        auth_token, secret_key, job_queue, priority_level
     )
-    queue_allowed = check_token_queue(auth_token, secret_key, queue)
+    queue_allowed = check_token_queue(auth_token, secret_key, job_queue)
+
+    reserve_data = job_data.get("reserve_data", {})
+    # default reservation timeout is 1 hour
+    reservation_timeout = reserve_data.get("timeout", 3600)
     reservation_time_allowed = check_token_reservation_timeout(
-        auth_token, secret_key, reservation_timeout, queue
+        auth_token, secret_key, reservation_timeout, job_queue
     )
     return priority_allowed and queue_allowed and reservation_time_allowed
 
@@ -222,15 +226,10 @@ def job_builder(data: dict, auth_token: str):
 
     priority_level = data.get("job_priority", 0)
     job_queue = data["job_queue"]
-    reserve_data = data.get("reserve_data", {})
-    # default reservation timeout is 1 hour
-    reservation_timeout = reserve_data.get("timeout", 3600)
     allowed = check_token_permissions(
         auth_token,
         os.environ.get("JWT_SIGNING_KEY"),
-        priority_level,
-        job_queue,
-        reservation_timeout,
+        data,
     )
     if not allowed:
         abort(
@@ -792,14 +791,7 @@ def generate_token(allowed_resources, secret_key):
         "iat": datetime.now(timezone.utc),  # Issued at time
         "sub": "access_token",
     }
-    if "max_priority" in allowed_resources:
-        token_payload["max_priority"] = allowed_resources["max_priority"]
-    if "allowed_queues" in allowed_resources:
-        token_payload["allowed_queues"] = allowed_resources["allowed_queues"]
-    if "max_reservation_time" in allowed_resources:
-        token_payload["max_reservation_time"] = allowed_resources[
-            "max_reservation_time"
-        ]
+    token_payload.update(allowed_resources)
     token = jwt.encode(token_payload, secret_key, algorithm="HS256")
     return token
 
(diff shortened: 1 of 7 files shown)


plars:
@val500 It looks pretty straightforward so far, but of course needs documentation. Not sure you intended it to be ready for review yet since it's not filled in and marked though.

server/src/api/v1.py
```diff
@@ -766,6 +796,10 @@ def generate_token(allowed_resources, secret_key):
         token_payload["max_priority"] = allowed_resources["max_priority"]
     if "allowed_queues" in allowed_resources:
         token_payload["allowed_queues"] = allowed_resources["allowed_queues"]
+    if "max_reservation_time" in allowed_resources:
+        token_payload["max_reservation_time"] = allowed_resources[
```
boukeas, line 800: I think it would be extremely useful if there was some documentation around the structure and contents of these "client permissions" or "allowed_resources". A schema, a reference or even an example somewhere would really really help. Also, I am wondering if using a different name in the database (client permissions) and the code (allowed resources) was deliberate? Aren't these two referring to the same concept?

server/src/api/v1.py
```diff
@@ -196,11 +222,15 @@ def job_builder(data: dict, auth_token: str):
 
     priority_level = data.get("job_priority", 0)
     job_queue = data["job_queue"]
+    reserve_data = data.get("reserve_data", {})
+    # default reservation timeout is 1 hour
+    reservation_timeout = reserve_data.get("timeout", 3600)
```
boukeas, line 227: I think there's a corresponding line in the device connector as well. Is that still necessary?

server/src/api/v1.py
```diff
@@ -157,8 +157,31 @@ def check_token_queue(auth_token: str, secret_key: str, queue: str) -> bool:
...
 
+def check_token_reservation_timeout(
+    auth_token: str, secret_key: str, reservation_timeout: int, queue: str
+) -> bool:
+    """
+    Checks if the requested reservation is either less than the max
+    or that their token gives them the permission to use a higher one
+    """
+    # Max reservation time defaults to 6 hours
+    max_reservation_time = 6 * 60 * 60
```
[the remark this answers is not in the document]
boukeas: Likewise, the default timeout could also be configurable.

server/src/api/v1.py
```diff
@@ -157,8 +157,31 @@ def check_token_queue(auth_token: str, secret_key: str, queue: str) -> bool:
...
+    return reservation_timeout <= max_reservation_time
+
+
 def check_token_permissions(
-    auth_token: str, secret_key: str, priority: int, queue: str
+    auth_token: str,
+    secret_key: str,
+    priority: int,
+    queue: str,
+    reservation_timeout: int,
```
boukeas, line 184: Every time you add a new permissions check you will probably need to add more parameters here. I think these three should be replaced by the entire job dict (I think it's called `data` in the `job_builder` method where this is called). What this method is doing (conceptually) is to check whether the permissions specified in the `auth_token` allow this job to be submitted. You can add as many permission checks as you like, the method signature would still be the same.

server/src/api/v1.py
```diff
@@ -168,7 +191,10 @@ def check_token_permissions(
         auth_token, secret_key, queue, priority
     )
     queue_allowed = check_token_queue(auth_token, secret_key, queue)
-    return priority_allowed and queue_allowed
+    reservation_time_allowed = check_token_reservation_timeout(
```
boukeas, line 194: I would say this permission check only makes sense if this is a reservation job, right?

boukeas (changes requested):
I've left quite a lot of comments here and there -- many of them are questions. I am not too familiar with the detail of how all this comes together and I am unaware of any previous discussions you might have had with Paul so I apologize in advance for any of the comments that might be missing the mark.

server/src/api/v1.py
```diff
@@ -157,8 +157,31 @@ def check_token_queue(auth_token: str, secret_key: str, queue: str) -> bool:
...
+    """
+    # Max reservation time defaults to 6 hours
+    max_reservation_time = 6 * 60 * 60
+    if reservation_timeout <= max_reservation_time:
+        return True
+    decoded_jwt = decode_jwt_token(auth_token, secret_key)
+    max_reservation_time_dict = decoded_jwt.get("max_reservation_time", {})
+    queue_reservation_time = max_reservation_time_dict.get(queue, 0)
+    star_reservation_time = max_reservation_time_dict.get("*", 0)
+    max_reservation_time = max(queue_reservation_time, star_reservation_time)
```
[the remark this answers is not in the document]
val500: max_reservation_time should be 6 * 60 * 60 if you look a couple lines above. First we check whether or not the requested reservation time is less than that.

server/src/api/v1.py
val500: I think so - this value isn't passed to the device connector unless explicitly specified in the job

server/src/api/v1.py
```diff
@@ -157,8 +157,31 @@ def check_token_queue(auth_token: str, secret_key: str, queue: str) -> bool:
...
+) -> bool:
+    """
+    Checks if the requested reservation is either less than the max
+    or that their token gives them the permission to use a higher one
+    """
+    # Max reservation time defaults to 6 hours
+    max_reservation_time = 6 * 60 * 60
+    if reservation_timeout <= max_reservation_time:
+        return True
+    decoded_jwt = decode_jwt_token(auth_token, secret_key)
```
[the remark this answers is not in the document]
val500: The user might not have passed in a token if they didn't use priority/restricted queues/extended reservations. The other functions first check to make sure that the token is needed for what they have requested.

server/src/api/v1.py
val500: Yeah, but if the job isn't a reservation job, it will allow it.

server/src/api/v1.py
boukeas: Perhaps in the docstring of this function we could add a reference like "see `retrieve_token` for more information on the contents of the token payload".

server/src/api/v1.py
```diff
@@ -766,6 +796,10 @@ def generate_token(allowed_resources, secret_key):
         token_payload["max_priority"] = allowed_resources["max_priority"]
     if "allowed_queues" in allowed_resources:
         token_payload["allowed_queues"] = allowed_resources["allowed_queues"]
+    if "max_reservation_time" in allowed_resources:
+        token_payload["max_reservation_time"] = allowed_resources[
+            "max_reservation_time"
+        ]
```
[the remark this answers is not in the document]
boukeas: Let me clarify something: the `allowed_resources` comes from the return value of `validate_key_client_pair` which comes directly from the database. And I can see in the code of `validate_key_client_pair` that the object retrieved from the database (it's called `client_permissions_entry` internally) has a `client_secret_hash` field. We don't want that in our token, do we? Can't we just pop it before returning the database object (like you do for `_id` for example)?

server/src/api/v1.py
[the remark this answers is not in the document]
boukeas: If the user has requested a reservation time of over 6 hours then we will reach this code and if there are no entries for `*` or this specific queue then `max_reservation_time` will be zero. Which is fine, in a way, because the job will be dropped but I find the logic hard to read.

server/src/api/v1.py
```diff
@@ -199,8 +229,7 @@ def job_builder(data: dict, auth_token: str):
     allowed = check_token_permissions(
```
boukeas, line 229: Here, you are returning `allowed` but if that is false you have lost the information about _why_ it is false, i.e. exactly what the permission problem is. So right underneath, you are aborting with an error message that has to do with priorities, even though the rejection might have different reasons, e.g. max reservation time. So perhaps you can move the (different) `abort`s inside the `check_token_permissions` function and display an appropriate message depending on the permission issue. Returning successfully from `check_token_permissions` will mean there are no permission issues.

server/src/api/v1.py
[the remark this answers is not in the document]
boukeas: If the user hasn't passed in a token then why would we be checking token permissions?

server/src/api/v1.py
```diff
@@ -755,17 +784,18 @@ def get_agents_on_queue(queue_name):
...
     token_payload = {
         "exp": expiration_time,
         "iat": datetime.now(timezone.utc),  # Issued at time
         "sub": "access_token",
     }
-    if "max_priority" in allowed_resources:
-        token_payload["max_priority"] = allowed_resources["max_priority"]
-    if "allowed_queues" in allowed_resources:
-        token_payload["allowed_queues"] = allowed_resources["allowed_queues"]
+    token_payload.update(allowed_resources)
```
boukeas, line 798: Do you think it might be a good idea to add a `permissions` field to the payload, next to `exp`, `iat` and `sub`, and then have the "allowed resources" live under that? I feel it might add a little more structure to the payload and also allow for the future addition of things like scopes or roles.

server/src/api/v1.py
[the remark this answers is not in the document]
val500: It's checking permissions more broadly- if the user submits something with 0 permissions for example, they wouldn't need a token

server/src/api/v1.py
```diff
@@ -755,17 +796,19 @@ def get_agents_on_queue(queue_name):
...
         "exp": expiration_time,
         "iat": datetime.now(timezone.utc),  # Issued at time
         "sub": "access_token",
+        "permissions": {},
     }
-    if "max_priority" in allowed_resources:
-        token_payload["max_priority"] = allowed_resources["max_priority"]
-    if "allowed_queues" in allowed_resources:
-        token_payload["allowed_queues"] = allowed_resources["allowed_queues"]
+    token_payload.get("permissions").update(allowed_resources)
```
boukeas, line 811: Unless I am missing something, this line can be removed and the `allowed_resources` can directly be assigned to the `permissions` field of the token payload, i.e.:
```
    token_payload = {
        "exp": expiration_time,
        "iat": datetime.now(timezone.utc),  # Issued at time
        "sub": "access_token",
        "permissions": allowed_resources,
    }
```

boukeas (approved):
I think the structure reflects the logic pretty well now. I have asked for one final refinement but please check that it makes sense and that it works properly. I am approving in any case so that I don't slow you down. Thank you for all this.

server/src/api/v1.py
val500: Thank you for the review! I think I need another approval after making these changes.

boukeas (approved):

PR #429, opened 2024-12-10
Add jwt_signing_key as a supported terraform variable

plars:
## Description
We need to pull the jwt signing key from vault, but first we need it to be something that can be applied with terraform.

## Resolved issues
N/A

## Documentation
N/A - but it's got a good description in the terraform variables.tf

## Web service API changes
N/A

## Tests
None - won't do anything on it's own. The other piece of this will be to add it to certification ops and have it pull from vault.


val500 (approved):
LGTM

PR #433, opened 2024-12-20
Update credential script to add ability to set maximum reservation time

val500:
## Description

Updates the client credentials script to allow for adding max_reservation_time. Follows similar logic to the job_priority part of the script.

## Resolved issues
Resolves https://warthogs.atlassian.net/browse/CERTTF-459

## Documentation
N/A

## Web service API changes
N/A

## Tests
Tested locally with a local deployment of Testflinger Server/MongoDB

### DIFF
--- a/server/src/tools/client_credentials_admin.py
+++ b/server/src/tools/client_credentials_admin.py
@@ -56,8 +56,8 @@ def handle_max_priority_input() -> dict:
         )
         if max_priority_input == "y":
             queue_name = input(
-                "Enter the name of the queue(* to apply max priority for"
-                + " all queues): "
+                "Enter the name of the queue(* to apply max priority for",
+                " all queues): ",
             )
             max_priority[queue_name] = int(
                 input("Enter the max priority for this queue: ")
@@ -74,8 +74,8 @@ def handle_allowed_queues_input() -> list:
     allowed_queues = []
     while True:
         allowed_queues_input = input(
-            "Do you want to specify a restricted queue that"
-            + " you are allowed to use? (y/n): "
+            "Do you want to specify a restricted queue that",
+            " you are allowed to use? (y/n): ",
         )
         if allowed_queues_input == "y":
             queue_name = input("Enter the name of the queue: ")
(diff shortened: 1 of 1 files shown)


plars (commented):
Looks ok, consistent with the other things at least. Needs the other part of this to land (see comments from George) and also seems to have a minor lint issue to fix

server/src/tools/client_credentials_admin.py
```diff
@@ -87,6 +87,29 @@ def handle_allowed_queues_input() -> list:
     return allowed_queues
 
 
+def handle_max_reservation_input() -> list:
+    """Gets a user inputted max_reservation_time dict"""
+    max_reservation = {}
+    while True:
+        max_reservation_input = input(
+            "Do you want to specify a queue to set"
+            + " max reservation time for? (y/n): "
```
plars, line 96: I just noticed this, because the + blended in with the other "+"s on the left :)
But you don't really need that + to concatenate the strings. Any string within parentheses is going to get automatically concatenated. This is similar to you how can handle things like line length limits by adding ()s and breaking a string across lines. It's a minor thing but let's go ahead and clean it up since we're also waiting for the other thing to land anyway.

plars (changes requested):
One minor thing in several places that I just noticed. it's worth cleaning up now while we wait on the other part to land.

server/src/tools/client_credentials_admin.py
```diff
@@ -87,6 +87,29 @@ def handle_allowed_queues_input() -> list:
...
+    while True:
+        max_reservation_input = input(
+            "Do you want to specify a queue to set",
+            " max reservation time for? (y/n): ",
+        )
+        if max_reservation_input == "y":
+            queue_name = input(
+                "Enter the name of the queue(* to apply max reservation",
+                " for all queues): ",
+            )
```
boukeas: I know it's a little pedantic but I'd prefer:
```
            queue_name = input(
                "Enter the name of the queue ",
                "(* to apply max reservation  for all queues): ",
            )
```

server/src/tools/client_credentials_admin.py
```diff
@@ -87,6 +87,29 @@ def handle_allowed_queues_input() -> list:
...
+            max_reservation[queue_name] = int(
+                input("Enter the maximum reservation time for this queue: ")
+            )
```
boukeas, line 105: You might want to `try-except` for a `ValueError` here, it wouldn't be nice to collapse because the user mistypes something that can't be converted to a number.

server/src/tools/client_credentials_admin.py
```diff
@@ -87,6 +87,29 @@ def handle_allowed_queues_input() -> list:
...
 
+def handle_max_reservation_input() -> list:
+    """Gets a user inputted max_reservation_time dict"""
+    max_reservation = {}
+    while True:
+        max_reservation_input = input(
+            "Do you want to specify a queue to set",
+            " max reservation time for? (y/n): ",
+        )
+        if max_reservation_input == "y":
```
boukeas, line 98: I would suggest going for `max_reservation_input.lower()` here. Just a little more resilience in case the user types `Y` instead of `y`. Same below for `n`.

boukeas (changes requested):
Fine overall, just a couple of minor requests here and there.

boukeas (approved):

PR #436, opened 2024-12-25
remove hostname field, we don't support this anymore

kiya956:
## Description

remove hostname field, we don't support this anymore

## Tests
provision x8high-pdk successfully

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/parser.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/parser.py
@@ -34,7 +34,6 @@ TESTPLAN_SCHEMA = {
                     "type": "array",
                     "items": {"$ref": "#/$defs/mail_format"},
                 },
-                "hostname": {"type": "string"},
             },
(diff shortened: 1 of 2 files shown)


plars (approved):
This particular device connector is an odd one because the docs for it are somewhere else. Otherwise I would ask that the docs be updated as part of this PR. But since that's possible, all I ask is to please make sure that the docs get updated in the appropriate place.

kiya956:
> This particular device connector is an odd one because the docs for it are somewhere else. Otherwise I would ask that the docs be updated as part of this PR. But since that's possible, all I ask is to please make sure that the docs get updated in the appropriate place.

H

kiya956:
> This particular device connector is an odd one because the docs for it are somewhere else. Otherwise I would ask that the docs be updated as part of this PR. But since that's possible, all I ask is to please make sure that the docs get updated in the appropriate place.

Hi Paul
    Thank you for your time, I know you are on vacation. 
I will update the doc in zapper repo. Really appreciate it.

PR #437, opened 2025-01-07
Unified agent host charm

plars:
## Description

We've actually been running this for a while now, but  have held off on landing it because the old version of the charms were running in the devices lab still. It looks like it should be safe to merge now though!

## Resolved issues

CERTTF-374

## Documentation

Updated READMEs and deployment instructions

## Web service API changes

N/A

## Tests

This is currently used in our labs and has been running smoothly for a while now.


plars:
@nancyc12 for the minfds thing, I'm looking at doing that as a separate commit if that's ok. There's some other fixes/cleanups I'd like to do first, but I figured we could land this one for now since it's well tested at this point. Also it won't revert any change to the supervisord.conf that you made even if you deployed it

plars:
@nancyc12 I also pushed a few more changes to fix some test issues

nancyc12 (approved):

PR #440, opened 2025-01-09
minor pyupdate improvements

plars:
- **Use unittest.mock instead of mock and remove deprecated import**
- **Use fstrings for server code**

## Description

Some minor cleanups which should definitely be safe. We *technically* still support python3.8 for the agent and device connectors, but I'm not sure we really need to since pretty much everything is on Jammy now, but I'll confirm with Taipei before going too far with that.

## Resolved issues

N/A

## Web service API changes

N/A

## Tests

Unit tests all still pass (and was even able to remove some deprecated imports during testing)

### DIFF
--- a/agent/testflinger_agent/tests/test_agent.py
+++ b/agent/testflinger_agent/tests/test_agent.py
@@ -6,10 +6,10 @@ import shutil
 import tarfile
 import tempfile
 import uuid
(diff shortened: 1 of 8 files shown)


omar-selo (approved):
+1

PR #441, opened 2025-01-10
update: Increase the timeout to 5400 seconds.

kevinyehk:
## Description

The current timeout for oemscript is 3600 seconds, however we have observed more and more machines hit the timeout, this might due to the slow network speeds or the machines still use spinning disk.
To reduce the failure rate for provisioning phase, we propose extending the timeout to 5400 seconds.


nancyc12 (approved):
LGTM

PR #448, opened 2025-01-23
Sort Order Improvements for the Agents table

val500:
## Description

Adds some improvements to the agents table - namely the provisioning streak column. This PR makes the sorting functionality of the table work with this column, first separating by failures/success, then sorting by the streak number. In addition, url parameters were added to persist sorting information on refresh for the agents table. Finally, this column was given the name `"Outcome",` and a tooltip was added describing this column.

To support this, new functions were added to filter.js - ordering functions were created to differentiate default ordering with provision streak ordering. Also, a new function to get URL parameters and use them to sort the table was also created. Finally, the sortTable function was split into two functions - sortTable and cycleTableSort. sortTable now just sorts a table given an ordering("ascending", "descending", "none"), while cycleTableSort cycles through the ordering states and uses sortTable to apply that ordering.

## Resolved issues
Resolves https://warthogs.atlassian.net/browse/CERTTF-396

## Documentation
Each new function was given documentation and comments.

## Web service API changes
N/A

## Tests
N/A

## Pictures
No sort applied - note the url
none_sort

Ascending Sort
ascending_sort

Descending Sort
descending_sort

New Tooltip for Column
Screenshot from 2025-01-28 12-19-50

### DIFF
--- a/server/src/templates/agents.html
+++ b/server/src/templates/agents.html
@@ -22,7 +22,7 @@
         <tr>
             <th aria-sort="none" class="has-overflow" style="width:10%" use-outcome-sort id="agentsOutcomeHeader">
                 <span class="p-tooltip--top-left" aria-describedby="tooltip">
-                    <span class="p-tooltip__message" role="tooltip" id="tooltip">The number of consecutively successful / failed provisioning attempts</span>
+                    <span class="p-tooltip__message" role="tooltip" id="tooltip" style="text-transform:none;font-weight:400">The number of consecutively successful / failed provisioning attempts</span>
                     Outcome
                 </span>
             </th>
--- a/server/src/templates/agents.html
+++ b/server/src/templates/agents.html
@@ -17,18 +17,18 @@
     <button type="reset" class="p-search-box__reset"><i class="p-icon--close"></i></button>
     <button type="submit" class="p-search-box__button"><i class="p-icon--search"></i></button>
 </form>
-<table aria-label="Agents table" class="p-table--mobile-card">
+<table aria-label="Agents table" class="p-table--mobile-card" id="agentsTable">
     <thead>
         <tr>
-            <th aria-sort="none" class="has-overflow" style="width: 80pt" use-outcome-sort>
+            <th aria-sort="none" class="has-overflow" style="width: 80pt" use-outcome-sort id="agentsOutcomeHeader">
 	        <span class="p-tooltip--top-left" aria-describedby="tooltip">
 		    <span class="p-tooltip__message" role="tooltip" id="tooltip">The number of consecutively successful / failed provisioning attempts</span>
 		    Outcome
 	        </span>
 	    </th>
-            <th aria-sort="none">Name</th>
-            <th aria-sort="none">State</th>
-            <th aria-sort="none">Updated At</th>
+            <th aria-sort="none" id="agentsNameHeader">Name</th>
+            <th aria-sort="none" id="agentsStateHeader">State</th>
+            <th aria-sort="none" id="agentsUpdatedAtHeader">Updated At</th>
             <th>Job ID</th>
         </tr>
     </thead>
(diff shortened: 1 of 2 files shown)


jocave:
Excited for this, would you mind including a screenshot(s) in this review so can quickly review UI changes? Thanks

val500:
> Excited for this, would you mind including a screenshot(s) in this review so can quickly review UI changes? Thanks

Attached new images

mz2:
I'll be annoying and request for a small tweak compared to the acceptance criteria I'd written earlier, now that I see it in action:

I would always start with the failures, and have ascending and descending rather secondarily sort by the numerical value in those groups. It's useful to have a way to highlight the low numbers quickly.

So, ascending:

```
1 fails
2 fails
3 fails
7 fails
1 successes
2 successes
5 successes
```

... and descending

```
7 fails
3 fails
3 fails
1 fails
5 successes
2 successes
1 successes
```

server/src/templates/agents.html
```diff
@@ -17,14 +17,18 @@ <h1 class="p-heading--3">
...
-<table aria-label="Agents table" class="p-table--mobile-card">
+<table aria-label="Agents table" class="p-table--mobile-card" id="agentsTable">
     <thead>
         <tr>
-            <th aria-sort="none" class="has-overflow" style="width: 40pt"></th>
-            <!-- Added an empty header for the warning icon column -->
-            <th aria-sort="none">Name</th>
-            <th aria-sort="none">State</th>
-            <th aria-sort="none">Updated At</th>
+            <th aria-sort="none" class="has-overflow" style="width: 80pt" use-outcome-sort id="agentsOutcomeHeader">
```
pedro-avalos, line 23: Hey, great work, I just have one comment in regards to the icons. You should add the `p-table__cell--icon-placeholder` on the `th` and `td` elements that have icons; this will make the icon display outside the column

See https://vanillaframework.io/docs/base/tables#icons

```suggestion
            <th aria-sort="none" class="p-table__cell--icon-placeholder has-overflow" style="width: 80pt" use-outcome-sort id="agentsOutcomeHeader">
```

pedro-avalos: The `td` element seems to be at line 38. I'm not sure how the icon placeholder behavior and the span and tooltip you have will interact with each other.

https://github.com/canonical/testflinger/blob/f8f32c38772361ad4c84328191120c1e3edfacd7/server/src/templates/agents.html#L38

pedro-avalos: I would also suggest you remove the `style="width: 80pt`, as it is generally bad practice to hard-code these styles into elements, and a hard-coded width can lead to some "fun" results...

image

server/src/templates/agents.html
```diff
@@ -38,14 +42,18 @@ <h1 class="p-heading--3">
...
                 {% elif agent.provision_streak_type == "pass" and agent.provision_streak_count > 0 %}
                 <span class="p-tooltip--right" aria-describedby="tooltip">
                     <i class="p-icon--success"></i>
                     <span class="p-tooltip__message" role="tooltip" id="tooltip">This agent has passed the last {{
                         agent.provision_streak_count }} provision attempts</span>
                 </span>
-                {{ agent.provision_streak_count }}
+		<span class="provision-streak">
+                  {{ agent.provision_streak_count }}
+		</span>
```
pedro-avalos, line 56: Could you fix the indentation here?

val500:
Updated images:

Descending Order:
image

Ascending Order:
image

Tooltip:
image

pedro-avalos (approved):
Looks good to me. Thanks for the changes!

server/src/templates/agents.html
pedro-avalos: Decided not to add the icons on the outside, as they are actually more informational/related to the column data, so they are fine to keep within the column.

PR #451, opened 2025-01-24
Add grafana integration

plars:
## Description

Add grafana integration so that (hopefully) it will push the prototype dashboard that shows mostly flask metrics right now.

This is based on the documentation at https://juju.is/docs/sdk/observe-your-charm-with-cos-lite#heading--integrate-with-grafana but I have not had the opportunity to test it yet. It seems to build ok, but that's all I know right now.

Aside from that, I've put this grafana template into a live instance, and it produces some useful data. The one exception that I didn't have time to sort out is the testflinger total jobs. I used a prometheus counter for this, but it can reset to 0 whenever the application restarts (due to upgrades, etc). Prometheus is supposed to have a way of automatically dealing with this, but I haven't sorted out how to make it work correctly so that the data is always monotonically increasing.

Hopefully someone else with much better grafana-fu than I can sort it out. :)


plars:
unsurprisingly, something more than what I see in the docs at https://juju.is/docs/sdk/observe-your-charm-with-cos-lite#heading--integrate-with-grafana seems to be needed, but I'm about out of time. Hopefully, at the very least, this will be a good starting point with the .json file which can be imported into a grafana dashboard and used as a starting point. Feel free to take whatever you find useful from this and close it whenever you see fit.

PR #452, opened 2025-02-04
Prevent token from expiring while waiting for agent

val500:
## Description
There have been some reported issues with tokens expiring between requesting them and submitting a job. This PR has several changes surrounding this issue

1. The Token Request was moved into the submit_job_data function to reduce the amount of time between requesting a token and submitting a job.
2.  The submit_job_data function attempts to retry job submission up to 3 times if the token becomes expired. 
3. Token expiration timeouts have been increased to 30 seconds in the server.
4. The server returns a 401 error on expiration instead of a 403.

## Resolved issues
https://warthogs.atlassian.net/browse/CERTTF-494

## Documentation
N/A

## Web service API changes
N/A

## Tests
Unit test was added to test_cli.py to test retry functionality.

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -574,43 +574,49 @@ class TestflingerCli:
 
     def submit_job_data(self, data: dict):
         """Submit data that was generated or read from a file as a test job"""
-        try:
-            auth_headers = self.build_auth_headers()
-            job_id = self.client.submit_job(data, headers=auth_headers)
-        except client.HTTPError as exc:
-            if exc.status == 400:
-                sys.exit(
-                    "The job you submitted contained bad data or "
-                    "bad formatting, or did not specify a "
-                    "job_queue."
-                )
-            if exc.status == 404:
-                sys.exit(
-                    "Received 404 error from server. Are you "
-                    "sure this is a testflinger server?"
-                )
-            if exc.status == 401:
-                sys.exit(
-                    "Received 401 error from server. You are "
-                    "attempting to use a feature that requires "
-                    "client authorisation without using client "
-                    "credentials. See https://testflinger.readthedocs"
-                    ".io/en/latest/how-to/authentication/ for more details"
-                )
-            if exc.status == 403:
-                sys.exit(
-                    "Received 403 error from server with reason "
-                    f"{exc.msg}"
-                    "The specified client credentials do "
-                    "not have sufficient permissions for the resource(s) "
-                    "you are trying to access."
-                )
-
-            # This shouldn't happen, so let's get more information
-            sys.exit(
-                "Unexpected error status from testflinger "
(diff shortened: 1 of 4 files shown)


cli/testflinger_cli/__init__.py
```diff
@@ -570,10 +572,11 @@ def check_online_agents_available(self, queue: str):
             "Waiting for agents to become available..."
         )
 
-    def submit_job_data(self, data: dict, headers: dict = None):
+    def submit_job_data(self, data: dict):
         """Submit data that was generated or read from a file as a test job"""
         try:
-            job_id = self.client.submit_job(data, headers=headers)
+            auth_headers = self.build_auth_headers()
+            job_id = self.client.submit_job(data, headers=auth_headers)
```
boukeas, line 579: Would it make sense to add some sort of retry logic here for the case where the token has expired?

server/src/api/v1.py
```diff
@@ -801,7 +801,7 @@ def generate_token(allowed_resources, secret_key):
     See retrieve_token for more information on the contents of
     the token payload
     """
-    expiration_time = datetime.utcnow() + timedelta(seconds=2)
+    expiration_time = datetime.utcnow() + timedelta(seconds=30)
```
boukeas, line 804: I am curious, why such a big increase?

cli/testflinger_cli/__init__.py
```diff
@@ -570,44 +572,51 @@ def check_online_agents_available(self, queue: str):
...
+                    )
+                if exc.status == 401:
+                    sys.exit(
+                        "Received 401 error from server. You are "
+                        "attempting to use a feature that requires "
+                        "client authorisation without using client "
+                        "credentials. See https://testflinger.readthedocs"
+                        ".io/en/latest/how-to/authentication/ for more details"
+                    )
+                if exc.status == 403:
```
boukeas, line 603: I read around a bit and it seems that the suggested status code for an expired token is 401, instead of a 403. Some sources do mention that a 403 is sometimes used but they don't recommend it. The reason why it makes sense to me (and why I researched it in the first place) is that a 403 is definitive and can't lead to a retry: you simply don't have the permissions. Whereas a 401 actually leaves room for a retry: you didn't include a token or the token has expired and in both cases you are welcome to take corrective action.

boukeas: I am conscious such a change isn't exactly in the scope of this PR but, if you agree, we can kill two birds with one stone.

val500: I can change this

server/src/api/v1.py
val500: I didn't want there to be a risk of a token expiration if the connection is slow. If we get some profiling data in the future, I can fine tune it.
It seems like tokens tend to have expirations on the scale of minutes in other applications.

boukeas (approved):

PR #455, opened 2025-02-05
RTD sphinx settings, don't fail on warning

jocave:
## Description

Change the sphinx section of the readthedocs settings to not fail on warnings which is currently causing build failures. Although we could/should fix the source of the warning, escalating these to errors seems unnecessary and not a setting that is applied in other projects maintained by Canonical

## Resolved issues

RTD docs build failures


Hook25 (approved):
Please do take a sec to fix the warning as well!

PR #458, opened 2025-02-06
[CERTTF-458] feat: use agent config in reimplemented log truncation

boukeas:
## Description

Log files are currently truncated to their last 1MB of data and that limit is hard-coded. This PR introduces a slight re-implementation of the way log files are truncated and allows the size of the truncated output log to be specified through the `output_bytes` agent configuration field.

The default for `output_bytes` when not explicitly set in the agent config is now 10MB.

## Resolved issues

Resolves [CERTTF-484](https://warthogs.atlassian.net/browse/CERTTF-484). Note that the original task was to remove truncation altogether but making the limit parametric was deemed to be safer and more flexible.

## Documentation

A paragraph was added to the section in the tutorial that discussed retrieving results. It is now explicitly stated that the results might be truncated and how the size of the truncated output log is set through the agent configuration.

## Tests

Tested locally with different values for `output_bytes` using (variations of) this Testflinger job, which outputs random 100-character lines of output:
```
job_queue: myqueue
test_data:
  test_cmds: |
    for i in {1..10000}; do
        # print a random 100-character line
        tr -dc 'A-Za-z0-9!"#$%&'\''()*+,-./:;<=>?@[\]^_`{|}~' < /dev/urandom | head -c 100
        echo
    done
```

[CERTTF-484]: https://warthogs.atlassian.net/browse/CERTTF-484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/testflinger_agent/job.py
+++ b/agent/testflinger_agent/job.py
@@ -16,7 +16,6 @@ import json
 import logging
 import os
 import time
-from typing import Optional
 
 from testflinger_agent.errors import TFServerError
 from .runner import CommandRunner, RunnerEvents
@@ -249,7 +248,7 @@ class TestflingerJob:
         yield "*" * (len(line) + 4)
 
 
-def read_truncated(filename: str, size: Optional[int]) -> str:
+def read_truncated(filename: str, size: int) -> str:
     """Return a string corresponding to the last bytes of a text file.
 
     Include a warning message at the end of the returned value if the file
--- a/agent/testflinger_agent/job.py
+++ b/agent/testflinger_agent/job.py
@@ -17,6 +17,7 @@ import json
 import logging
 import os
 import time
+from typing import Optional
 
 from testflinger_agent.errors import TFServerError
 from .runner import CommandRunner, RunnerEvents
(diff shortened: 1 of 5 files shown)


agent/testflinger_agent/job.py
```diff
@@ -215,22 +213,34 @@ def wait_for_completion(self):
                 logger.warning("Failed to get allocated job status, retrying")
             time.sleep(60)
 
-    def _set_truncate(self, f, size=1024 * 1024):
-        """Set up an open file so that we don't read more than a specified
-           size. We want to read from the end of the file rather than the
-           beginning. Write a warning at the end of the file if it was too big.
+    def _read_truncated(
+        self, filename: str, size: Optional[int] = None
```
val500, line 217: Is this size parameter ever used?

boukeas: It was used exclusively in the tests but your question made me realize that this could be further simplified: the configuration value for `output_bytes` should not be retrieved in the method but rather in the caller and `read_truncated` can just be a standalone function. I've pushed the relevant commit now.

agent/testflinger_agent/job.py
```diff
@@ -265,14 +249,24 @@ def banner(self, line):
         yield "*" * (len(line) + 4)
 
 
-def set_nonblock(fd):
-    """Set the specified fd to nonblocking output
+def read_truncated(filename: str, size: Optional[int]) -> str:
```
val500, line 252: Minor thing, signature says it takes an Optional, but the function doesn't handle the None case

val500 (approved):
LGTM

val500 (approved):
LGTM

PR #466, opened 2025-02-18
Add builtin account support for Classic image

kiya956:
## Description
Add builtin account support for Classic image
This change require another change
https://github.com/canonical/zapper/pull/453

## Resolved issues
we have some classic images have builtin account
need to handle this case


seankingyang:
Hi @kiya956,
This change will be very helpful for flashing classic image into IOT device.
Thank you~~

seankingyang (approved):
Hi @kiya956,
This change is very helpful for flashing classic image into IOT device.
Thank you ~~

Issue #465, opened 2025-02-17
[Zapper IOT] should not apply `ssh-copy-id` for all zapper iot provision method


seankingyang:
If using the zapper iot provision image (Ubuntu Core) to machine and using the console-conf to setting the profile.
We should use the `ce-certification-qa` email, once machine setting is done, the ssh key already in the DUT.
The agent already have the access to the DUT.

https://github.com/canonical/testflinger/blob/main/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py#L79-L82
```
    def _post_run_actions(self, args):
        """Run further actions after Zapper API returns successfully."""

        self._copy_ssh_id()
```

If using the console-conf, there is no local password in it.
If still run `self._copy_ssh_id()`, the `sshpass -p <test_password> ssh-copy-id ce-certification-qa@<DUT IP>` will be run, and the return code will be non-zero.

```
$ sshpass -p <test_password> ssh-copy-id ce-certification-qa@<DUT IP>
$ echo $?
> 6
```

PR #464, opened 2025-02-17
Remvoe method from deploy in schema check.

kiya956:
We dont support method in deploy anymore

## Description

Remvoe method from deploy in schema check.
We dont support method in deploy anymore

## Documentation

The Doc is in zapper repo

## Tests
provision x8high-pdk with uuu and seed_override successfully

PR #472, opened 2025-02-25
[Zapper-IOT] Not do the copy_ssh_id when using the console-conf

seankingyang:
## Description
This change is based on https://github.com/canonical/zapper/pull/450 and https://github.com/canonical/testflinger/pull/464 to do the initial login method check.

Add the `ubuntu_email_sso` key in provision_data, and it will overwrite the `test_username`.

```yaml
provision_data:
  # If provide the ubuntu_email_sso, it will overwrite the username which is used in the provisioning.
  ubuntu_email_sso: < ubuntu email sso >
  urls:
    - < URL 1 >
    - < URL 2 >
  provision_plan:
    config:
      project_name: < project name >
      serial_console:
        port: < port >
        baud_rate: < baud rate >
      network: <Eth interface name>
      hostname:  < Host name >
    run_stage:
      - deploy:
          utility: < utility >
          utility_extra_args: <utility extra args>
       - initial_login:
           method: < console-conf | builtin-account >
test_data: 
  # if using console-conf, we can leave it 
  test_username: <ubuntu | {or others}> 
  test_password: <ubuntu | {or others}>
```

## Resolved issues
#465 [Zapper IOT] should not apply ssh-copy-id for all zapper iot provision method

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
@@ -45,6 +45,10 @@ class DeviceConnector(ZapperConnector):
             "test_password", "ubuntu"
         )
 
+        ubuntu_sso_email = self.job_data["provision_data"].get(
+            "ubuntu_sso_email", ""
+        )
+
         provisioning_data = {
             "username": username,
             "password": password,
@@ -56,16 +60,44 @@ class DeviceConnector(ZapperConnector):
         if provision_plan:
 
             try:
-                validate_provision_plan(provision_plan)
-
                 # Make sure the user created at provision time is
                 # the same used during the test phase.
                 provision_plan["config"]["username"] = username
                 provision_plan["config"]["password"] = password
 
+                # If the initial login is using console-conf, we need to
+                # check if the user has provided an ubuntu_sso_email.
+                # If not, we raise an error.
+                # ubuntu_sso_email will be used to set the username for the
+                # initial login.
+                run_stages = provision_plan["run_stage"]
+                for stage in run_stages:
+                    if (
+                        isinstance(stage, dict)
+                        and "initial_login" in stage.keys()
+                        and stage["initial_login"].get("method")
+                        == "console-conf"
+                    ):
+                        if ubuntu_sso_email:
+                            provision_plan["config"][
+                                "username"
+                            ] = ubuntu_sso_email
+                            break
+                        else:
+                            raise ValueError(
+                                "ubuntu_sso_email is required "
+                                "when initial login using console-conf"
+                            )
+
+                validate_provision_plan(provision_plan)
+
                 provisioning_data["custom_provision_plan"] = provision_plan
-            except ValueError as e:
+            except (ValueError, KeyError) as e:
                 raise ProvisioningError from e
+        # If there is an ubuntu_sso_email and no provision plan, we use it
+        # as the username.
+        elif ubuntu_sso_email:
+            provisioning_data["username"] = ubuntu_sso_email
 
         urls = self.job_data["provision_data"].get("urls", [])
         try:
@@ -80,8 +112,12 @@ class DeviceConnector(ZapperConnector):
         """Run further actions after Zapper API returns successfully."""
         super()._post_run_actions(args)
 
+        # Default to do copy the ssh id.
         do_copy_ssh_id = True
+
         provision_plan = self.job_data["provision_data"].get("provision_plan")
+        # If provision plan is provided, we check if the initial login
+        # is using console-conf. If so, we do not copy the ssh id.
         if provision_plan:
             run_stages = provision_plan["run_stage"]
             for stage in run_stages:
@@ -91,6 +127,15 @@ class DeviceConnector(ZapperConnector):
                     and stage["initial_login"].get("method") == "console-conf"
                 ):
                     do_copy_ssh_id = False
+                    break
+        # If no provision plan is provided, we check if the user has
+        # provided an ubuntu_sso_email. If so, we do not copy the ssh id.
+        else:
+            ubuntu_sso_email = self.job_data["provision_data"].get(
+                "ubuntu_sso_email", ""
+            )
+            if ubuntu_sso_email:
+                do_copy_ssh_id = False
 
         if do_copy_ssh_id:
             self._copy_ssh_id()
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
@@ -14,6 +14,7 @@
 
 """Zapper Connector for IOT provisioning."""
 import logging
+import contextlib
 from typing import Any, Dict, Tuple
 from testflinger_device_connectors.devices.zapper import ZapperConnector
 from testflinger_device_connectors.devices import ProvisioningError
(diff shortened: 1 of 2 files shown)


kiya956 (dismissed):
LGTM

p-gentili:
I need more details on the problem: can you share an example job yaml? How would the device be accessible from TF if you don't copy the SSH keys? Remember that provisioning is only one of the phases of a testflinger job.

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
p-gentili, line 1: I don't think this is the right condition to check since you're making the assumption we're using a custom provision plan and not a preset.

p-gentili (changes requested):
I think we're trying to solve the symptom and not the issue: when using console-conf one thing is the email, another thing is the username. For instance, `ce-certification-qa@lists.canonical.com` is the Ubuntu SSO account and `ce-certification-qa` is the associated username (and it's just a coincidence that the email base matches the username).

I'd recommend to introduce another optional argument to the job, in a similar way as `zapper_kvm`: `ubuntu_sso_email`.
```
# job.yaml
provision_data:
    provision_plan: <...>
    urls: []
    ubuntu_sso_email: ce-certification-qa@lists.canonical.com
test_data:
    username: <...>
    password: <...>
```

when this is the case, you can call the zapper provision API with:
```
"username": self.job_data["provision_data"].get("ubuntu_sso_email") or username
```
this would make the Zapper side happy, correct?

and in `_post_run_action`:
```
if not self.job_data["provision_data"].get("ubuntu_sso_email"):
    self._copy_ssh_id()
```

I'll be happy to pair program or discuss about this face to face if needed.

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -45,6 +45,10 @@ def _validate_configuration(
             "test_password", "ubuntu"
         )
 
+        ubuntu_sso_email = self.job_data["provision_data"].get(
+            "ubuntu_sso_email", ""
```
p-gentili, line 49: No need to provide an empty string as default, `.get()` already gives with `None` in that case.

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -45,6 +45,10 @@ def _validate_configuration(
...
+        )
+
         provisioning_data = {
             "username": username,
```
p-gentili, line 53: What about we already assign it here?
```suggestion
            "username": username if not ubuntu_sso_email else ubuntu_sso_email,
```

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -56,16 +60,44 @@ def _validate_configuration(
         if provision_plan:
 
             try:
-                validate_provision_plan(provision_plan)
-
                 # Make sure the user created at provision time is
                 # the same used during the test phase.
                 provision_plan["config"]["username"] = username
                 provision_plan["config"]["password"] = password
```
p-gentili, line 66: And if you liked my suggestion above, you should also apply this change here.
```suggestion
                provision_plan["config"]["username"] = provision_data["username"]
                provision_plan["config"]["password"] = provision_data["password"]
```

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -56,16 +60,44 @@ def _validate_configuration(
...
+                        if ubuntu_sso_email:
+                            provision_plan["config"][
+                                "username"
+                            ] = ubuntu_sso_email
+                            break
+                        else:
+                            raise ValueError(
+                                "ubuntu_sso_email is required "
+                                "when initial login using console-conf"
+                            )
```
p-gentili, line 90: In cases like this, I personally enjoy using `contextlib.suppress` which will let us iterate w/o bothering with exceptions until we find exactly what we need.
```suggestion
                    with contextlib.suppress(KeyError):
                        if stage["initial_config"]["method"] == "console-conf":
                            if not ubuntu_sso_email:
                                raise ValueError(
                                    "ubuntu_sso_email is required "
                                    "when initial login using console-conf"
                                )
                            provision_plan["config"][
                                "username"
                            ] = ubuntu_sso_email    
                            break
```
and you'll need to `import contextlib` on top.

( I hope I used the correct indentation, the GitHub editor is a PITA with code )

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -56,16 +60,44 @@ def _validate_configuration(
...
+                validate_provision_plan(provision_plan)
+
                 provisioning_data["custom_provision_plan"] = provision_plan
-            except ValueError as e:
+            except (ValueError, KeyError) as e:
                 raise ProvisioningError from e
+        # If there is an ubuntu_sso_email and no provision plan, we use it
+        # as the username.
+        elif ubuntu_sso_email:
+            provisioning_data["username"] = ubuntu_sso_email
```
p-gentili, line 100: And again, if you enjoyed my suggestions above, this would not be needed anymore.
```suggestion
```

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -78,5 +110,32 @@ def _validate_configuration(
...
+        # provided an ubuntu_sso_email. If so, we do not copy the ssh id.
+        else:
+            ubuntu_sso_email = self.job_data["provision_data"].get(
+                "ubuntu_sso_email", ""
+            )
+            if ubuntu_sso_email:
+                do_copy_ssh_id = False
+
+        if do_copy_ssh_id:
+            self._copy_ssh_id()
```
p-gentili, line 141: Since this is the final step, we can assume everything else has already been checked and completed successfully. I honestly would just:
```
if not self.job_data["provision_data"].get("ubuntu_sso_email"):
    self._copy_ssh_id()
```

p-gentili (commented):
We're going in the right direction! I left a bunch of comments and suggestions below, feel free to ping me directly in case something is not clear.

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
seankingyang: Got it, thanks~

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
seankingyang: Agree, it will be clearer.

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
seankingyang: Thanks for double checking others I need to modify.

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
seankingyang: Yes, it can make the code clearer and readable.

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
seankingyang: Agree, we can assume that and it will be readable.

p-gentili (approved):
LGTM! Thanks for all the changes.

PR #478, opened 2025-02-26
Make agent-host ssh variables sensitive

pedro-avalos:
## Description

- Make SSH public and private keys sensitive variables.
  - Technically the public one isn't "sensitive", but it makes the output of `terraform plan` less verbose


omar-selo (approved):
Looks good

PR #481, opened 2025-03-03
Minor fix in the agent table for absent provision-streak items

val500:
## Description
This PR has a minor fix in the agent table so the sort behavior of the outcome column works when a provision streak element is not present.


pedro-avalos (approved):
+1

PR #486, opened 2025-03-05
Add "Circle of Friends" favicon

pedro-avalos:
## Description

Adds the Ubuntu Circle of Friends as the favicon for Testflinger.


plars (approved):
I think we used to have the favicon but lost it somehow during the vanilla conversion and ended up just turning it off so that we didn't generate 404s with the thought that maybe we might have a "testflinger" favicon at some point. But this is better than not having one at all!
+1 from me but I don't think it counts now. Hi anyway :)

pedro-avalos:
Hi and thanks Paul! We all miss you over here :(

val500 (approved):
LGTM

PR #487, opened 2025-03-06
Remove provision plan validation from device connector

kiya956:
## Description
Remove provision plan validation from device connector
We move it to zapper side

## Resolved issues

We don't have to maintain both side

## Tests

provision x8high-pdk successfully


p-gentili (approved):
The service is more up-to-date with changes to the job schema, and it already runs a similar check. Makes sense to have the validation only there.

PR #491, opened 2025-03-07
413 too large POST `v1/result` requests

mz2:
## Description

The production testflinger API server appears to be hit with too large requests to the `v1/result` endpoint. Looking at the implementation, no input validation is done to avoid sending problematically large data to MongoDB by this endpoint. This can be done cheaply by checking the content size header (this will prevent the issue at least when the, well behaved `requests` client library using testflinger-cli is the source of the request).

## Resolved issues

No issue tracked yet, incident going on.

## Documentation

None

## Web service API changes

413 in case of `v1/results` receiving too large an input

## Tests

Integration test added to the server implementation


andrejvelichkovski:
I suggest we don't remove pip in this PR, in hotfix I think it would make sense to just keep the main fix only

mz2:
I suggest we do, since we run tests in CI and see that the tests pass.

server/src/api/v1.py
```diff
@@ -255,6 +255,10 @@ def result_post(job_id, json_data):
     if not check_valid_uuid(job_id):
         abort(400, message="Invalid job_id specified")
 
+    content_length = request.content_length
+    if content_length and content_length >= 16 * 1024 * 1024:
```
jocave, line 259: I think it should be explained where this limit comes from i.e. it's related to some mongodb restriction

jocave (changes requested):

server/src/api/v1.py
mz2: Good one, sorted.

mz2:
Actually, the tox.ini change was nonsense, I did remove that (it was the wrong tox.ini!) -- highlights that we don't actually run tox for the CLI at all in CI.

mz2:
I based this branch on an ancient commit from a local working copy, and got in rebasing hell --> superseded by a https://github.com/canonical/testflinger/compare/413-for-too-large-post-result-input

PR #492, opened 2025-03-07
Fail POST v1/result with a 413 if the input is too large

mz2:
## Description

The production testflinger API server appears to be hit with too large requests to the `v1/result` endpoint. Looking at the implementation, no input validation is done to avoid sending problematically large data to MongoDB by this endpoint. This can be done cheaply by checking the content size header (this will prevent the issue at least when the, well behaved `requests` client library using testflinger-cli is the source of the request).

## Resolved issues

No issue tracked yet, incident going on.

## Documentation

None

## Web service API changes

413 in case of `v1/results` receiving too large an input, which should stop the agent from retrying requests (will only retry 50x)

## Tests

Integration test added to the server implementation


jocave (approved):
Comment addressed from previous PR, thanks

boukeas (approved):

PR #495, opened 2025-03-07
feat: use unique identifier as a header when pinging testflinger

boukeas:
## Description

The `submit` action includes a step that hits the Testflinger API `/jobs` endpoint in order to test if Testflinger is accesible. This PR switches to simply hitting the `/v1` root but, most importantly, includes a unique request ID in the header (the run ID of the Github workflow) so that the request is traceable in the Testflinger ingress logs.

## Resolved issues

Not tracked, a simple refinement.

## Documentation

N/A

## Tests

This is a [workflow run](https://github.com/canonical/certification-lab-ci/actions/runs/13724870134) with run ID `13724870134` that uses the `submit` action from this branch. And this is how the corresponding request appears in the ingress log (with the run ID appearing right at the end of the log entry):
```
10.141.84.170 - - [07/Mar/2025:16:24:10 +0000] "HEAD /v1 HTTP/2.0" 403 0 "-" "curl/7.81.0" 66 0.000 [testflinger-production-relation-12-testflinger-service-5000] [] - - - - 13724870134
```

### DIFF
--- a/.github/actions/submit/action.yaml
+++ b/.github/actions/submit/action.yaml
@@ -36,9 +36,10 @@ runs:
       shell: bash {0}  # allow curl to fail
       env:
         SERVER: https://${{ inputs.server }}
+        REQUEST_ID: ${{ github.run_id }}
       run: |
         echo "::group::Test connection to Testflinger server"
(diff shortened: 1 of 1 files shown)


jocave (approved):
Looks like a helpful use of the header to me +1

val500 (approved):
LGTM

PR #497, opened 2025-03-07
Extract agents table

pedro-avalos:
## Description

The agents table is duplicated between the agents page and the queue detail page. This PR extracts that template and inserts it into these two pages.

## Screenshots

The updated queue detail page with the new table
image


val500 (approved):
Much cleaner now, nice!

PR #500, opened 2025-03-07
Fix tab title

pedro-avalos:
## Description

The `title` block was not used before. This PR adds a page title for each page we serve.

## Screenshots

Before:
image

After:
image


val500 (approved):
LGTM

PR #505, opened 2025-03-10
Use versioned GitHub actions

pedro-avalos:
## Description

- There were some actions that were pinned to a specific commit; we should instead pin them to a versioned release/tag.

### DIFF
--- a/.github/workflows/device-tox.yml
+++ b/.github/workflows/device-tox.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         python: ["3.8", "3.10"]
     steps:
-    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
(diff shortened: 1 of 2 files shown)


boukeas (commented):
Pedro, I checked the git blame for these lines and it seems that they are being changed automatically by dependabot. So I definitely agree with the spirit of this PR but we need to make sure that the versions don't get changed back by dependabot (or that dependabot also uses version tags instead of commit hashes?)

pedro-avalos:
Hm, well we use the versioned tags elsewhere, so I'd assume dependabot would be fine with these changes

boukeas (approved):
My concern is that the renovate bot might create PRs to override this one using commit hashes once again. However, I am approving this because Pedro hopes that switching to major versions will lead the renovate bot to remain consistent and also only update on major versions.

PR #509, opened 2025-03-10
Add data headings to cards

pedro-avalos:
## Description

- This adds the heading to the mobile cards when the screen is too small for the full table.
- Fix cases where we had a row header, but we didn't use the `<th scope="row">...</th>` element

## Screenshots

For example:
image


val500 (approved):
LGTM

PR #514, opened 2025-03-12
fix(cli): Fix stdin filename argument for submit

pedro-avalos:
## Description

Follow-up from #503.

It seems `-` is a valid argument for `testflinger-cli submit -`. This PR changes the argument type and adds some more error-handling to the `submit` function.

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -363,10 +363,12 @@ class TestflingerCli:
         parser.add_argument("--poll", "-p", action="store_true")
         parser.add_argument("--quiet", "-q", action="store_true")
         parser.add_argument("--wait-for-available-agents", action="store_true")
-        parser.add_argument("filename", type=Path).completer = (
-            argcomplete.completers.FilesCompleter(
-                allowednames=("*.yaml", "*.yml", "*.json")
-            )
+        parser.add_argument(
+            "filename",
+            type=str,
+            help="YAML or JSON file with your job definition, '-' for stdin",
+        ).completer = argcomplete.completers.FilesCompleter(
+            allowednames=("*.yaml", "*.yml", "*.json")
         )
         parser.add_argument(
             "--client_id",
@@ -521,13 +523,13 @@ class TestflingerCli:
         if self.args.filename == "-":
             data = sys.stdin.read()
         else:
+            path = Path(self.args.filename)
             try:
-                with open(
-                    self.args.filename, encoding="utf-8", errors="ignore"
-                ) as job_file:
-                    data = job_file.read()
+                data = path.read_text(encoding="utf-8", errors="ignore")
+            except PermissionError:
+                sys.exit(f"File not readable: {path}")
             except FileNotFoundError:
-                sys.exit(f"File not found: {self.args.filename}")
+                sys.exit(f"File not found: {path}")
         job_dict = yaml.safe_load(data)
 
         # Check if agents are available to handle this queue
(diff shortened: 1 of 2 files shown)


cli/testflinger_cli/__init__.py
```diff
@@ -521,13 +523,13 @@ def submit(self):
         if self.args.filename == "-":
             data = sys.stdin.read()
         else:
+            path = Path(self.args.filename)
```
boukeas, line 526: It doesn't make much difference but I would suggest naming this `file` or `filename` instead of `path`. Up to you, I just personally would find that more readable.

pedro-avalos: Done

cli/testflinger_cli/__init__.py
```diff
@@ -521,13 +523,13 @@ def submit(self):
...
-                with open(
-                    self.args.filename, encoding="utf-8", errors="ignore"
-                ) as job_file:
-                    data = job_file.read()
+                data = path.read_text(encoding="utf-8", errors="ignore")
+            except PermissionError:
+                sys.exit(f"File not readable: {path}")
             except FileNotFoundError:
-                sys.exit(f"File not found: {self.args.filename}")
+                sys.exit(f"File not found: {path}")
```
boukeas: I know it's "always" been like this but I am wondering what the point of displaying a custom error message when we could simply use the exception itself, e.g.:
```
except FileNotFoundError as error:
  sys.exit(f"{error}")
```
Or `sys.exit(f"{type(error).__name__}: {error}").
Same above with the permission error.

cli/testflinger_cli/__init__.py
```diff
@@ -363,10 +363,12 @@ def _add_submit_args(self, subparsers):
         parser.add_argument("--poll", "-p", action="store_true")
         parser.add_argument("--quiet", "-q", action="store_true")
         parser.add_argument("--wait-for-available-agents", action="store_true")
-        parser.add_argument("filename", type=Path).completer = (
-            argcomplete.completers.FilesCompleter(
-                allowednames=("*.yaml", "*.yml", "*.json")
-            )
+        parser.add_argument(
+            "filename",
+            type=str,
```
boukeas, line 368: I _really_ don't mind either way, but I am wondering if it's worth leaving this as `type=Path` (to capture the essence of what we expect this argument to be, at least for 99.99% of cases) and check for `-` below. Something like:
```
if self.args.filename.name == "-":
```
instead of:
```
if self.args.filename == "-":
```
Again, not a major thing, up to you.

pedro-avalos: Hm, that is a good solution. I agree, it is almost always a path (I didn't even know we had the stdin option)

cli/testflinger_cli/__init__.py
pedro-avalos: I've simplified it using the `logger` and `sys.exit`

boukeas (approved):

cli/testflinger_cli/__init__.py
```diff
@@ -518,16 +520,15 @@ def build_auth_headers(self):
 
     def submit(self):
         """Submit a new test job to the server"""
-        if self.args.filename == "-":
+        if self.args.filename.name == "-":
             data = sys.stdin.read()
         else:
+            filename = Path(self.args.filename)
```
boukeas, line 526: You don't need this now, do you?

pedro-avalos: Doh, right

boukeas (approved):

PR #516, opened 2025-03-13
Update tf-cleanup to use environment variables provided by the charm

val500:
## Description
This PR uses environment variables provided the agent charm to pass in the agent configuration to the device connector. This change was supposed to be included in #405, but it was seemingly accidentally dropped in #437. This prevents cleanup from knowing the provision type which means MAAS provisioned devices aren't being released during the cleanup phase.

This also includes a change to log errors with the cleanup phase in the agent.

## Resolved issues
Resolves https://warthogs.atlassian.net/browse/CERTTF-518


pedro-avalos (approved):
Look good to me

PR #517, opened 2025-03-13
fix(server): Fix filter disappearing when pressing enter

pedro-avalos:
## Description

This PR ensures that a search/filter is not lost when pressing enter on the search box. It also ensures that the filters are reset when the search box is reset.


val500 (approved):
LGTM

PR #521, opened 2025-03-14
fix(server): Add reserve_data schema in server

pedro-avalos:

### DIFF
--- a/server/src/api/schemas.py
+++ b/server/src/api/schemas.py
@@ -18,8 +18,7 @@ Testflinger v1 OpenAPI schemas
 """
 
 from apiflask import Schema, fields
-from apiflask.validators import OneOf
-
+from apiflask.validators import OneOf, Regexp
 
 ValidJobStates = (
     "setup",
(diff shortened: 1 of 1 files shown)


server/tests/test_v1.py
```diff
@@ -33,6 +33,17 @@ def test_home(mongo_app):
     assert "/agents" == response.headers.get("Location")
 
 
+def test_add_job_invalid_reserve_data(mongo_app):
```
boukeas, line 36: Since the scope of this PR is to introduce a nested schema for `reserve_data` I would suggest introducing tests for the full scope. Some examples would include ssh keys that are strings but don't match the regexp or requests that include additional fields (are they acceptable?). You can group some of these tests together, e.g. you are testing cases of malformed inputs. And I would say you definitely need a test for valid requests. How else will you make sure that a future PR doesn't break something?

boukeas: One last (type of) test I think would be appropriate is one where multiple keys are provided. Both in the case of valid keys (e.g. one of the valid keys could be a `gh:...` key) and invalid keys (e.g. `{'ssh_keys': ['lp:me', {'lp': 'alsome'}]}`)

pedro-avalos: Good idea, I've added the tests now

boukeas (approved):
Smashing Pedro, thank you.

PR #525, opened 2025-03-18
chore(agent): Use operator-libs-linux passwd in agent-host charm

pedro-avalos:
## Description

- We already have `passwd` as a `charm-libs` dependency and within the repository, so we should use it over our check_output wrapper


nancyc12 (changes requested):
Thanks for the changes! Can you resolve the conflict?

rene-oromtz (approved):

pedro-avalos:
@nancyc12 I'm trying to clean up some open PRs I have; could you review this when you get a chance? It seems even though Rene approved it, it still is blocked on your re-review

nancyc12 (approved):

PR #527, opened 2025-03-20
Bugfix to client in agent to handle error status codes from the server

### LINKED ISSUE #438: Agents show EXITED in supervisorctl
## Issue
Some agents in TEL agent hosts were found to be `EXITED` and could not recover themselves.  
```
ubuntu@juju-462031-0:~$ sudo supervisorctl status | grep EXITED
hp-eliteone800-g6-all-in-one-pc-c28167                        EXITED    Jan 06 11:01 AM
rpi3bp009                                                     EXITED    Jan 06 11:12 AM
rpi4b1g001                                                    EXITED    Jan 06 11:09 AM
rpi4b2g001                                                    EXITED    Jan 06 02:30 PM
rpi4b2g002                                                    EXITED    Jan 06 11:06 AM
rpi4b4g001                                                    EXITED    Jan 06 02:29 PM
rpi4b4g002                                                    EXITED    Jan 06 11:04 AM
rpi4b8g001                                                    EXITED    Jan 06 02:22 PM
rpi4b8g002                                                    EXITED    Jan 06 11:08 AM

ubuntu@juju-47ff0e-1:~$ sudo supervisorctl status | grep EXITED
rpi400001                                                      EXITED    Jan 06 02:32 PM

ubuntu@juju-ccf3c0-0:~$ sudo supervisorctl status | grep EXITED
cm3p001                                                        EXITED    Jan 06 11:21 AM
dell-precision7550-c27818                                      EXITED    Jan 09 08:29 AM
rpi400002                                                      EXITED    Jan 06 11:09 AM
```

## Log
```
ubuntu@juju-462031-0:~$ sudo supervisorctl tail -f rpi4b8g002
==> Press Ctrl-C to exit <==
68c75fc38c/events (Caused by ResponseError('too many 503 error responses'))
[25-01-06 11:08:04]   ERROR: (cmd.py:32)| 'NoneType' object has no attribute 'status_code'
  File "/srv/testflinger-venv/lib/python3.10/site-packages/testflinger_agent/cmd.py", line 27, in main
    start_agent()
  File "/srv/testflinger-venv/lib/python3.10/site-packages/testflinger_agent/__init__.py", line 139, in start_agent
    agent.process_jobs()
  File "/srv/testflinger-venv/lib/python3.10/site-packages/testflinger_agent/agent.py", line 328, in process_jobs
    event_emitter.emit_event(TestEvent.CLEANUP_START)
  File "/srv/testflinger-venv/lib/python3.10/site-packages/testflinger_agent/event_emitter.py", line 55, in emit_event
    self.client.post_status_update(
  File "/srv/testflinger-venv/lib/python3.10/site-packages/testflinger_agent/client.py", line 448, in post_status_update
    % (status_update_uri, job_request.status_code)
AttributeError: 'NoneType' object has no attribute 'status_code
```

## Workaround
Manually start each agent.
```
sudo supervisorctl start rpi4b8g002
```

val500:
## Description
client.py inside of the agent doesn't handle error codes from the server correctly leading to potential agent crashes. This PR addresses those issues by explicitly checking for HTTP error codes.

## Resolved issues
Resolves #438

### DIFF
--- a/agent/testflinger_agent/client.py
+++ b/agent/testflinger_agent/client.py
@@ -117,7 +117,7 @@ class TestflingerClient:
         """
         uri = urljoin(self.server, f"/v1/job/{job_id}/attachments")
         with requests.get(uri, stream=True, timeout=600) as response:
-            if response.status_code != 200:
+            if not response:
                 logger.error(
                     f"Unable to retrieve attachments for job {job_id} "
                     f"(error: {response.status_code})"
@@ -439,11 +439,16 @@ class TestflingerClient:
             job_request = self.session.post(
                 status_update_uri, json=status_update_request, timeout=30
             )
+            # Response code is greater than 399
+            if not job_request:
+                logger.error(
+                    "Unable to post status updates to: %s (error: %s)",
+                    status_update_uri,
+                    job_request.status_code,
+                )
         except RequestException as exc:
-            logger.error("Server Error: %s" % exc)
-            job_request = None
-        if not job_request:
             logger.error(
-                "Unable to post status updates to: %s (error: %s)"
-                % (status_update_uri, job_request.status_code)
+                "Unable to post status updates to: %s (error: %s)",
+                status_update_uri,
+                exc,
             )
--- a/agent/testflinger_agent/tests/test_client.py
+++ b/agent/testflinger_agent/tests/test_client.py
@@ -20,6 +20,7 @@ from unittest.mock import patch
 import requests_mock as rmock
 
 from testflinger_agent.client import TestflingerClient as _TestflingerClient
+from testflinger_agent.errors import TFServerError
 
 
 class TestClient:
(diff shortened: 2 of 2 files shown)


agent/testflinger_agent/client.py
```diff
@@ -441,8 +441,7 @@ def post_status_update(
             )
         except RequestException as exc:
             logger.error("Server Error: %s" % exc)
-            job_request = None
-        if not job_request:
+        if job_request and job_request.status_code != 200:
```
boukeas, line 444: You can avoid this `if job_request` part here if:
- you include the check for the status code in the `try` block
- you use `try-except-else`
I think I'd go for the first option

agent/testflinger_agent/client.py
```diff
@@ -153,7 +153,7 @@ def repost_job(self, job_data):
         except RequestException as exc:
             logger.error(exc)
             raise TFServerError("other exception") from exc
-        if not job_request:
+        if job_request != 200:
```
boukeas, line 156: And this is why we need tests!

agent/testflinger_agent/client.py
```diff
@@ -441,8 +441,7 @@ def post_status_update(
             )
         except RequestException as exc:
             logger.error("Server Error: %s" % exc)
```
boukeas, line 443: I think I would like the message in this case to also be more informative, and more in line with the message we display if we actually do get a response. For example:
```
logger.error(
    "Unable to post status updates to: %s (error: %s)"
    % (status_update_uri, exc)
)
```
Also, I think we can write this in a more modern way:
```
logger.error(
    "Unable to post status updates to: %s (error: %s)",
    status_update_uri, exc
)
```

boukeas (changes requested):
I have suggested a couple of modifications. If possible, I'd also like you to go over the other methods in the client and make sure they are in line with the changes. ~~For example, in `save_artifacts` the response's status code is not checked. In `post_live_output` as well, the response is converted to a boolean and that will always be `True`.~~ Update: that's not the case actually, `Response` implements `__bool__` [[ref](https://github.com/psf/requests/blob/1764cc938efc3cc9720188dfa6c3852c45211aa0/src/requests/models.py#L730)] so it's valid to check the truth value of the response.

agent/testflinger_agent/client.py
val500: is exc the full job_request object?

boukeas (approved):
Excellent, thank you for fixing this.

PR #531, opened 2025-03-24
Added logging to node read function in MaasStorage

val500:
## Description
There seems to be some situations in which MAAS provisioning fails without logging anything. The only MAAS command that is run before any logging seems to be the `_node_read`  function inside of `maas_storage.py`. This PR adds error logging to that function to hopefully catch some more errors.


pedro-avalos (approved):
Looks good

PR #537, opened 2025-03-31
chore(cli): Add uv to cli

pedro-avalos:
## Description

- Adds [uv](https://docs.astral.sh/uv/) as the package manager for the CLI package

## Resolved issues

- Resolves [CERTTF-560](https://warthogs.atlassian.net/browse/CERTTF-560)

## Tests

* Tox tests should pass with new changes.
* Built the package with `snapcraft` locally and tested


.github/workflows/cli-tox.yml
```diff
@@ -2,11 +2,11 @@ name: "[cli] Run unit tests"
 
 on:
   push:
-    branches: [ main, try-self-hosted ]
+    branches: [main]
```
boukeas, line 5: I can see you're only keeping what's already there but would you mind checking what the standard practice is on the rest of our repos? We don't push to `main` directly but rather only after a PR so I don't see the point of this rule.

cli/pyproject.toml
```diff
@@ -6,14 +6,30 @@ build-backend = "setuptools.build_meta"
...
+  "black>=25.1.0",
+  "flake8>=7.2.0",
+  "pylint>=3.3.6",
+  "pytest>=8.3.5",
+  "pytest-cov>=6.0.0",
+  "pytest-mock>=3.14.0",
+  "requests-mock>=1.12.1",
+]
+
 [tool.black]
```
boukeas, line 33: Are the `[tool.black]` and `[tool.pylint.*]` sections actually used anymore? I really don't know how this works with `uv` but all this used to be taken care of through `tox.ini` which is now gone so I wonder if, for example, black or pylint might actually be configured in the `tools.tox.env.format` and `tools.tox.env.lint` sections. Otherwise I don't really now how the connection is made with `[tool.black]` and `[tool.pylint.*]`

.github/workflows/cli-tox.yml
pedro-avalos: @boukeas The standard is to use the `on.push` for these cases, there isn't a "after merge" rule. A merge is considered pushing to `main`.

pedro-avalos: As an example, Checkbox runs their tox tests on PR and push as well: https://github.com/canonical/checkbox/blob/main/.github/workflows/tox-checkbox-ng.yaml

cli/pyproject.toml
pedro-avalos: These sections are still used. If you were to run `black` or `pylint` from `cli/`, these tools would pull in the configuration from their respective `[tool]` subtable. `tox` is yet another "tool" used by our project, so I've moved its configuration to its corresponding subtable in our project definition.

Similarly, `tox` will pull in the configuration for these `[tool]` subtables when running these tools.

boukeas: Thank you. Also ref: https://tox.wiki/en/latest/config.html#pyproject-toml-native for the configuration style used here.

pedro-avalos: And https://github.com/tox-dev/tox-uv for specifics relating to using `tox` with `uv`

cli/pyproject.toml
boukeas, line 1: When we use `poetry` we employ the poetry "build-backend". Isn't there a `uv` "build-backend"? Why are we still using `setuptools`?

pedro-avalos: There is a `uv-build` package that we could move to, but I thought that PR was big enough as it was. I also am not sure if `uv-build` has a way to extract the version from `git` (which we do with `setuptools-scm`).

In my opinion, that should be a follow-up PR, as I think we should set up a versioning scheme before we start using `uv-build` (which is a relatively new addition to the `uv` suite and is less documented than the rest of `uv`).

pedro-avalos: See https://github.com/astral-sh/uv/pull/11446

boukeas: So we are essentially only using `uv` for running the tests, linting, etc but not for building the package?

cli/pyproject.toml
```diff
@@ -22,3 +38,28 @@ good-names = ["k", "v"]
...
+[tool.tox.env.lint]
+description = "Run linting tests"
+commands = [["flake8", "testflinger_cli"], ["pylint", "testflinger_cli"]]
+
+[tool.tox.env.unit]
+description = "Run unit tests"
+commands = [["pytest", "--cov=.", "--doctest-modules"]]
+
+[tool.uv]
+cache-keys = [{ file = "pyproject.toml" }, { git = { commit = true } }]
```
pedro-avalos, line 65: There is a `dir` key for this as well, perhaps the `testflinger_cli` should also be a cache key?

FYI: The default value for this is: `[{ file = "pyproject.toml" }, { file = "setup.py" }, { file = "setup.cfg" }]`

pedro-avalos: Actually, it seems like `dir` is not a valid key, marking this as resolved

cli/pyproject.toml
pedro-avalos: I tried adding `uv_build` as the build backend, but it failed. I do not think that it is possible with the dynamic versioning we use at the moment.

pedro-avalos: See https://github.com/astral-sh/uv/issues/8714

pedro-avalos: Also see https://github.com/astral-sh/uv/issues/11718

pedro-avalos: @boukeas,

One possible solution is to set the version to something static like `1.1.0-dev`. The Python package is not really published anywhere, and its version does not matter. The snap is still versioned with the commit hash in a completely different way.

This way we could start using the `uv_build` build backend here and close off completely the issue of using uv in the CLI package.

boukeas: I am quite conflicted. I don't see the point of versioning (not even manually) a package that never gets released or published as a package. Please follow your instinct around this, I am not sure what's best in this case.

pedro-avalos: I think for now we should just keep using `setuptools` for our build backend. I've already filed a follow-up card to replace the build backend: [CERTTF-580](https://warthogs.atlassian.net/browse/CERTTF-580).

Other projects that use `uv`, such as `snapcraft`, also use `setuptools` as their build backend.

.github/workflows/cli-tox.yml
```diff
@@ -18,15 +18,12 @@ jobs:
     runs-on: [self-hosted, linux, jammy, X64]
     strategy:
       matrix:
-        python: ["3.10"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
```
boukeas, line 21: I definitely don't mind this but can you explain the point of testing on this range of Python versions?

cli/HACKING.md
```diff
@@ -0,0 +1,101 @@
...
+
+```shell
+uv sync
+```
+
+This will create a virtual environment under `.venv`, to activate it, you need
+to source the following file:
+
+```shell
+source .venv/bin/activate
```
boukeas, line 49: So there's no equivalent to e.g. `poetry shell`?

boukeas (approved):
Thanks Pedro and apologies for taking too long to review :slightly_smiling_face:

cli/HACKING.md
pedro-avalos: Nope, it just creates a virtual environment

PR #541, opened 2025-04-01
feat(server): Define `provision_data` schema

pedro-avalos:
## Description

- Add `marshmallow-oneofschema` dependency
- Define the `provision_data` section of the job schema as a nested schema
- Define separate schemas for the supported device connectors

## Resolved issues

- Resolves [CERTTF-546](https://warthogs.atlassian.net/browse/CERTTF-546)

## Tests

Added unit tests

[CERTTF-546]: https://warthogs.atlassian.net/browse/CERTTF-546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/src/testflinger/api/schemas.py
+++ b/server/src/testflinger/api/schemas.py
@@ -83,6 +83,60 @@ class Attachment(Schema):
     device = fields.String(required=False)
 
 
+class ProvisionData(Schema):
+    """Schema for the `provision_data` section of a Testflinger job."""
+
+    url = fields.URL(required=False)
+    distro = fields.String(required=False)
+    kernel = fields.String(required=False)
+
+    attachments = fields.List(fields.Nested(Attachment), required=False)
+    use_attachment = fields.String(required=False)
+
+    user_data = fields.String(required=False)
+    ubuntu_sso_email = fields.Email(required=False)
+
+    # OEM Autoinstall specific fields
+    authorized_keys = fields.String(required=False)
+    redeploy_cfg = fields.String(required=False)
+    token_file = fields.String(required=False)
+
+    # MuxPi specific fields
+    boot_check_url = fields.String(required=False)
+    create_user = fields.Boolean(required=False)
+    media = fields.String(required=False)
+
+    # MAAS specific fields
+    # [TODO] Specify Nested schema to improve validation
+    disks = fields.List(fields.Dict(), required=False)
+
+    # Multi-device specific fields
+    # [TODO] Specify Nested schema to improve validation
+    jobs = fields.List(fields.Dict(), required=False)
+
+    # Zapper specific fields
+    zapper_provisioning_timeout = fields.Integer(required=False)
+
+    # Zapper IOT specific fields
+    preset = fields.String(required=False)
+    # [TODO] Specify Nested schema to improve validation
+    provision_plan = fields.Dict(required=False)
+    urls = fields.List(fields.URL(), required=False)
+
+    # Zapper KVM specific fields
+    alloem_url = fields.URL(required=False)
+    autoinstall_base_user_data = fields.String(required=False)
+    autoinstall_oem = fields.Boolean(required=False)
+    autoinstall_storage_layout = fields.String(required=False)
+    cmdline_append = fields.String(required=False)
+    live_image = fields.Boolean(required=False)
+    oem = fields.String(required=False)
+    robot_retries = fields.Integer(required=False)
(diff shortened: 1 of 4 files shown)


server/src/testflinger/api/schemas.py
```diff
@@ -83,6 +85,168 @@ class Attachment(Schema):
...
+                return slug
+            except ValidationError:
+                continue
+        raise ValidationError("Invalid provision data schema.")
+
+    def _dump(self, obj, *, update_fields=True, **kwargs):
+        """Removes the type field from the result."""
+        result = super()._dump(obj, update_fields=update_fields, **kwargs)
+        result.pop(self.type_field)
+        return result
```
boukeas, line 247: I was under the impression that what we actually do when we validate the job is to deserialize the data (without actually converting it into an object). So we definitely need `get_data_type`. But I don't think we ever actually serialize anything, do we? And if that's the case, I don't understand why we need to overload `_dump` and `get_obj_type`.

boukeas (commented):
I think I need some more information about how the polymorphic schema works and why we need to redefine so much of its internals.

server/src/testflinger/api/schemas.py
pedro-avalos: We need to override these because `marshmallow-oneofschema` expects and adds a `type` field in the schema. `get_obj_type` and `get_data_type` both expect and add a field. So I needed to remove the `type` field from the schema before dumping the data.

When I was running tests, I ran into some errors because of these functions. I'm not sure if we do serialize them in production, but the tests did serialize them.

boukeas: To answer the question, the `Job` schema isn't only used to validate input when POSTing to the `job/` endpoint, it is also used to validate the response when GETting from the `job/` and `job/<job_id>` endpoints, and this is why the methods for serializing are also relevant.

boukeas: @pedro-avalos if we can please add a comment that the parent `_dump` performs the equivalent of:
```
result[self.type_field] = self.get_obj_type(obj)
```
So we need to override it in order to reverse that.

server/src/testflinger/api/schemas.py
```diff
@@ -83,6 +85,168 @@ class Attachment(Schema):
...
+        """Get schema type depending on which schema is correctly parsed."""
+        for slug, schema in self.type_schemas.items():
+            try:
+                schema().load(data)
+                return slug
+            except ValidationError:
+                continue
+        raise ValidationError("Invalid provision data schema.")
+
+    def _dump(self, obj, *, update_fields=True, **kwargs):
```
boukeas, line 243: You are not using or modifying `update_fields`, I think it's safe to simplify the signature to:
```
def _dump(self, obj, *, **kwargs):
```
and modify the call to the parent method:
```
result = super()._dump(obj, **kwargs)
```

server/src/testflinger/api/schemas.py
```diff
@@ -83,6 +85,168 @@ class Attachment(Schema):
...
+
+    # [TODO] Specify Nested schema to improve validation
+    jobs = fields.List(fields.Dict(), required=True, validate=Length(min=1))
+
+
+class MuxPiProvisionData(Schema):
+    """Schema for the `provision_data` section of a MuxPi job."""
+
+    url = fields.URL(required=False)
+    use_attachment = fields.String(required=False)
```
boukeas, line 115: Attachments can actually be provided for _every_ phase, it's just that it wouldn't really make sense _except_ if you're provisioning using the attachment. For example, the `provision_data` for KernOS looks like this:
```
provision_data:
  attachments:
    - local: smoke_1.0_rpi-3b-plus-arm64.kernos.img.xz
  use_attachment: smoke_1.0_rpi-3b-plus-arm64.kernos.img.xz
  ...
```

In other words, you need another field here:
```
    attachments = fields.List(fields.Nested(Attachment), required=False)
```
(I'm only copying that from the test phase).

server/src/testflinger/api/schemas.py
pedro-avalos: Added the comment and updated the signature

boukeas (approved):
Thank you Pedro, separating the different provisioning schemas must have been hard work. This is really detailed now; much needed and appreciated.

PR #551, opened 2025-04-04
Parallelize agent unit tests

val500:
## Description
Agent unit tests take a while which can be limiting when running tests locally for development. This PR introduces pytest-xdist to the tox dependencies which is used to parallelize the running of pytest unit tests. This adds `-n auto` option to the pytest execution which auto detects the amount of cores and utilizes them for parallel test execution. Locally, this improved execution time from 254 seconds to 127 seconds.

## Tests
Run Tox locally in a virtual environment.

### DIFF
--- a/agent/tox.ini
+++ b/agent/tox.ini
@@ -8,6 +8,7 @@ deps =
     flake8
     pytest
     pylint
+    pytest-xdist
     pytest-mock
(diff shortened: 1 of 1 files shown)


pedro-avalos (approved):
Love it

PR #573, opened 2025-04-07
Add health check to pebble configuration

val500:
## Description
There has been some intermittent issues where one of the deployed server units becomes unresponsive but does not crash. This causes `juju status` to report that the unit is healthy, and requests sent to that unit end up timing out. Pebble has the ability to add [health checks](https://documentation.ubuntu.com/pebble/how-to/run-services-reliably/) to running services that can periodically check whether the service is responsive. This PR adds a health check that pings localhost:5000/v1/ every 10 seconds, timing out after 3 seconds. If this ping fails more than 5 times, the pebble service will automatically restart.

## Tests
Deployed these changes on staging. `/charm/bin/pebble checks` shows the `v1_up` check.

### DIFF
--- a/server/charm/src/charm.py
+++ b/server/charm/src/charm.py
@@ -41,6 +41,7 @@ class TestflingerCharm(ops.CharmBase):
         """Initialize the charm"""
         super().__init__(*args)
         self.pebble_service_name = "testflinger"
+        self.pebble_check_name = "v1_up"
         self.container = self.unit.get_container("testflinger")
         self._stored.set_default(
             reldata={},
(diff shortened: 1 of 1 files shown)


server/charm/src/charm.py
```diff
@@ -177,13 +178,27 @@ def _pebble_layer(self):
...
             "summary": "Testflinger server",
             "description": "pebble config layer for Testflinger server",
+            "checks": {
+                self.pebble_check_name: {
+                    "override": "replace",
+                    "period": "10s",
+                    "timeout": "3s",
+                    "threshold": 5,
+                    "http": {
+                        "url": "0.0.0.0:5000/v1/",
```
pedro-avalos, line 188: I assume this is done automatically, but could you add the scheme (`http://`) to the URL?

pedro-avalos (changes requested):
Minor nitpick, unless you know it won't work with the requested changes.

server/charm/src/charm.py
val500: Good catch, not including this caused pebble to seg fault.

pedro-avalos (approved):
LGTM

PR #594, opened 2025-04-10
feat(ci): Generalize Charm promotion CI

pedro-avalos:
## Description

- Allows the charms to be promoted to any channel we support

## Resolved issues

- Right now our process to promote to beta is manual, this should address that


.github/workflows/agent-host-charm-promote.yml
```diff
@@ -0,0 +1,35 @@
...
+  workflow_dispatch:
+    inputs:
+      destination-channel:
+        description: The channel to promote to
+        required: true
+        default: latest/stable
+        type: choice
+        options:
+          - latest/stable
+          - latest/beta
```
andrejvelichkovski, line 13: Maybe we can add `latest/candidate` here as well just to make it complete.

.github/workflows/agent-host-charm-promote.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Promote testflinger-agent-host charm
+        uses: canonical/charming-actions/promote-charm@2.7.0
+        with:
+          credentials: ${{ secrets.CHARMHUB_TOKEN }}
+          destination-channel: ${{ inputs.destination-channel }}
+          origin-channel: ${{ inputs.origin-channel }}
+          charm-path: agent/charms/testflinger-agent-host-charm
```
andrejvelichkovski, line 35: Maybe we can make charm path an option input parameter too so that we only define one action?

andrejvelichkovski (dismissed):
LGTM! Just left two small suggestions.

andrejvelichkovski (approved):
LGTM!

PR #595, opened 2025-04-10
fix(CI): Validate charm promote inputs

pedro-avalos:
## Description

- Validates that `charm-promote.yml` doesn't try to promote from the same channel to the same channel


val500 (approved):
LGTM

PR #596, opened 2025-04-10
fix(CI): Validate snap promote CI

pedro-avalos:
## Description

- Generalizes Snap promotion CI
- Specifies supported channels for the `to-channel` and `from-channel`
- Validates that the workflow is not called with to and from being the same channel


val500 (approved):
LGTM

PR #434, opened 2024-12-20
[CERTTF-457] feat: Github action for retrieving data from multi-device jobs

boukeas:
## Description

This PR introduces a Testflinger Github action that polls a multi-device job (receiving its ID as its input) and returns the job IDs and machine IPs of its child jobs as a JSON string.

## Resolved issues

Resolves [CERTTF-457](https://warthogs.atlassian.net/browse/CERTTF-457).

## Tests

The workflow below was used to submit a multi-device job without polling, use the `poll-multi` action to retrieve the data for the child jobs, and then display that data and confirm it's valid JSON.

```
name: Temporary workflow for testing multi-device polling action
on:
  workflow_dispatch:

jobs:
  submit-and-poll:
    name: Submit multi-device job and poll it for its data
    runs-on: [self-hosted, testflinger]
    steps:

    - name: Submit job to Testflinger
      id: submit
      uses: canonical/testflinger/.github/actions/submit@main
      with:
        poll: false
        job: |
          job_queue: multi-1
          name: multi-device-job
          output_timeout: 43200
          provision_data:
            jobs: 
            - job_queue: 202407-34216
              name: dell-xps-13-9350-0cc9-c34216
              reserve_data:
                ssh_keys:
                - lp:boukeas
                timeout: 600
            - job_queue: 202008-28168
              name: hp-eliteone800-g627-all-in-one-pc-c28168
              reserve_data:
                ssh_keys:
                - lp:boukeas
                timeout: 600

    - name: Retrieve multi-device job data
      id: poll
      uses: canonical/testflinger/.github/actions/poll-multi@CERTTF-457-multi-device-polling
      with:
        job-id: ${{ steps.submit.outputs.id }}
        sentinel-phase: reserve
    
    - name: Verify
      shell: bash
      env:
        JOBS: ${{ steps.poll.outputs.jobs }}
      run: |
        echo $JOBS | jq
```
Here is a [successful test run](https://github.com/canonical/certification-lab-ci/actions/runs/12907500543/job/35991056870). This is the output of the action:
```
{
  "7fcc82b1-1ab8-4fcc-9e3f-b0f9afbcec49": "10.102.182.56",
  "52fa5098-eb7d-4084-b9b0-b830214c089f": "10.102.161.241"
}
```
A syntactically incorrect version of the job was [also submitted](https://github.com/canonical/certification-lab-ci/actions/runs/12428851840), to test that the action detects provisioning (i.e. allocation) failures for multi-device jobs.

[...]

PR #605, opened 2025-04-11
chore: Document GitHub actions

pedro-avalos:
## Description

- Depends on #630
- Resolves a few minor linting/schema issues with the actions
  - Composite action inputs can only be strings
  - `poll-multi` was missing a `description` 
- Adds documentation for GitHub actions
  - This includes basic usage and API
- Simplifies the main README
  - Also adds a RTD badge for Testflinger

## Documentation

- `README.md` created for both `submit` and `poll-multi` actions.


.github/actions/poll-multi/README.md
```diff
@@ -0,0 +1,36 @@
...
+### Inputs
+
+| Key              | Description                                                                                 | Required           |
+| ---------------- | ------------------------------------------------------------------------------------------- | ------------------ |
+| `job-id`         | Job ID of the parent multi-device job.                                                      | :white_check_mark: |
+| `sentinel-phase` | Phase that all child jobs must reach before this action terminates. Defaults to `complete`. |                    |
+
+### Outputs
+
+- `jobs`: A JSON string containing an arry of objects with machine IPs and child job IDs
```
boukeas, line 31: I would argue that this output is the main incentive for using this action (at least this is why it was created). You can submit through the `submit` action but you don't really want to poll the multi-device job to completion. Instead you want to poll until the job reaches the point where you can extract this information, i.e. you can grab the IPs and job IDs of the child jobs. This action does this for you.

Let me suggest revisiting the [original PR](https://github.com/canonical/testflinger/pull/434) for a more complete example of how this can be used.

.github/actions/submit/action.yaml
```diff
@@ -10,11 +10,11 @@ inputs:
   poll:
     description: Specify if the submitted job should be tracked to completion
     required: false
-    default: false
+    default: "false"
```
boukeas, line 13: Why are the quotes necessary?

boukeas (changes requested):
That's very useful Pedro. I have linked to the original PR for the `poll-multi` action that contains a complete more meaningful example of how this action is meant to be used.

.github/actions/submit/action.yaml
pedro-avalos: Linters (e.g., vscode editor) will complain because composite actions are supposed to have only string inputs. Technically `false` is interpreted as a string by the action, so this makes that a little clearer and removes the complaint for editors

.github/actions/poll-multi/README.md
```diff
@@ -0,0 +1,62 @@
...
+### Inputs
+
+| Key              | Description                                                         | Required           | Default    |
+| ---------------- | ------------------------------------------------------------------- | ------------------ | ---------- |
+| `job-id`         | Job ID of the parent multi-device job.                              | :white_check_mark: |            |
+| `sentinel-phase` | Phase that all child jobs must reach before this action terminates. |                    | `complete` |
+
+### Outputs
+
+- `jobs`: A JSON string containing an array of objects with machine IPs
```
boukeas, line 56: This is clear enough but it would be nice to have an illustrative example (optional).

.github/actions/submit/README.md
```diff
@@ -0,0 +1,72 @@
...
+### Submit from a job file
+
+Assuming there is a previous `create-job` step in the workflow that creates a
+job file and outputs the path to it, you can submit a job with the following
+step:
+
+```yaml
+- name: Submit job
+  id: submit-job
+  uses: canonical/testflinger/.github/actions/submit@v1
```
boukeas, line 18: We don't have tagging in place just yet and, when we do, it won't be as straightforward as that because there are multiple actions (currently two) as well as other entities that might require tagging. So the tag will probably (unfortunately need to) resemble something like `v1-action-submit`. For now, we'll need to go with `main`

.github/actions/poll-multi/README.md
boukeas: I meant a real one but never mind.

boukeas (approved):

PR #622, opened 2025-04-15
Fix: add retries when trying to copy the agent SSH key

p-gentili:
## Description

This action might fail from time to time, especially with low-spec devices which might still be adjusting at this point. Adding retries to be a bit more robust.

## Resolved issues

N/A

## Documentation

N/A

## Web service API changes

None

## Tests

Added!

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
@@ -24,6 +24,7 @@ validating the configuration and preparing the API arguments.
 import json
 import logging
 import subprocess
+import time
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Tuple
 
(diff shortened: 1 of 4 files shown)


device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
```diff
@@ -131,7 +133,20 @@ def _copy_ssh_id(self):
...
+                break
+            except (
+                subprocess.CalledProcessError,
+                subprocess.TimeoutExpired,
+            ):
+                time.sleep(30)
+        else:
+            raise ProvisioningError(
+                "Cannot copy the agent's SSH key to the DUT",
+            )
```
omar-selo, line 149: This logic is already used somewhere [else in the code](https://github.com/canonical/testflinger/blob/594caf3acd85c6b5adc00df10936289622886c51/device-connectors/src/testflinger_device_connectors/devices/__init__.py#L262-L276). Can you extract both usages into a common helper function

omar-selo (changes requested):

omar-selo (approved):
LGTM!

PR #634, opened 2025-04-17
chore: Update snapcraft channels for lxd and juju

pedro-avalos:
## Description

- Local deployment has been failing for me recently; I noticed some of the snap channels seem to be out of date/incorrect.
- `lxd`'s channel `6.1/stable` doesn't seem to exist
- Set `juju` channel to latest in v3 to reduce times we need to change this

### DIFF
--- a/server/terraform/cloud-init.yaml
+++ b/server/terraform/cloud-init.yaml
@@ -4,8 +4,8 @@ package_upgrade: true
 package_reboot_if_required: true
 snap:
   commands:
-    - snap install lxd --channel=6.1/stable
-    - snap install juju --channel=3.6/stable
+    - snap install lxd --channel=6/stable
+    - snap install juju --channel=3/stable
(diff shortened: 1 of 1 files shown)


server/terraform/cloud-init.yaml
```diff
@@ -4,8 +4,8 @@ package_upgrade: true
 package_reboot_if_required: true
 snap:
   commands:
-    - snap install lxd --channel=6.1/stable
-    - snap install juju --channel=3.6/stable
+    - snap install lxd --channel=6/stable
+    - snap install juju --channel=3/stable
```
omar-selo, line 8: since juju 3.6 is LTS I'd recommend keeping it specifically

pedro-avalos: Done

omar-selo (approved):
Thanks for fixing that!

PR #637, opened 2025-04-18
update: reset the efi before releasing the machine

kevinyehk:
## Description

For most machines in TEL using MAAS provisioning, the boot order is modified during the oem-image installation process.
Previously `reset_efi` was executed at the beginning of provisioning, However, since the machine is now released after the test run, there is no opportunity to reset the boot order at the start.
Therefore, I propose this change.

## Resolved issues

The boot order of the machine can't be reset.


pedro-avalos (approved):
LGTM

PR #636, opened 2025-04-17
feat(agent): Update testflinger-agent to use uv

rene-oromtz:
## Description
- Adds [uv](https://docs.astral.sh/uv/) as the package manager for the testflinger-agent package
- Adds uv-build as the build backend.
- Updates layout of agent/ to a "src layout"
- Use tox-uv for running tox tests with uv.

## Resolved issues
Related to [CERTTF-576](https://warthogs.atlassian.net/browse/CERTTF-576) and [CERTTF-594](https://warthogs.atlassian.net/browse/CERTTF-594)

## Tests

[CERTTF-576]: https://warthogs.atlassian.net/browse/CERTTF-576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CERTTF-594]: https://warthogs.atlassian.net/browse/CERTTF-594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #642, opened 2025-04-22
Create new logging handlers in agent

val500:
## Description
This PR renames existing logging handlers, creates new ones, and adds some extra functionality to them. 

- All logging handlers follow the new abstract interface, `LogHandler`.
- The existing `LogUpdateHandler` has been renamed to `FileLogHandler` and inherits from `LogHandler`
- A new abstract class, `EndpointLogHandler` supports writing logs to an endpoint in Testflinger Server by fragment
- `OutputLogHandler` and `SerialLogHandler` support writing to the output and serial_output endpoints in Testflinger server respectively and inherit from `EndpointLogHandler`
- `TestflingerClient` has been modified to support posting to the output and serial_output endpoints with the correct fragment schema defined in a TypedDict `LogEndpointInput`
- `repost_job` function was removed from `TestflingerClient` due to it being unused

## Resolved issues
Resolves https://warthogs.atlassian.net/browse/CERTTF-570

## Tests
Tested in new test file `test_handlers.py`

### DIFF
--- a/agent/src/testflinger_agent/client.py
+++ b/agent/src/testflinger_agent/client.py
@@ -19,7 +19,7 @@ import shutil
 import tempfile
 import time
 from pathlib import Path
-from typing import Dict, List
+from typing import Dict, List, TypedDict
 from urllib.parse import urljoin
 
 import requests
@@ -31,6 +31,15 @@ from requests.packages.urllib3.util.retry import Retry
 from testflinger_agent.errors import TFServerError
 
 logger = logging.getLogger(__name__)
+
+
+class LogEndpointInput(TypedDict):
+    """Schema for Testflinger Log endpoints"""
+
+    fragment_number: int
+    timestamp: str
+    phase: str
+    log_data: str
 
 
 class TestflingerClient:
@@ -131,33 +140,6 @@ class TestflingerClient:
         job_data = self.get_result(job_id)
         if job_data:
             return job_data.get("job_state")
-
-    def repost_job(self, job_data):
-        """Resubmit the job to the testflinger server with the same id.
-
-        :param job_id:
-            id for the job on which we want to post results
-        """
-        job_uri = urljoin(self.server, "/v1/job")
-        job_id = job_data.get("job_id")
-        logger.info("Resubmitting job: %s", job_id)
-        job_output = """
-            There was an unrecoverable error while running this stage. Your job
-            will attempt to be automatically resubmitted back to the queue.
-            Resubmitting job: {}\n""".format(job_id)
-        self.post_live_output(job_id, job_output)
-        try:
-            job_request = self.session.post(job_uri, json=job_data)
-        except requests.exceptions.RequestException as exc:
-            logger.error(exc)
-            raise TFServerError("other exception") from exc
-        if not job_request:
-            logger.error(
-                "Unable to re-post job to: %s (error: %d)",
-                job_uri,
-                job_request.status_code,
-            )
-            raise TFServerError(job_request.status_code)
 
     def post_job_state(self, job_id, phase):
         """Update the job_state on the testflinger server."""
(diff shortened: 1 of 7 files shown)


pedro-avalos:
BTW, the charm tests are not running right now, but they are fixed on #636. I'd say let that merge first.

agent/testflinger_agent/handlers.py
```diff
@@ -12,22 +12,55 @@
...
-class LiveOutputHandler:
-    def __init__(self, client: TestflingerClient, job_id: str):
-        self.client = client
-        self.job_id = job_id
 
+class LogHandler(ABC):
+    @abstractmethod
     def __call__(self, data: str):
-        self.client.post_live_output(self.job_id, data)
+        pass
```
boukeas, line 24: This will never actually get called since this is an `ABC` but I think it's standard practice to `raise NotImplementedError` instead of just `pass`

agent/testflinger_agent/handlers.py
```diff
@@ -12,22 +12,55 @@
...
+from datetime import datetime, timezone
 
+from .client import TestflingerClient, LogEndpointInput
 
-class LiveOutputHandler:
-    def __init__(self, client: TestflingerClient, job_id: str):
-        self.client = client
-        self.job_id = job_id
 
+class LogHandler(ABC):
```
boukeas, line 21: Please add a meaningful docstring here, so that we know what to expect from derived classed.

boukeas: Reading forward, I think the derived classes also need docstrings.

agent/testflinger_agent/handlers.py
boukeas: Same applies to `EndpointLogHandler.write_to_endpoint` below.

agent/testflinger_agent/handlers.py
```diff
@@ -12,22 +12,55 @@
...
 
     def __call__(self, data: str):
         with open(self.log_file, "a") as log:
             log.write(data)
+
+
+class EndpointLogHandler(LogHandler):
+    def __init__(self, client: TestflingerClient, job_id: str, phase: str):
+        self.fragment_number = 0
+        self.log_buffer = ""
```
boukeas, line 39: Is this used anywhere or is it a leftover?

agent/testflinger_agent/client.py
```diff
@@ -291,26 +272,48 @@ def save_artifacts(self, rundir, job_id):
...
+        self,
+        job_id: str,
+        log_dict: LogEndpointInput,
+    ):
+        """Post output data to the testflinger server for this job"""
+        serial_uri = urljoin(
+            self.server, "/v1/result/{}/serial_output".format(job_id)
+        )
+        return self.post_log(log_dict, serial_uri)
+
```
boukeas, line 316: Looking at this, I am wondering if it would make sense to:
- Change the server endpoints to:
  - `v1/result/<job_id>/log/output`
  - `v1/result/<job_id>/log/serial`
- Merge `post_log`, `post_output` and `post_serial` to:
```
    def post_log(
        self,
        log_dict: LogEndpointInput,
        type: str,
    ):
        endpoint = urljoin(self.server, f"/v1/result/{job_id}/log/{type}")
        ...
```

boukeas (changes requested):
I really like the hierarchy for the handlers but I think it's important to actually document it in the code as well, for the future generations :grin: I have left a few comments and suggestions.

agent/testflinger_agent/handlers.py
val500: yeah that's a leftover, will remove

agent/testflinger_agent/client.py
val500: I do like this a lot better - do you think it's ok to introduce breaking API changes? We can update the CLI to use these endpoints, but do any users directly call the API?

boukeas: As far as I'm aware only the agents call the POST endpoints. For the GET endpoints I am also pretty sure it's only the CLI but, just in case, we could introduce these new ones endpoints and perhaps have the server redirect any GET requests to the new ones.

agent/src/testflinger_agent/client.py
```diff
@@ -33,6 +33,15 @@
 logger = logging.getLogger(__name__)
 
 
+class LogEndpointInput(TypedDict):
+    """Schema for Testflinger Log endpoints"""
+
+    fragment_number: int
+    timestamp: str
+    phase: str
+    log_data: str
```
boukeas: I am not a huge fan of `TypeDict`s, they "only" give you static type checking. Personally, I would use a `NamedTuple` or a frozen `dataclass` here; I prefer the way they are constructed and they both provide a mechanism for exporting to a dict, which you would only need to do in `post_log`, when you want to convert the data it holds to the dict you submit to the endpoint.

agent/src/testflinger_agent/handlers.py
```diff
@@ -12,22 +12,76 @@
...
+    """
+
+    def __init__(self, client: TestflingerClient, job_id: str, phase: str):
+        self.fragment_number = 0
+        self.client = client
+        self.phase = phase
+        self.job_id = job_id
+
+    @abstractmethod
+    def write_to_endpoint(self, data_dict: LogEndpointInput):
```
boukeas, line 56: This is really trivial but I recently read a suggestion to avoid using identifiers that refer to type (as in `_dict` here). I know it's common practice but I do think the suggestion actually makes sense. Feel free to ignore this if you don't.

val500: That's fine with me

agent/src/testflinger_agent/handlers.py
```diff
@@ -12,22 +12,78 @@
...
+
+
+class OutputLogHandler(EndpointLogHandler):
+    """
+    Implementation of EndpointLogHandler that writes logs to the output
+    endpoint in Testflinger server.
+    """
+
+    def write_to_endpoint(self, data: LogEndpointInput):
+        self.client.post_log(self.job_id, data, LogType.NORMAL_OUTPUT)
```
boukeas, line 79: Optional, but perhaps it would make sense to rename this to `STANDARD_OUTPUT` or `STD_OUTPUT`.

agent/tests/test_agent.py
```diff
@@ -398,7 +398,7 @@ def test_recovery_failed(self, agent, requests_mock):
             m.get("http://127.0.0.1:8000/v1/result/" + job_id, text="{}")
             m.post("http://127.0.0.1:8000/v1/result/" + job_id, text="{}")
             m.post(
-                "http://127.0.0.1:8000/v1/result/" + job_id + "/output",
+                "http://127.0.0.1:8000/v1/result/" + job_id + "/log/output",
```
boukeas, line 401: Perhaps use the `LogType` enum here? (and a couple of other places)

boukeas (approved):
I thinks this looks really great and thanks for addressing all the comments (even the more pedantic ones).

PR #643, opened 2025-04-22
Disable fail-fast for charm-update-libs.yml

pedro-avalos:
## Description

This allows the lib updates to run for all charms, independent of failures for the other jobs.

This helps in case the charm update libs run crashes for the server, but could otherwise work for the server (e.g., error when trying to merge a PR for one, but could still create a PR for the other).

## Tests

https://github.com/canonical/testflinger/actions/runs/15735740328

### DIFF
--- a/.github/workflows/charm-update-libs.yml
+++ b/.github/workflows/charm-update-libs.yml
@@ -13,6 +13,7 @@ jobs:
       contents: write
       pull-requests: write
     strategy:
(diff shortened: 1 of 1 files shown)


boukeas:
This looks innocuous and I'd be happy to approve but I'll just make my usual recurring comment first: can you please trigger the affected workflow and link to the run(s) in the description? Regardless of how small the change is, we need to demonstrate that we have tested it.

PR #644, opened 2025-04-23
fix(agent): Fix agent host charm

pedro-avalos:
## Description

We are seeing some errors with the agent host charm integration tests. These are due to the venv not finding a suitable installation of `testflinger-common`.

This PR updates the charm to install `testflinger-common` before installing the agent or device connectors packages.

Additionally, the `pip` call has been edited from using `pip install -I` to `pip install -U` so that only necessary upgrades are performed (the Testflinger packages are still reinstalled, but the dependencies are only upgraded if needed).

Adds `startsecs=0` to supervisor configuration files. See https://stackoverflow.com/questions/40909842/supervisor-fatal-exited-too-quickly-process-log-may-have-details

Thank you @rene-oromtz for helping to track, test, and debug this issue

## Resolved issues

See: https://github.com/canonical/testflinger/actions/runs/14604890358/job/40974459185

## Tests

Existing integration tests should pass

### DIFF
--- a/.github/workflows/agent-tox.yml
+++ b/.github/workflows/agent-tox.yml
@@ -4,11 +4,13 @@ on:
   push:
     branches: [ main ]
     paths:
+      - common/**  # testflinger-agent depends on testflinger-common
       - agent/**
       - .github/workflows/agent-tox.yml
   pull_request:
     branches: [ main ]
     paths:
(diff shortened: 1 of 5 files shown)


val500 (approved):
LGTM

PR #652, opened 2025-04-25
Server endpoint changes for output and serial logging

val500:
## Description

- GET and POST endpoints for "/result/<job_id>/(serial_)output" were renamed to "/result/<job_id>/log/(output/serial)"
- New `log_handlers.py`
    - Abstract class `LogHandler` is a generic interface for storing and retrieving logs to any backend
    - Class `MongoLogHandler` is an implementation of `LogHandler` which writes and reads specifically to Mongo
    - MongoLogHandler allows you to retrieve fragments by fragment_number and timestamp

## Web service API changes
The output and serial endpoints were renamed and given the following behavior:

POST /v1/result/<job_id>/log/output
Request:
```
{	
    fragment_number: int,
    timestamp: datetime,
    phase: string,
    log_data: string,
}
```

GET /v1/result/<job_id>/output
Parameters:
start_fragment_number: int
    - default: 0
timestamp: DateTime
phase: String

Response:
```
{
phase_logs: {
    <phase>: {
        last_fragment_number: <int>
        log_data: <string>
     }
     ...
}
}
```

## Tests
Unit tests were added to test_logging.py


server/src/testflinger/api/schemas.py
```diff
@@ -366,6 +366,31 @@ class StatusUpdate(Schema):
     events = fields.List(fields.Nested(JobEvent), required=False)
 
 
+class LogPost(Schema):
+    """Output Post and Serial Output Post Schema."""
```
boukeas, line 370: I see this pattern of mentioning output + serial output in many of the the docstrings and I don't think you need to do that (especially with the change to the endpoint where all/both types of output are now under `/log`). I believe it's sufficient to describe this as the schema for POSTing log fragments (or something to that effect).

server/src/testflinger/api/v1.py
```diff
@@ -457,86 +459,86 @@ def artifacts_get(job_id):
...
-@v1.get("/result/<job_id>/output")
+def log_get(job_id: str, args: dict, log_type: LogType):
+    """Get normal and serial output for a specified job_id."""
+    if not check_valid_uuid(job_id):
+        abort(400, message="Invalid job id\n")
+    query_schema = schemas.LogQueryParams()
+    try:
+        query_params = query_schema.load(args)
+    except ValidationError as err:
+        abort(400, message=err.messages)
```
boukeas, line 470: I don't think you need to do that here explicitly. Is it not possible to add a `v1.input(schemas.LogQueryParams, location="json")` decorator to the corresponding endpoints that use `log_get`?

server/src/testflinger/api/v1.py
```diff
@@ -457,86 +459,86 @@ def artifacts_get(job_id):
...
+        job_id,
+        log_type,
+        start_fragment,
+        start_timestamp,
+    )
+
+
+@v1.get("/result/<job_id>/log/output")
+@v1.output(schemas.LogGet)
 def output_get(job_id):
```
boukeas, line 484: I'm wondering if`output_{get|post}` and `serial_output_{get|post}` can be merged. I am not completely sure about the details but perhaps the decorator could be something like:
```
@v1.get("/result/<job_id>/log/<log_type>")
```
And the function signature:
```
def something_get(job_idL str, log_type: LogType):
```
Would that work?

server/src/testflinger/log_handlers.py
```diff
@@ -0,0 +1,130 @@
...
+class MongoLogHandler(LogHandler):
+    """Implementation of Log Handler for MongoDB backend."""
+
+    def __init__(self, mongo):
+        """Initialize mongo db object."""
+        self.mongo = mongo
+
+    def get_collection(self, log_type: LogType):
+        """Get the MongoDB collection corresponding to the LogType."""
+        match log_type:
```
boukeas, line 97: Are we OK using `match` with the Python version on the server? Is it 3.10?

server/src/testflinger/log_handlers.py
```diff
@@ -0,0 +1,130 @@
...
+            case LogType.NORMAL_OUTPUT:
+                return self.mongo.db.logs.output_fragments
+            case LogType.SERIAL_OUTPUT:
+                return self.mongo.db.logs.serial_output_fragments
```
boukeas, line 101: I am not sure it's a good idea, so treat this more as a thought rather than a suggestion: my instinctive reaction when reading this was "why aren't _all_ fragments stored in `mongo.db.logs`? The log type could be a field.

boukeas (changes requested):
This is starting to shape up very nicely. I have left a few comments and suggestions.

server/src/testflinger/api/v1.py
val500: I think that's specifically for validating the request body itself. The fragment_number/timestamp come in as URL parameters, and I don't think there's any Flask decorator for validating those afaik.

server/src/testflinger/api/v1.py
val500: I think so - can abort with a 404 if it isn't 'output' or 'serial'

server/src/testflinger/log_handlers.py
val500: I believe so, according to server/.python-version

server/src/testflinger/api/v1.py
val500: Seems like the flask way of doing this is Custom Converters: https://werkzeug.palletsprojects.com/en/stable/routing/#custom-converters

server/src/testflinger/log_handlers.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class LogHandler(ABC):
+    """
+    Abstract class with methods for storing and retrieving logs
+    from a generic backend.
+    """
+
+    @abstractmethod
+    def store_log_fragment(self, job_id: str, data: str, log_type: LogType):
```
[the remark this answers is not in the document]
boukeas: Feel free to disagree with me here, but if you are looking to specify the interface for the `LogHandler` then that's the only thing you should be thinking about, not how it is used by the rest of the code. So for me, it feels more natural for the `store_log_fragment` method to receive a log fragment (i.e. some kind of `LogFragment` object) as an argument, not parts of the fragment (the job ID, the log_type, and a fuzzy `data` dict that could contain anything) that the method assembles.

server/src/testflinger/log_handlers.py
```diff
@@ -0,0 +1,117 @@
...
+    def retrieve_log_fragments(
+        self,
+        job_id: str,
+        log_type: LogType,
+        phase: TestPhase,
+        start_fragment: int = 0,
+        start_timestamp: datetime = None,
+    ) -> List[dict]:
+        """
+        Retrieve log fragments from the handler specific backend.
```
boukeas, line 47: Why does this have to be a list instead of e.g. an `Iterable`? This could be a generator (which makes a lot of sense especially if you are retrieving a lot of fragments). Also, I am not sure if you should be returning `dict`s instead of `LogFragment` objects. It's convenient to be returning `dict`s now because that's exactly what MongoDB stores but that doesn't suggest a structure for the contents of these dicts.

server/src/testflinger/log_handlers.py
```diff
@@ -0,0 +1,117 @@
...
+    ) -> dict:
+        """
+        Return a dictionary with the combined log fragments and the last
+        fragment number retrieved.
+        """
+        fragments = self.retrieve_log_fragments(
+            job_id, log_type, phase, start_fragment, start_timestamp
+        )
+        data_list = [f["log_data"] for f in fragments]
+        log_data = "".join(data_list)
```
boukeas, line 68: This is quite wasteful and probably slow, especially if there are a lot of fragments. First you store the fragments in a list, then you store the log data in a separate list and then you join. My suggestion would be to use something like this:
```
log_data = "".join(
  fragment["log_data"]
  for fragment in self.retrieve_log_fragments(...)
)
```
with `retrieve_log_fragments` being a generator.

boukeas: Ok, I see how this wouldn't readily allow you to access the `last_fragment_number` below. It needs a bit more thought :grin:

server/src/testflinger/log_handlers.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, mongo):
+        """Initialize mongo db object."""
+        self.mongo = mongo
+
+    def store_log_fragment(self, job_id: str, data: dict, log_type: LogType):
+        """Store logs in the output/serial_output collection in MongoDB."""
+        log_collection = self.mongo.db.logs
+        timestamp = datetime.now(timezone.utc)
+        data["job_id"] = job_id
+        data["log_type"] = log_type
```
boukeas, line 92: And if you received a `LogFragment` as an argument you wouldn't have to do this. You are mirroring how the endpoint works (i.e. the fact that the job_id and the log type come from the endpoint, and the `data` comes from the payload) in a class that doesn't need to do that.

server/src/testflinger/log_handlers.py
```diff
@@ -0,0 +1,117 @@
...
+        query = {
+            "job_id": job_id,
+            "log_type": log_type,
+            "phase": phase,
+            "fragment_number": {"$gte": start_fragment},
+        }
+        if start_timestamp is not None:
+            query["timestamp"] = {"$gte": start_timestamp}
+
+        return list(log_collection.find(query).sort("fragment_number"))
```
boukeas, line 117: If you wanted to turn this into a generator, it would be `yield from log_collection.find...`

boukeas (changes requested):

server/src/testflinger/log_handlers.py
val500: I think I like the generator idea, but I'm not sure if it's possible to avoid loading the list of fragments, creating a list of log_data strings, then joining those into one string. I think even if you use a generator, all of that will still implicitly happen.

boukeas (commented):
I have some final (I swear) comments :grimacing:

server/src/testflinger/api/v1.py
```diff
@@ -457,85 +461,83 @@ def artifacts_get(job_id):
...
-    if output:
-        return "\n".join(output)
-    return "", 204
+    def to_python(self, value):
+        """Validate log type URL parameter."""
+        try:
+            log_type = LogType(value)
+            return log_type
+        except ValueError:
+            raise ValidationError("Invalid log type") from None
```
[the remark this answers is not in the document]
val500: I think this is necessary to implement the BaseConverter interface: https://werkzeug.palletsprojects.com/en/stable/routing/#custom-converters

PR #654, opened 2025-04-28
Update to jammy

pedro-avalos:
## Description

Updates all components to Ubuntu 22.04 LTS and Python version 3.10:

- Updated `.python-version` to 3.10: common, device-connectors, and agent.
- Updated testenv image used in `tf-test` to jammy in agent host charm.
- Updated agent host charm to build for jammy rather than both focal and jammy.
- Updated black formatting for device-connectors (updated version of black pulled in).
- Updated `uv.lock` for agent, common, and device-connectors


val500 (approved):
LGTM

PR #656, opened 2025-04-29
fix(server): Update dockerfile

pedro-avalos:
## Description

- Add uv venv to path
- optimize layering
- add comments

## Tests

Built and tested locally


val500 (approved):
LGTM

PR #662, opened 2025-05-01
fix(server): Add proxy variables to server charm and terraform

pedro-avalos:
## Description

Adds proxy variables to server charm

### DIFF
--- a/server/terraform/variables.tf
+++ b/server/terraform/variables.tf
@@ -65,6 +65,24 @@ variable "jwt_signing_key" {
   default     = "secret"
 }
 
+variable "http_proxy" {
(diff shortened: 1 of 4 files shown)


server/terraform/variables.tf
```diff
@@ -65,6 +65,24 @@ variable "jwt_signing_key" {
...
+  type        = string
+  default     = ""
+}
+
+variable "no_proxy" {
+  description = "Resources that we should abe able to access bypassing proxy"
+  type        = string
+  default     = "localhost,127.0.0.1,::1"
+}
+
```
boukeas, line 85: Are you sure these are being anywhere in the Terraform plan?

boukeas (approved):

PR #663, opened 2025-05-01
fix(server): Add python3-venv installation

pedro-avalos:
## Description

Python is not included in the base image


val500 (approved):
LGTM

PR #664, opened 2025-05-01
fix(server): Add server venv to path

pedro-avalos:
## Description

- Adds the server virtual environment to the container's path
- Adds some optimizations to the startup time (compile bytecode)
- Fixes warnings that might show up when rebuilding with cache


val500 (approved):
LGTM

PR #665, opened 2025-05-02
Fix warnings related to non-test classes picked up by pytest

pedro-avalos:
## Description

This silences some of the warnings we get in `pytest` runs (caused by pytest picking up our `Testflinger*` or `Test*` classes as if they were test classes.

To avoid this, I added `__test__=False` to the affected classes.

### DIFF
--- a/agent/.gitignore
+++ b/agent/.gitignore
@@ -1,0 +1,1 @@
+*.coverage.*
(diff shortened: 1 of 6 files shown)


val500 (approved):
LGTM

PR #666, opened 2025-05-05
Logging Redesign

val500:
## Description
This PR combines several other PRs relating to logging redesign to be merged at the same time


pedro-avalos (approved):
Considering the components have been tested and reviewed in the past PRs, I don't think there is much more I can comment on here. The tests are passing, and the implementation looks good. Thank you for all the incredible work @val500!

PR #667, opened 2025-05-05
feat(ci): Split tox tests in workflows

pedro-avalos:
## Description

This should make parsing what is failing more readable.

I've also separated the charm tests from the main tests so we don't have to wait for concierge setup.


val500 (approved):
LGTM

PR #668, opened 2025-05-06
Cli check snap confined

### LINKED ISSUE #511: Can't download artifacts in some directories
Hey,
it seems odd and a common issue, but I found no existing report about it.
Let me describe what I see and maybe it is known and trivial ... or not :-)

I wanted to get the artifacts of a run and I didn't want to bother about cleaning them up later.
So I tried to download to /tmp, but nothing happened.

```
$ cd /tmp
$ ll artifacts.tgz
ls: cannot access 'artifacts.tgz': No such file or directory
$ testflinger-cli artifacts 521d1ed6-fa3e-4cb5-9595-68784dff4948
Downloading artifacts tarball...
Artifacts downloaded to artifacts.tgz
$ ll artifacts.tgz
ls: cannot access 'artifacts.tgz': No such file or directory
```

Testflinger has no error message, nor a bad RC - but the tarball isn't there.
I checked dmesg for snap isolation denials - none are reported.
The very same works if changing to another directory.

```
$ cd ~/Downloads
$ testflinger-cli artifacts 521d1ed6-fa3e-4cb5-9595-68784dff4948
Downloading artifacts tarball...
Artifacts downloaded to artifacts.tgz
$ ll artifacts.tgz
-rw-rw-r-- 1 paelzer paelzer 39886 Mär 11 07:17 artifacts.tgz
```

`/tmp` is writable, not full or any of that AFAICS.
The next obvious thought is snap isolation and defaults.
Connections look sane:

```
$ snap connections testflinger-cli 
Interface  Plug                     Slot      Notes
home       testflinger-cli:home     :home     -
network    testflinger-cli:network  :network  -
```

But I've found it in the private tmp of the snap

```
$ sudo ls -laF snap-private-tmp/snap.testflinger-cli/tmp/artifacts.tgz
-rw-rw-r-- 1 paelzer paelzer 39886 Mär 11 07:18 snap-private-tmp/snap.testflinger-cli/tmp/artifacts.tgz
```

Well that is not too helpful.

I understand that if you just take `pwd` as an argument initially being `/tmp` and then without any mapping use `/tmp` inside the snap it will land in the private temp. But that is confusing to users. Since the UX of `testflinger-cli artifacts` is meant to download to the current dir, or to the one given by `--filename` the tool needs to understand that it's internal view of `/tmp` is not matching what the user knows and sees.

I checked this against `--filename` and it behaved the same:

```
$ testflinger-cli artifacts 521d1ed6-fa3e-4cb5-9595-68784dff4948 --filename /tmp/test.tgz
Downloading artifacts tarball...
Artifacts downloaded to /tmp/test.tgz
$ ll /tmp/test.tgz
ls: cannot access '/tmp/test.tgz': No such file or directory
$ sudo ls /tmp/snap-private-tmp/snap.testflinger-cli/tmp/test.tgz
/tmp/snap-private-tmp/snap.testflinger-cli/tmp/test.tgz
```

The snap feature of private-tmp is fine when the app components only talk to themselves, nice for isolation.
But in the case of `testflinger-cli` I'd have expected either:

 * a) it works and the user gets the file in the actual /tmp
    (might need workarounds or reducing isolation - so it might not be the way you want to go)

 * b) it detects that it actually can't write to the place the user specified - and tell him that + bad return code
    (And check if the same is true for other directories to handle them the same)

pedro-avalos:
## Description

Check if file arguments are in snap-private directories (at the moment, I'm only aware of `/tmp` being an issue for us.).

## Resolved issues

Resolves #511

## Tests

- Added unit tests
- Existing tests CLI should pass.

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -35,7 +35,7 @@ import argcomplete
 import requests
 import yaml
 
-from testflinger_cli import autocomplete, client, config, history
+from testflinger_cli import autocomplete, client, config, helpers, history
 
 logger = logging.getLogger(__name__)
 
@@ -182,7 +182,7 @@ class TestflingerCli:
         parser.add_argument(
             "-c",
             "--configfile",
-            type=Path,
+            type=helpers.parse_filename,
             default=None,
             help="Configuration file to use",
         )
@@ -217,7 +217,9 @@ class TestflingerCli:
             help="Download a tarball of artifacts saved for a specified job",
         )
         parser.set_defaults(func=self.artifacts)
-        parser.add_argument("--filename", type=Path, default="artifacts.tgz")
+        parser.add_argument(
+            "--filename", type=helpers.parse_filename, default="artifacts.tgz"
+        )
         parser.add_argument("job_id").completer = partial(
             autocomplete.job_ids_completer, history=self.history
         )
@@ -361,7 +363,7 @@ class TestflingerCli:
         parser.add_argument("--wait-for-available-agents", action="store_true")
         parser.add_argument(
             "filename",
-            type=Path,
+            type=lambda x: helpers.parse_filename(x, parse_stdin=True),
             help="YAML or JSON file with your job definition, '-' for stdin",
         ).completer = argcomplete.completers.FilesCompleter(
             allowednames=("*.yaml", "*.yml", "*.json")
@@ -517,16 +519,15 @@ class TestflingerCli:
 
     def submit(self):
         """Submit a new test job to the server."""
-        if self.args.filename.name == "-":
-            data = sys.stdin.read()
-        else:
-            try:
-                data = self.args.filename.read_text(
-                    encoding="utf-8", errors="ignore"
-                )
-            except (PermissionError, FileNotFoundError) as error:
-                logger.exception(error)
-                sys.exit(1)
+        try:
+            data = (
+                self.args.filename.read_text(encoding="utf-8", errors="ignore")
+                if self.args.filename
+                else sys.stdin.read()
+            )
+        except (PermissionError, FileNotFoundError) as exc:
+            logger.exception(exc)
+            sys.exit(1)
         job_dict = yaml.safe_load(data)
 
         # Check if agents are available to handle this queue
--- a/cli/pyproject.toml
+++ b/cli/pyproject.toml
@@ -61,6 +61,7 @@ ignore = [
 "testflinger_cli/tests/*" = [
   "S101", # use of assert
   "S105", # use of harcoded passwords for mock
+  "S108", # use of temporary file or directory
 ]
 
 [tool.ruff.lint.pydocstyle]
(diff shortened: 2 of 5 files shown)


cli/testflinger_cli/__init__.py
```diff
@@ -361,7 +363,7 @@ def _add_submit_args(self, subparsers):
         parser.add_argument("--wait-for-available-agents", action="store_true")
         parser.add_argument(
             "filename",
-            type=Path,
+            type=lambda x: helpers.parse_filename(x, parse_stdin=True),
```
boukeas, line 366: What you need here is `partial(helpers.parse_filename, parse_stdin=True)`, with `partial` imported from `functools`

cli/testflinger_cli/__init__.py
```diff
@@ -361,7 +363,7 @@ def _add_submit_args(self, subparsers):
         parser.add_argument("--wait-for-available-agents", action="store_true")
         parser.add_argument(
             "filename",
-            type=Path,
+            type=partial(helpers.parse_filename, parse_stdin=True),
```
boukeas, line 366: You need a `from functools import partial`. Doesn't this fail when you try running it?

boukeas (changes requested):
Seeing `args.filename` as None when reading from standard input reminded me that `args.filename` is used, in some cases, to [determine the directory](https://github.com/canonical/testflinger/blob/5182b45a313bc0dc75296d1f79da06b4c44cb4e3/cli/testflinger_cli/__init__.py#L470) where attachments are read from. That line will raise an exception when `args.filename` is None so that should be handled appropriately.

cli/testflinger_cli/__init__.py
pedro-avalos: It's already imported

pedro-avalos:
@boukeas any updates on this PR?

cli/testflinger_cli/__init__.py
```diff
@@ -517,16 +530,15 @@ def build_auth_headers(self):
...
-                sys.exit(1)
+        try:
+            data = (
+                self.args.filename.read_text(encoding="utf-8", errors="ignore")
+                if self.args.filename
+                else sys.stdin.read()
+            )
+        except (PermissionError, FileNotFoundError) as exc:
+            logger.exception(exc)
+            sys.exit(1)
```
boukeas, line 541: I prefer the previous version of this because `data = sys.stdin.read()` cannot raise the exceptions we are catching and the previous version, although slightly longer, wraps the `try-except` around the code that actually can raise these exceptions.

boukeas (dismissed):
Thanks for addressing all the comments and apologies for taking so much time to get back to reviewing this again. All good, just a minor comment you might want to consider.

cli/testflinger_cli/__init__.py
pedro-avalos: Makes sense, I've undone it

boukeas (approved):

PR #669, opened 2025-05-07
oem_autoinstall: Disable copy oem public ssh keys after provisioning

Artur-at-work:
## Description
This disables the copy of `/etc/ssh/authorized_keys `file to DUT after provisioning. It means that only TF agents keys are copied after provisioning, but all the other users need to login with user/password or use TF reserve function.

## Documentation
n/a

## Web service API changes
n/a

## Tests
This change was tested in hot lab with user-data file. After provisioning, only agent's keys are added and` /etc/ssh/authorized_keys` doesn't exist.


Artur-at-work:
hi @kevinyehk, could you help to review this?

kevinyehk (approved):
LGTM +1

PR #672, opened 2025-05-08
Modify results endpoint to send combined log fragments

val500:
## Description
Since output and serial logs are now sent in fragments to the logging endpoints, they no longer need to be included as results. This PR introduces changes in the agent and server to reflect that.

Agent Changes:
- Output and Serial Logs are removed from results before being sent.
- `write_from_file` method was added to the Endpoint Handler to support sending logs from a file to the Testflinger logging endpoint in 1024 byte chunks
- At the end of each phase, the serial log file is sent to the logging endpoint through a new serial log handler using this new method.

Server Changes:
- `ResultGet` and `ResultPost` schema were added to replace the `Result` schema.
- `ResultGet` schema identical to `Result` schema
- For the `ResultGet` endpoint, log fragments are accumulated using a log handler and returned in the JSON.
- Output and serial fields for `ResultPost` were removed
- Status are included as dictionaries with the TestPhases as keys for `ResultPost`

## Web service API changes
POST Result Schema:
```
{
    "status": {
         <TestPhase>: <int>,
         ....
    }
    device_info: {...}, //Unchanged
    job_state: <string>,
}
```

### DIFF
--- a/agent/src/testflinger_agent/client.py
+++ b/agent/src/testflinger_agent/client.py
@@ -260,6 +260,9 @@ class TestflingerClient:
             logger.info("Submitting job outcome for job: %s", job_id)
             with open(outcome_file) as f:
                 data = json.load(f)
+                # Only include status in posted results
+                data.pop("output", None)
+                data.pop("serial", None)
                 data["job_state"] = "complete"
                 self.post_result(job_id, data)
             # Remove the outcome file so we don't retransmit
--- a/server/src/testflinger/database.py
+++ b/server/src/testflinger/database.py
@@ -433,3 +433,19 @@ def delete_restricted_queue(queue: str, client_id: str):
     mongo.db.client_permissions.update_one(
         {"client_id": client_id}, {"$pull": {"allowed_queues": queue}}
     )
+
+
+def get_job_results(job_id: str):
+    """Retrieve results for a specific job id."""
+    return mongo.db.jobs.find_one(
+        {"job_id": job_id}, {"result_data": True, "_id": False}
+    )
+
+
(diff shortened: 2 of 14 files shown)


agent/src/testflinger_agent/client.py
```diff
@@ -237,6 +237,9 @@ def transmit_job_outcome(self, rundir):
             logger.info("Submitting job outcome for job: %s", job_id)
             with open(outcome_file) as f:
                 data = json.load(f)
+                # Only include status in posted results
+                data.pop("output", None)
+                data.pop("serial", None)
```
boukeas, line 242: Why do we even need to include these in the data?

server/src/testflinger/database.py
```diff
@@ -399,3 +399,19 @@ def get_client_permissions(client_id: str) -> dict:
...
+    """Retrieve results for a specific job id."""
+    return mongo.db.jobs.find_one(
+        {"job_id": job_id}, {"result_data": True, "_id": False}
+    )
+
+
+def add_job_results(job_id: str, json_data: dict):
+    """Add results to specified job id with "result_data" prepended."""
+    # First, we need to prepend "result_data" to each key in the result_data
+    for key in list(json_data):
```
boukeas, line 414: if `json_data` is a dict then `for key in json_data` will iterate over the keys, you don't need the `list`.

agent/src/testflinger_agent/client.py
val500: Yeah, the output and serial logs are written here - I'm not sure if this is used by anyone which is why I kept it.

PR #678, opened 2025-05-19
Add CLI client endpoints to common module

pedro-avalos:
## Description

This PR creates a common client module that can be re-used by the CLI, agent, or custom scripts, libraries, etc.

## Tests

TODO

### DIFF
--- a/common/src/testflinger_common/client.py
+++ b/common/src/testflinger_common/client.py
@@ -1,0 +1,187 @@
+# Copyright (C) 2025 Canonical Ltd.
+"""Testflinger client module."""
+
+import base64
+import logging
+import urllib.parse
+from functools import cached_property
+from pathlib import Path
+from typing import Optional
+
+import requests
+
+logger = logging.getLogger(__name__)
+
+
+class Client:
+    """Testflinger client."""
+
+    def __init__(
+        self,
+        server: str,
+        client_id: Optional[str] = None,
+        client_secret: Optional[str] = None,
+    ) -> None:
+        self.server = server
+        self.client_id = client_id
+        self.client_secret = client_secret
+
+    @cached_property
+    def token(self) -> str:
+        """Return the token for the client."""
+        if not self.client_id or not self.client_secret:
+            return ""
+
+        endpoint = "/v1/oauth2/token"
+        id_key_pair = f"{self.client_id}:{self.client_secret}"
+        encoded_id_key_pair = base64.b64encode(id_key_pair.encode()).decode()
+        headers = {"Authorization": f"Basic {encoded_id_key_pair}"}
+        response = self.post(endpoint=endpoint, data={}, headers=headers)
+        return response.text
+
+    @cached_property
+    def auth_headers(self) -> Optional[dict]:
+        """Create an authorization header for the client."""
+        if not self.token:
+            return None
+        return {"Authorization": f"{self.token}"}
+
+    def submit_job(self, job_data: dict) -> str:
+        """Submit a test job to the server."""
+        endpoint = "/v1/job"
+        headers = self.auth_headers
+        response = self.post(endpoint=endpoint, data=job_data, headers=headers)
+        return response.json().get("job_id")
+
+    def submit_job_attachment(
+        self, job_id: str, path: Path, timeout_sec: float = 30
+    ) -> None:
+        """Submit a test job's attachment to the server."""
+        endpoint = f"/v1/job/{job_id}/attachments"
+        self.post_file(endpoint=endpoint, path=path, timeout_sec=timeout_sec)
+
+    def get_job_results(self, job_id: str) -> dict:
+        """Get the results of a test job."""
+        endpoint = f"/v1/result/{job_id}"
+        response = self.get(endpoint=endpoint)
+        return response.json()
+
+    def get_job_artifacts(self, job_id: str, path: Path) -> None:
+        """Get the artifacts of a test job."""
+        endpoint = f"/v1/result/{job_id}/artifact"
+        self.get_file(endpoint=endpoint, path=path)
+
+    def get_job(self, job_id: str) -> dict:
+        """Get the job definition for a specified ID."""
+        endpoint = f"/v1/job/{job_id}"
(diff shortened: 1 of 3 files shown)


common/src/testflinger_common/client.py
```diff
@@ -0,0 +1,321 @@
...
+        response = self.session.post(url, headers=headers)
+        response.raise_for_status()
+        return response.text
+
+    @cached_property
+    def auth_headers(self) -> Optional[dict]:
+        """Create an authorization header for the client."""
+        if not self.token:
+            return None
+        return {"Authorization": f"{self.token}"}
```
boukeas, line 74: `return {"Authorization": self.token}` should be fine here

common/src/testflinger_common/client.py
```diff
@@ -0,0 +1,321 @@
...
+        adapter = HTTPAdapter(max_retries=retry)
+        session.mount("http://", adapter)
+        session.mount("https://", adapter)
+        return session
+
+    @cached_property
+    def token(self) -> str:
+        """Return the token for the client."""
+        if not self.client_id or not self.client_secret:
+            return ""
```
boukeas, line 60: Personally I'd return `None` instead of the empty string here. But I am also wondering whether not having a token is actually cause for raising an exception.

common/src/testflinger_common/client.py
```diff
@@ -0,0 +1,307 @@
...
+        :param agent_id: The ID of the agent.
+        :param data: The data to post.
+        """
+        self.post(f"/v1/agents/data/{agent_id}", json=data, timeout_sec=30)
+
+    def post_job_action(self, job_id: str, action: str) -> None:
+        """Post an action to a test job."""
+        self.post(f"/v1/job/{job_id}/action", json={"action": action})
+
+    def get(
```
boukeas, line 263: Minor: I'd suggest moving these closer to the top of the class, before all the methods that actually use them. I think it makes it easier to read.

common/src/testflinger_common/client.py
```diff
@@ -0,0 +1,307 @@
...
+    @cached_property
+    def auth_headers(self) -> Optional[dict]:
+        """Create an authorization header for the client."""
+        return {"Authorization": self.token} if self.token else None
+
+    def submit_job(self, job_data: dict) -> str:
+        """Submit a test job to the server."""
+        headers = self.auth_headers
+        response = self.post("/v1/job", json=job_data, headers=headers)
+        return response.json().get("job_id")
```
boukeas, line 79: I think this should also return `response.json()`. Some questions to serve as arguments:
- Should the client module process responses or return them as they are?
- What if the response contains additional fields in the future?
- Don't we have requests for the CLI to include a `--json` option that displays output in JSON format?

common/src/testflinger_common/client.py
```diff
@@ -0,0 +1,307 @@
...
+        """Get the attachments of a test job."""
+        endpoint = f"/v1/job/{job_id}/attachments"
+        response = self.get(endpoint, timeout_sec=timeout_sec, stream=True)
+        with path.open("wb") as attachments:
+            for chunk in response.iter_content(chunk_size=4096):
+                attachments.write(chunk)
+
+    def post_job_results(self, job_id: str, data: dict) -> None:
+        """Post the results of a test job."""
+        self.post(f"/v1/result/{job_id}", json=data, timeout_sec=30)
```
boukeas, line 106: I see most of these methods using `self.post` returning `None`. What are the actual responses and are we not interested in them? Perhaps the client should return the response and we should leave it up to whoever is using the client to decide if they should be ignored.

common/src/testflinger_common/client.py
```diff
@@ -0,0 +1,307 @@
...
+
+    def get_job_artifacts(
+        self, job_id: str, path: Path, timeout_sec: float = 30
+    ) -> None:
+        """Get the artifacts of a test job."""
+        endpoint = f"/v1/result/{job_id}/artifact"
+        response = self.get(endpoint, timeout_sec=timeout_sec, stream=True)
+        with path.open("wb") as artifacts:
+            for chunk in response.iter_content(chunk_size=4096):
+                artifacts.write(chunk)
```
boukeas, line 121: This essentially "streams" a file from an endpoint and the code is essentially identical to `get_job_attachments`. So there's an opportunity for refactoring the streaming part into a new method.

boukeas: Same with `post_job_artifacts` and `submit_job_attachments`.

common/src/testflinger_common/client.py
```diff
@@ -0,0 +1,307 @@
...
+
+    def get_job_output(self, job_id: str) -> str:
+        """Get the latest output for a test job."""
+        response = self.get(f"/v1/job/{job_id}/output")
+        return response.text
+
+    def get_job_serial_output(self, job_id: str) -> str:
+        """Get the latest serial output for a test job."""
+        response = self.get(f"/v1/job/{job_id}/serial_output")
+        return response.text
```
boukeas, line 200: Not much you can do about it now but these endpoints will change with Varun's work on the log fragments.

common/src/testflinger_common/client.py
pedro-avalos: I've created common `get_file` and `post_file`

common/src/testflinger_common/client.py
pedro-avalos: I've updated them to return either a JSON dictionary or text.

We should go through and update the server to always return a JSON response. I don't know why the responses are sometimes text...

common/src/testflinger_common/client.py
pedro-avalos: Done

common/src/testflinger_common/client.py
pedro-avalos: Moved them right after the properties

common/src/testflinger_common/client.py
pedro-avalos: I've now updated the response to always be JSON

pedro-avalos:
Closing as this is stale and we need further work to even be able to use it in our CLI snap.

PR #679, opened 2025-05-20
[CERTTF-550] Support private-endpoint runners in `submit` action

boukeas:
## Description

Github workflows that use the Testflinger `submit` action need to use runners that are able to access the Testflinger server. In order to be able to achieve that in private-endpoint runners, a [workaround](https://chat.canonical.com/canonical/pl/uf7yxdxphinqzmfnwu1eyn158o) to bypass `aproxy` is required.

This PR incorporates the workaround in a new step of the `submit` action. The step is optional; it is only performed when a private-endpoint runner is used. The simplest way to detect that currently is to check that the runner name begins with `private`. 

## Resolved issues

Partly resolves [CERTTF-550](https://warthogs.atlassian.net/browse/CERTTF-550). The original issue is about documenting how to use private-endpoint runners but this PR precedes it as it facilitates their usage (it performs the workaround so that the user doesn't have to).

## Documentation

The documentation for the `submit` action does not need to change. A follow-up PR will extend the documentation available
in the `certification-ops` repo on how to use Testflinger in Github CI workflows.

## Tests

A simple test workflow that uses the `submit` action was created on a branch of the `certification-lab-ci` repo, where Testflinger-enabled runners are available. The runners that are used for running this workflow are controlled through an input parameter. 
```
name: Test submission through private-endpoint runner
run-name: Test submission on runner ${{ inputs.runs-on }}
on:
  workflow_dispatch:
    inputs:
      runs-on:
        type: string
      queue:
        type: string
      dry-run:
        type: boolean
        required: false
        default: true
      server:
        type: string
        required: false
        default: testflinger.canonical.com

jobs:
  generate-submit-job:
    runs-on: ${{ fromJSON(inputs.runs-on) }}
    steps:
    - name: Submit job to Testflinger
      id: submit
      uses: canonical/testflinger/.github/actions/submit@CERTTF-550-support-private-endpoint-submit
      with:
        poll: true
        dry-run: false
        server: ${{ inputs.server }}
        job: |
          job_queue: ${{ inputs.queue }}
          test_data:
            test_cmds:
              echo "Hello world"
```

The sections below describe the different ways that the workflow was triggered, in order to test the `submit` action.

### Testflinger-enabled runner

```
gh workflow run test-private-endpoint-submit.yaml \
--ref CERTTF-550-test-private-endpoint-submit \
--field runs-on='["self-hosted","testflinger"]' \
--field queue=rpi3 \
--field dry-run=false
```

The [log](https://github.com/canonical/certification-lab-ci/actions/runs/15184728765/job/42702357789) shows that the workflow was executed on `gh-runner-certification-lab-ci-...` (a Testflinger-enabled runner deployed specifically for the `certification-lab-ci` repo) and the Testflinger job was submitted successfully _without_ applying the workaround, as expected.

### Private-endpoint runner

```
gh workflow run test-private-endpoint-submit.yaml \
--ref CERTTF-550-test-private-endpoint-submit \
--field runs-on='["self-hosted-linux-amd64-jammy-private-endpoint-medium"]' \
--field queue=rpi3 \
--field dry-run=false
```

The [log](https://github.com/canonical/certification-lab-ci/actions/runs/15184797536/job/42702592367) shows that the workflow was executed on `private-openstack-amd64-jammy-medium-ps6-...` (a Canonical-hosted private-endpoint runner) and the Testflinger job was submitted successfully after applying the workaround.

Similarly for the Testflinger staging server:

```
gh workflow run test-private-endpoint-submit.yaml \
--ref CERTTF-550-test-private-endpoint-submit \
--field runs-on='["self-hosted-linux-amd64-jammy-private-endpoint-medium"]' \
--field server=testflinger-staging.canonical.com \
--field queue=audino \
--field dry-run=false
```
Here is the corresponding [log](https://github.com/canonical/certification-lab-ci/actions/runs/15184825462/job/42702685657).

### "Regular" self-hosted runner

```
gh workflow run test-private-endpoint-submit.yaml \
--ref CERTTF-550-test-private-endpoint-submit \
--field runs-on='["self-hosted","medium","amd64"]' \
--field queue=rpi3 \
--field dry-run=false
```

The [log](https://github.com/canonical/certification-lab-ci/actions/runs/15159871327/job/42623303683) shows that the workflow was executed on a Canonical-hosted runner that doesn't have network access to the Testflinger server and the `submit` action failed at the point where the access to the Testflinger server is checked.

[CERTTF-550]: https://warthogs.atlassian.net/browse/CERTTF-550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/submit/action.yaml
+++ b/.github/actions/submit/action.yaml
@@ -34,6 +34,8 @@ runs:
   steps:
     - name: Bypass aproxy (for private-endpoint runners)
       if: startsWith(runner.name, 'private-')
+      env:
+        SERVER: ${{ inputs.server }}
       shell: bash
       run: |
         echo "::group::Bypass aproxy (private-endpoint runner)"
@@ -41,7 +43,7 @@ runs:
         sudo DEBIAN_FRONTEND=noninteractive apt-get install -qqy dnsutils
         sudo snap set aproxy listen=:28443
         sudo nft -f - << EOF
-        define testflinger-ips = { $(dig +short testflinger.canonical.com | grep -o '[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}' | tr '\n' ',') }
+        define testflinger-ips = { $(dig +short $SERVER | grep -o '[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}' | tr '\n' ',') }
         define default-ip = $(ip route get $(ip route show 0.0.0.0/0 | grep -oP 'via \K\S+') | grep -oP 'src \K\S+')
         define exclude-ips = { 10.0.0.0/8, 127.0.0.1/8, 172.16.0.0/12, 192.168.0.0/16, \$testflinger-ips }
         table ip aproxy
--- a/.github/actions/submit/action.yaml
+++ b/.github/actions/submit/action.yaml
@@ -39,10 +39,9 @@ runs:
       shell: bash
       run: |
         echo "::group::Bypass aproxy (private-endpoint runner)"
-        sudo apt-get -qq update && sudo DEBIAN_FRONTEND=noninteractive apt-get install -qqy dnsutils
         sudo snap set aproxy listen=:28443
         sudo nft -f - << EOF
-        define testflinger-ips = { $(dig +short $SERVER | grep -o '[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}' | tr '\n' ',') }
+        define testflinger-ips = { $(getent ahostsv4 $SERVER | awk '/RAW/ {print $1}' | paste -sd,) }
         define default-ip = $(ip route get $(ip route show 0.0.0.0/0 | grep -oP 'via \K\S+') | grep -oP 'src \K\S+')
         define exclude-ips = { 10.0.0.0/8, 127.0.0.1/8, 172.16.0.0/12, 192.168.0.0/16, \$testflinger-ips }
         table ip aproxy
@@ -58,6 +57,7 @@ runs:
           }
         }
         EOF
+        echo Bypass complete
         echo "::endgroup::"
 
     - name: Test connection to Testflinger server
--- a/.github/actions/submit/action.yaml
+++ b/.github/actions/submit/action.yaml
@@ -36,6 +36,7 @@ runs:
       if: startsWith(runner.name, 'private-')
       shell: bash
       run: |
+        export DEBIAN_FRONTEND=noninteractive 
         sudo apt-get -qqy update && sudo apt-get install -qqy dnsutils
         sudo snap set aproxy listen=:28443
         sudo nft -f - << EOF
--- a/.github/actions/submit/action.yaml
+++ b/.github/actions/submit/action.yaml
@@ -39,26 +39,12 @@ runs:
       shell: bash
       run: |
         echo "::group::Bypass aproxy (private-endpoint runner)"
-        sudo snap set aproxy listen=:28443
-        sudo nft -f - << EOF
-        define testflinger-ips = { $(getent ahostsv4 $SERVER | awk '/RAW/ {print $1}' | paste -sd,) }
-        define default-ip = $(ip route get $(ip route show 0.0.0.0/0 | grep -oP 'via \K\S+') | grep -oP 'src \K\S+')
-        define exclude-ips = { 10.0.0.0/8, 127.0.0.1/8, 172.16.0.0/12, 192.168.0.0/16, \$testflinger-ips }
-        delete chain ip aproxy prerouting
-        delete chain ip aproxy output
-        table ip aproxy {
-          chain prerouting {
-            type nat hook prerouting priority dstnat; policy accept;
-            ct state established,related accept
-            ip daddr != \$exclude-ips tcp dport { 80, 443 } counter dnat to \$default-ip:28443
-          }
-          chain output {
-            type nat hook output priority -100; policy accept;
-            ct state established,related accept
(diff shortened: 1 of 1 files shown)


.github/actions/submit/action.yaml
```diff
@@ -32,6 +32,33 @@ outputs:
 runs:
   using: composite
   steps:
+    - name: Bypass aproxy (for private-endpoint runners)
+      if: startsWith(runner.name, 'private-')
+      shell: bash
+      run: |
+        echo "::group::Bypass aproxy (private-endpoint runner)"
+        sudo DEBIAN_FRONTEND=noninteractive apt-get -qqy update && \
+        sudo DEBIAN_FRONTEND=noninteractive apt-get install -qqy dnsutils
```
pedro-avalos, line 41: Since these two commands are joined by `&&`, could you indent the second line?

```suggestion
        sudo DEBIAN_FRONTEND=noninteractive apt-get -qqy update && \
          sudo DEBIAN_FRONTEND=noninteractive apt-get install -qqy dnsutils
```

Additionally, we should probably just export `DEBIAN_FRONTEND=noninteractive` at the start of this step; I can't think of a reason for that not to be set throughout anyways.

.github/actions/submit/action.yaml
```diff
@@ -32,6 +32,33 @@ outputs:
...
+        sudo snap set aproxy listen=:28443
+        sudo nft -f - << EOF
+        define testflinger-ips = { $(dig +short testflinger.canonical.com | grep -o '[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}' | tr '\n' ',') }
```
pedro-avalos, line 44: Hm, the action allows a `server` parameter. Should this be generalized for any server?

Or perhaps this step only makes sense in our production Testflinger deployment; if that's the case, add a check in the `if` that `server` is `testflinger.canonical.com`...

pedro-avalos (changes requested):
The networking logic looks okay to me, and the tests are passing.

I have left some comments regarding when this step should run (since the action accepts a `server` parameter).

.github/actions/submit/action.yaml
boukeas: > we should probably just export DEBIAN_FRONTEND=noninteractive at the start of this step; I can't think of a reason for that not to be set throughout anyways.

I could be wrong but I don't think the commands that run after `sudo` inherit the environment of the parent shell. I will try it out.

.github/actions/submit/action.yaml
boukeas: No, you are right, if we're accepting the server as an argument this should be generalized. I don't think it's ever going to be used but right is right.

.github/actions/submit/action.yaml
pedro-avalos: You are right, I think it requires a `-E` flag, but that's a bit of a big security hole.

.github/actions/submit/action.yaml
boukeas: I extended the test workflow to accept a `server` input. I then triggered the workflow on the private-endpoint runner as before (i.e. default server):
```
gh workflow run test-private-endpoint-submit.yaml \
--ref CERTTF-550-test-private-endpoint-submit \
--field runs-on='["self-hosted-linux-amd64-jammy-private-endpoint-medium"]'
```
And then retriggered, specifying the staging server:
```
gh workflow run test-private-endpoint-submit.yaml \
--ref CERTTF-550-test-private-endpoint-submit \
--field runs-on='["self-hosted-linux-amd64-jammy-private-endpoint-medium"]' \
--field server=testflinger-staging.canonical.com
```
The [log](https://github.com/canonical/certification-lab-ci/actions/runs/15157947001/job/42617107955) for the first one shows the workflow was, once again, successful. The [log](https://github.com/canonical/certification-lab-ci/actions/runs/15157974037/job/42617193173) for the second one shows the Testflinger staging server was successfully pinged but the job submission failed, as expected, because there is no agent for the device the job is submitted to.

.github/actions/submit/action.yaml
boukeas: So. `DEBIAN_FRONTEND=noninteractive` for the update was only necessary because I was using the `-y` option. That doesn't really need to be there for the update, so now the line is much shorter.

mihalicyn:
Great stuff! Thank you!

.github/actions/submit/action.yaml
```diff
@@ -32,6 +32,34 @@ outputs:
...
+      if: startsWith(runner.name, 'private-')
+      env:
+        SERVER: ${{ inputs.server }}
+      shell: bash
+      run: |
+        echo "::group::Bypass aproxy (private-endpoint runner)"
+        sudo apt-get -qq update && sudo DEBIAN_FRONTEND=noninteractive apt-get install -qqy dnsutils
+        sudo snap set aproxy listen=:28443
+        sudo nft -f - << EOF
+        define testflinger-ips = { $(dig +short $SERVER | grep -o '[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}' | tr '\n' ',') }
```
mihalicyn, line 45: nit: you can also take a look on this https://github.com/canonical/lxd-ci/blob/f329b8464f807ba3852371a83564a687f0f01ff4/.github/actions/set-aproxy-bypass-for-testflinger/action.yml#L12

we haven't used any extra dependencies like `dig`.

boukeas: That's very useful, I'll modify and re-test.

pedro-avalos (dismissed):
This looks good to me, thanks for working on this George!

pedro-avalos (approved):

PR #680, opened 2025-05-21
feat(cli): Add enhancements for testflinger-cli to retrieve agent/queues status.

### LINKED ISSUE #442: CLI command to return the agent status
Our workflow has us allocating jobs to a pool of queues (all maas servers with their own agent), with the ability to failover to the next device in the pool if one can't be provisioned. However, this doesn't work when the agent is in an offline state. We can't tell from the testflinger output if the agent is offline or just busy servicing an existing job. The outcome is that we end up waiting for hours before timing out.

In this case, it would be helpful to know the status of the agent (offline, waiting, etc), so that we can skip it when offline.

For example:
```
# testflinger agent-status
usage: testflinger agent-status [-h] agent
# testflinger agent-status lubba
offline
```

### LINKED ISSUE #443: CLI command to return the queue status and length
Our workflow has us allocating jobs to a pool of queues (all maas servers with their own queue). As we have a pool of servers, we want to load balance the jobs if possible. Currently we can only see the length of the queue only after submitting a request. We then have to immediately cancel the job if the queue is too long. Also, the testflinger output only indicates how many jobs are queued. As a result, if the device is idle or in the middle of a job, the queue length is 0. There is no way to determine if the queue has active jobs that I'll end up having to wait on.

What would be helpful is a way to return the status of the queue and how many jobs are waiting. For example:

```
# testflinger queue-status
usage: testflinger queue-status [-h] queue
# testflinger agent-status hidon
available agents: 1
busy agents: 1
jobs queued: 2
```

rene-oromtz:
## Description

There are two enhancements added to testflinger-cli to be able to retrieve the agent status (`testflinger-cli agent-status`) and the queue-status (`testflinger-cli queue-status`). 

## Resolved issues
Resolves [CERTTF-476](https://warthogs.atlassian.net/browse/CERTTF-476)
Resolves [CERTTF-477](https://warthogs.atlassian.net/browse/CERTTF-477)
Resolves #442 
Resolves #443 

## Web service API changes

New endpoint was added for retrieving the queue status. specifically, to be able to retrieve the amount of jobs that are waiting in queue. 

Current endpoint `"/job/<job_id>/position"` will return the position of the job only after creating the job. 

New endpoint `/job/queues/<queue_name>` will return a JSON with the status, ID and creation timestamp for all jobs assigned in a queue. 

```
curl http://localhost:5000/v1/job/queues/test_queue6
[{"job_id": "80a416a6-238b-4879-970e-09c9125fd2b9", "created_at": {"$date": "2025-05-16T20:44:32.418Z"}, "job_state": "waiting"}]
```

Then, the `queue-status` option is filtering only those that are "waiting".  

## Tests

This was tested locally by creating a local snap and validate the desired functionality. 

Agent Status:
```
testflinger-cli --server=http://localhost:5000 agent-status agent0
waiting

testflinger-cli --server=http://localhost:5000 agent-status agent9
provision
```

Queue Status::

```
testflinger-cli --server=http://localhost:5000 queue-status test_queue0
Total agents in queue: 3
Available:             1
Busy:                  2
Offline:               0
Jobs waiting:          2

testflinger-cli --server=http://localhost:5000 queue-status test_queue9
Total agents in queue: 0
Available:             0
Busy:                  0
Offline:               0
Jobs waiting:          3
```

[CERTTF-476]: https://warthogs.atlassian.net/browse/CERTTF-476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CERTTF-477]: https://warthogs.atlassian.net/browse/CERTTF-477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -46,7 +46,11 @@ from testflinger_cli import (
     helpers,
     history,
 )
-from testflinger_cli.errors import AttachmentError, SnapPrivateFileError
+from testflinger_cli.errors import (
+    AttachmentError,
+    SnapPrivateFileError,
+    TFStatusError,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -369,13 +373,10 @@ class TestflingerCli:
 
     def agent_status(self):
         """Show the status of a specified agent."""
-        agent_status = self.get_agent_state(self.args.agent_name)
-        if agent_status["state"] == "unknown":
-            print(
-                "Unable to retrieve agent state from the server, check your "
-                "connection or try again later."
-            )
-            return
+        try:
+            agent_status = self.get_agent_state(self.args.agent_name)
+        except TFStatusError as exc:
+            sys.exit(exc)
 
         if self.args.json:
             output = json.dumps(
@@ -391,14 +392,11 @@ class TestflingerCli:
 
     def queue_status(self):
         """Show the status of the agents in a specified queue."""
-        queue_state = self.get_queue_state(self.args.queue_name)
-        jobs_queued = self.get_queued_jobs(self.args.queue_name)
-        if queue_state == "unknown":
-            print(
-                "Unable to retrieve queue state from the server, check your "
-                "connection or try again later."
-            )
-            return
+        try:
+            queue_state = self.get_queue_state(self.args.queue_name)
+            jobs_queued = self.get_queued_jobs(self.args.queue_name)
+        except TFStatusError as exc:
+            sys.exit(exc)
 
         if self.args.json:
             output = json.dumps(
@@ -1103,9 +1101,9 @@ class TestflingerCli:
             # For other types of network errors, or JSONDecodeError if we got
             # a bad return from get_agent_status()
             logger.debug("Unable to retrieve agent state: %s", exc)
-        return {"state": "unknown", "queues": []}
+        raise TFStatusError("agent")
 
-    def get_queue_state(self, queue_name: str) -> list[dict] | str:
+    def get_queue_state(self, queue_name: str) -> list[dict]:
         """Return the state of the agents from within a specified queue.
 
         :param queue_name: Queue name
@@ -1129,7 +1127,7 @@ class TestflingerCli:
             # For other types of network errors or JSONDecodeError if we got
             # a bad return from get_queue_status()
             logger.debug("Unable to retrieve queue state: %s", exc)
-        return "unknown"
+        raise TFStatusError("queue")
 
     def get_queued_jobs(self, queue_name: str) -> list:
         """Return the jobs waiting on a specified queue.
@@ -1157,4 +1155,4 @@ class TestflingerCli:
             # For other types of network errors or JSONDecodeError is we got
             # a bad return from get_jobs_in_queue()
             logger.debug("Unable to retrieve queue state: %s", exc)
-        return "unknown"
+        raise TFStatusError("job")
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -409,8 +409,8 @@ class TestflingerCli:
         else:
             agents = Counter(
                 (
-                    agent["state"]
-                    if agent["state"] in ("waiting", "offline")
+                    agent["status"]
+                    if agent["status"] in ("waiting", "offline")
                     else "busy"
                 )
                 for agent in queue_state
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -409,14 +409,14 @@ class TestflingerCli:
                 }
             )
         else:
-            agents = Counter()
-            for agent in queue_state:
-                status = (
+            agents = Counter(
+                (
                     agent["state"]
                     if agent["state"] in ("waiting", "offline")
                     else "busy"
                 )
-                agents[status] += 1
+                for agent in queue_state
+            )
 
             output = (
                 f"Agents in queue: {agents.total()}\n"
@@ -1074,12 +1074,12 @@ class TestflingerCli:
             logger.debug("Unable to retrieve job state: %s", exc)
         return "unknown"
 
-    def get_agent_state(self, agent_name):
+    def get_agent_state(self, agent_name: str) -> dict:
         """Return the state for the specified agent.
 
-        :param str agent_name: Agent Name
+        :param agent_name: Agent Name
         :raises SystemExit: Exit with HTTP error code
-        :return Dict: Agent status which includes the state and the queues
+        :return: Agent status which includes the state and the queues
         """
         try:
             agent_data = self.client.get_agent_data(agent_name)
@@ -1105,12 +1105,12 @@ class TestflingerCli:
             logger.debug("Unable to retrieve agent state: %s", exc)
         return "unknown"
 
-    def get_queue_state(self, queue_name):
+    def get_queue_state(self, queue_name: str) -> dict:
         """Return the state of the agents from within a specified queue.
 
-        :param str queue_name: Queue name
+        :param queue_name: Queue name
         :raises SystemExit: Exit with HTTP error code
-        :return Dict: Agent status from specified queue.
+        :return: Agent status from specified queue.
         """
         try:
             return self.client.get_agent_status_by_queue(queue_name)
@@ -1131,12 +1131,12 @@ class TestflingerCli:
             logger.debug("Unable to retrieve queue state: %s", exc)
         return "unknown"
 
-    def get_queued_jobs(self, queue_name):
+    def get_queued_jobs(self, queue_name: str) -> list:
         """Return the jobs waiting on a specified queue.
 
-        :param str queue_name: Queue name
+        :param queue_name: Queue name
         :raises SystemExit: Exit with HTTP error code
-        :return list: List with the job ids for the waiting jobs
+        :return: List with the job ids for the waiting jobs
         """
         try:
             jobs_waiting = [
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -140,6 +140,7 @@ class TestflingerCli:
         self._add_reserve_args(subparsers)
         self._add_status_args(subparsers)
         self._add_agent_status_args(subparsers)
+        self._add_queue_status_args(subparsers)
         self._add_results_args(subparsers)
         self._add_show_args(subparsers)
         self._add_submit_args(subparsers)
@@ -284,6 +285,15 @@ class TestflingerCli:
         parser.set_defaults(func=self.agent_status)
         parser.add_argument("agent_name")
 
+    def _add_queue_status_args(self, subparsers):
+        """Command line arguments for queue status."""
+        parser = subparsers.add_parser(
+            "queue-status",
+            help="Show the status of the agents in a specified QUEUE_NAME",
+        )
+        parser.set_defaults(func=self.queue_status)
+        parser.add_argument("queue_name")
+
     def _add_results_args(self, subparsers):
         """Command line arguments for results."""
         parser = subparsers.add_parser(
@@ -358,6 +368,23 @@ class TestflingerCli:
         else:
             print(
                 "Unable to retrieve agent state from the server, check your "
+                "connection or try again later."
+            )
+
+    def queue_status(self):
+        """Show the status of the agents in a specified QUEUE_NAME."""
+        queue_state = self.get_queue_state(self.args.queue_name)
+        jobs_queued = self.get_queued_jobs(self.args.queue_name)
+        if queue_state != "unknown":
+            self.history.update(self.args.queue_name, queue_state)
+            print("Total agents in queue: {}".format(queue_state["total"]))
+            print("Available:             {}".format(queue_state["waiting"]))
+            print("Busy:                  {}".format(queue_state["busy"]))
+            print("Offline:               {}".format(queue_state["offline"]))
+            print("Jobs waiting:          {}".format(jobs_queued))
+        else:
+            print(
+                "Unable to retrieve queue state from the server, check your "
                 "connection or try again later."
             )
 
@@ -1015,7 +1042,6 @@ class TestflingerCli:
         :raises SystemExit: Exit with HTTP error code
         :return str: Agent state
         """
-
         try:
             return self.client.get_agent_status(agent_name)
         except client.HTTPError as exc:
@@ -1026,11 +1052,66 @@ class TestflingerCli:
                 )
             if exc.status == 404:
                 sys.exit(
-                    "Received 404 error from server. Are you"
+                    "Received 404 error from server. Are you "
                     "sure this is a testflinger server?"
                 )
         except (IOError, ValueError) as exc:
             # For other types of network errors, or JSONDecodeError if we got
-            # a bad return from get_status()
-            logger.debug("Unable to retrieve job state: %s", exc)
+            # a bad return from get_agent_status()
+            logger.debug("Unable to retrieve agent state: %s", exc)
+        return "unknown"
+
+    def get_queue_state(self, queue_name):
+        """Return the state of the agents from within a specified queue.
+
+        :param str queue_name: Queue name
+        :raises SystemExit: Exit with HTTP error code
+        :return Dict: Agent states from specified queue.
+        """
+        try:
+            return self.client.get_queue_status(queue_name)
+        except client.HTTPError as exc:
+            if exc.status == 204:
+                sys.exit(
+                    "Invalid queue specified. Check the queue name "
+                    "to make sure its correct."
+                )
+            if exc.status == 404:
+                sys.exit(
+                    "Received 404 error from server. Are you "
+                    "sure this is a testflinger server?"
+                )
+        except (IOError, ValueError) as exc:
+            # For other types of network errors or JSONDecodeError if we got
+            # a bad return from get_queue_status()
+            logger.debug("Unable to retrieve queue state: %s", exc)
+        return "unknown"
+
+    def get_queued_jobs(self, queue_name):
+        """Return the number of jobs waiting in queue.
+        :param str queue_name: Queue name
+        :raises SystemExit: Exit with HTTP error code
+        :return int: Number of jobs waiting in queue.
+        """
+        try:
+            jobs_waiting = len(
+                [
+                    job
+                    for job in self.client.get_jobs_in_queue(queue_name)
+                    if job["job_state"] == "waiting"
+                ]
+            )
+            return jobs_waiting
+        except client.HTTPError as exc:
+            if exc.status == 204:
+                return 0
+            if exc.status == 404:
+                sys.exit(
+                    "Received 404 error from server. Are you "
+                    "sure this is a testflinger server?"
+                )
+        except (IOError, ValueError) as exc:
+            # For other types of network errors or JSONDecodeError is we got
+            # a bad return from get_jobs_in_queue()
+            logger.debug("Unable to retrieve queue state: %s", exc)
         return "unknown"
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -409,6 +409,20 @@ class TestflingerCli:
                 queue_status = self.client.get_agent_status_by_queue(
                     self.args.queue_name
                 )
+            except client.HTTPError as exc:
+                if exc.status == HTTPStatus.NO_CONTENT:
+                    sys.exit("No agent serving to specified queue")
+                if exc.status == HTTPStatus.NOT_FOUND:
+                    sys.exit("Specified queue does not exists")
+                # If any other HTTP error, raise UnknownStatusError
+                raise UnknownStatusError("queue") from exc
+            except (IOError, ValueError) as exc:
+                # For other types of network errors or JSONDecodeError
+                # if we got a bad return
+                logger.debug("Unable to retrieve agent state: %s", exc)
+                raise UnknownStatusError("queue") from exc
+
+            try:
                 jobs_queued = [
                     job["job_id"]
                     for job in self.client.get_jobs_on_queue(
@@ -418,16 +432,16 @@ class TestflingerCli:
                 ]
             except client.HTTPError as exc:
                 if exc.status == HTTPStatus.NO_CONTENT:
-                    sys.exit("No agent serving to specified queue")
-                if exc.status == HTTPStatus.NOT_FOUND:
-                    sys.exit("Specified queue does not exists")
-                # If any other HTTP error, raise UnknownStatusError
-                raise UnknownStatusError("agent") from exc
+                    jobs_queued = []
+                else:
+                    # If any other HTTP error, raise UnknownStatusError
+                    raise UnknownStatusError("job") from exc
             except (IOError, ValueError) as exc:
                 # For other types of network errors or JSONDecodeError
                 # if we got a bad return
-                logger.debug("Unable to retrieve queue state: %s", exc)
-                raise UnknownStatusError("queue") from exc
+                logger.debug("Unable to retrieve job state: %s", exc)
+                raise UnknownStatusError("job") from exc
+
         except UnknownStatusError as exc:
             sys.exit(exc)
 
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -27,6 +27,7 @@ import tarfile
 import tempfile
 import time
 from argparse import ArgumentParser
+from collections import Counter
 from datetime import datetime, timezone
 from functools import partial
 from http import HTTPStatus
@@ -281,7 +282,7 @@ class TestflingerCli:
     def _add_agent_status_args(self, subparsers):
         """Command line arguments for agent status."""
         parser = subparsers.add_parser(
-            "agent-status", help="Show the status of a specified AGENT_NAME"
+            "agent-status", help="Show the status of a specified agent"
         )
         parser.set_defaults(func=self.agent_status)
         parser.add_argument("agent_name")
@@ -293,7 +294,7 @@ class TestflingerCli:
         """Command line arguments for queue status."""
         parser = subparsers.add_parser(
             "queue-status",
-            help="Show the status of the agents in a specified QUEUE_NAME",
+            help="Show the status of the agents in a specified queue",
         )
         parser.set_defaults(func=self.queue_status)
         parser.add_argument("queue_name")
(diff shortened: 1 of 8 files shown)


server/src/testflinger/api/v1.py
```diff
@@ -378,6 +378,34 @@ def search_jobs(query_data):
...
+        String with the queue name where to perform the query.
+    :return:
+        JSON data with the jobs allocated to the specified queue.
+    """
+    response = list(
+        database.mongo.db.jobs.find({"job_data.job_queue": queue_name})
+    )
+
+    if not response:
+        return [], 204
```
pedro-avalos, line 395: Same as before, can you use `http` module to make this more readable?

server/src/testflinger/api/v1.py
```diff
@@ -378,6 +378,34 @@ def search_jobs(query_data):
...
+    """Get the jobs in a specified queue along with its state.
+
+    :param queue_name
+        String with the queue name where to perform the query.
+    :return:
+        JSON data with the jobs allocated to the specified queue.
+    """
+    response = list(
+        database.mongo.db.jobs.find({"job_data.job_queue": queue_name})
+    )
```
pedro-avalos, line 392: Let's rename this from `response` to `jobs`

server/src/testflinger/api/v1.py
```diff
@@ -378,6 +378,34 @@ def search_jobs(query_data):
     return jsonify(list(jobs))
 
 
+@v1.get("/job/queues/<queue_name>")
```
pedro-avalos, line 381: Hm, I think `/queues/<queue_name>/jobs` makes more sense as an API endpoint

server/API.md
```diff
@@ -83,6 +83,31 @@ curl 'http://localhost:8000/v1/job/search?tags=foo&tags=bar&match=all'
 
 This will find jobs tagged with both "foo" and "bar".
 
+## `[GET] /v1/job/queues/<queue_name>`
```
pedro-avalos, line 86: Same as before, I think `/queues/<queue_name>/jobs` makes more sense.

cli/testflinger_cli/client.py
```diff
@@ -139,6 +139,54 @@ def get_status(self, job_id):
...
+
+            # Determine the status of each agent in queue
+            if agent["state"] == "waiting":
+                queue_status["waiting"] += 1
+            elif agent["state"] == "offline":
+                queue_status["offline"] += 1
+            else:
+                queue_status["busy"] += 1
+
+        return queue_status
```
pedro-avalos, line 188: This seems like a job for a [`Counter`](https://docs.python.org/3/library/collections.html#collections.Counter). For example:

```python
c = Counter()
# not sure if you need to initialize it with the values, if you find that you need to:
# c = Counter(["busy", "waiting", "offline"])

for agent in data:
    status = agent["state"] if agent["state"] in ("waiting", "offline") else "busy"
    c[status] += 1

c.total()
```

pedro-avalos (changes requested):
Some suggestions regarding the new CLI output formats and new API endpoint.

I've also left a recommendation to simplify the counting functionality.

rene-oromtz:
All suggested changes have been addressed, most important is that the new API endpoint added is now /queues/<queue_name>/jobs. 

The json output is now also included:

```
testflinger-cli --server=http://localhost:5000 queue-status test_queue5 --json |jq
{
  "total_agents": 2,
  "available_agents": 0,
  "busy_agents": 2,
  "offline_agents": 0,
  "jobs_waiting": 0
}
```

```
testflinger-cli --server=http://localhost:5000 agent-status agent0 --json | jq
{
  "agent": "agent0",
  "status": "waiting"
}
```

cli/testflinger_cli/client.py
boukeas: `Counter` can be initialized using an iterable:
```
counter = Counter(state if (state := agent["state"]) in {"waiting", "offline"} else "busy" for agent in data)
counter.total()
```

cli/testflinger_cli/client.py
```diff
@@ -139,6 +140,49 @@ def get_status(self, job_id):
...
 
+    def get_agent_status(self, agent_name):
+        """Get the status of a specified agent.
+
+        :param agent_name
+            Name of the agent to retrieve status from
+        :return
+            String containing the agent_state for the specified agent.
+        """
+        endpoint = "/v1/agents/data"
```
boukeas, line 151: Doesn't the Testflinger API have an endpoint to get the status of one specific agent? Do we need to iterate over all of them to get what we need?

cli/testflinger_cli/client.py
```diff
@@ -139,6 +140,49 @@ def get_status(self, job_id):
...
+    def get_queue_status(self, queue_name):
+        """Get the status of a specified queue.
+
+        :param queue_name
+            Name of the queue to retrieve its agent status
+        :return
+            collections.Counter with the status of the agents
+            for the specified queue.
+        """
+        endpoint = "/v1/queues/{}/agents".format(queue_name)
```
boukeas, line 171: Any particular reason for not using an f-string here?

cli/testflinger_cli/__init__.py
```diff
@@ -395,6 +421,53 @@ def status(self):
...
+        if queue_state != "unknown":
+            if self.args.json:
+                print(
+                    json.dumps(
+                        {
+                            "total_agents": queue_state.total(),
+                            "available_agents": queue_state["waiting"],
+                            "busy_agents": queue_state["busy"],
+                            "offline_agents": queue_state["offline"],
+                            "jobs_waiting": jobs_queued,
```
boukeas, line 455: I'd like to debate the results we are returning here. I am wondering if it might be even more useful for the user to return the names of actual agents and the IDs of actual jobs, instead of their numbers. Besides, if they have the agent data, they can always easily get their numbers. If I was doing this from scratch, I would probably go for this sort of JSON response:
```
{
  "queue": <queue_name>,  # never hurts to include this
  "jobs": <list of job-ids>,
  "agents": [
    {
      "name": <agent_name>
      "status": <agent_status>
    }
}
```

server/src/testflinger/api/v1.py
```diff
@@ -835,6 +836,34 @@ def get_agents_on_queue(queue_name):
...
+    )
+
+    if not jobs:
+        return [], HTTPStatus.NO_CONTENT
+
+    jobs_in_queue = [
+        {
+            "job_id": job.get("job_id"),
+            "created_at": job.get("created_at"),
+            "job_state": job.get("result_data", {}).get("job_state"),
```
boukeas, line 859: Using `get` means that if a job doesn't have values for these fields then we simply use `None`. But I am wondering:
- Is there any chance a job won't have a value for these fields?
- If it doesn't, should we actually fail (raise an exception) instead of silently using `None`?

server/src/testflinger/api/v1.py
```diff
@@ -835,6 +836,34 @@ def get_agents_on_queue(queue_name):
...
+def get_jobs_by_queue(queue_name):
+    """Get the jobs in a specified queue along with its state.
+
+    :param queue_name
+        String with the queue name where to perform the query.
+    :return:
+        JSON data with the jobs allocated to the specified queue.
+    """
+    jobs = list(
+        database.mongo.db.jobs.find({"job_data.job_queue": queue_name})
```
boukeas, line 849: I am opposed to these functions being tightly coupled with MongoDB. I know it's not your fault, I know this is how most other API functions do this but in the new ones we introduced functions in the `database` module (e.g. `get_agents_on_queue`) that abstract the way the fact that we use MongoDB as a database. Think of the `database` module as an abstraction layer, a database client that could be swapped with another. So my suggestion would be to introduce something like `database.get_jobs_in_queue` and use that.

cli/testflinger_cli/client.py
boukeas: Alright, so we could extend the `v1/agents/data` [GET endpoint](https://github.com/canonical/testflinger/blob/6142e53b9ad2b947cbde55d41384a9bbbfb9f77c/server/src/testflinger/api/v1.py#L638) to accept a query parameter _or_ introduce `v1/agents/data/<agent_name>` since there is already a [corresponding POST endpoint](https://github.com/canonical/testflinger/blob/6142e53b9ad2b947cbde55d41384a9bbbfb9f77c/server/src/testflinger/api/v1.py#L646).

cli/testflinger_cli/client.py
rene-oromtz: Not really, I just tried to be consistent with the `get_status` function I used for reference:
```
def get_status(self, job_id):
        ...
        endpoint = "/v1/result/{}".format(job_id)
```

That being said, I was looking at this `client.py` and seems there is no standardization on which formatting to use for strings as there are some that use .format(), f-string and even one function is using a concatenation. 

If preferred I can use the f-string for now and consider in the future create a task for standardizing the code.

cli/testflinger_cli/__init__.py
rene-oromtz: I like this idea! Specially since we are using this JSON format it can be really useful for the user to get additional information about the agent/jobs in the queue. I can modify the JSON output to look something like this. 

For the human readable format, maybe we can keep the output as it current is. Just in case the user only want to retrieve an overall information of the queue.

server/src/testflinger/api/v1.py
rene-oromtz: I don't think there is, AFAIK once the job is created, it will add the ID, timestamp and set the initial status to "waiting". I don't think there is a reason why any of those values are missing on a job. 

I can add an exception to handle this but ideally, we shouldn't hit any case. I'm guessing that if there is a problem to set any of those values, it will be caught during job creation.

server/src/testflinger/api/v1.py
rene-oromtz: Sounds good, I will keep it in mind and add this function in the `database` module. Again, it would be nice if we eventually add a standardization task to be able to get all of testflinger code consistent.

cli/testflinger_cli/client.py
rene-oromtz: This has been added to #687, once approved, we can use GET /agents/data/<agent_name>

server/src/testflinger/api/v1.py
rene-oromtz: This was covered in #689 based on your suggestions to use the database module as the abstraction layer. 
It was integrated as:
`jobs = database.get_jobs_on_queue(queue_name)`

server/src/testflinger/api/v1.py
rene-oromtz: This was also covered in #689 and an exception was included in case the server is unable t retrieve any of these keys.

pedro-avalos (changes requested):
Some formatting questions/suggestions. I also noticed we are not using the new `/v1/agents/data/<agent>` endpoint.

rene-oromtz:
Yes, sorry I rebased to work with latest changes but in next commit I'll use the new endpoint and complete the additional modifications.

rene-oromtz:
Thanks for the suggestion @boukeas @pedro-avalos 
Based on your inputs and with the recent additions to the endpoint in server, I included the required modifications. 

The output for the cli command was slightly changed to reflect the suggestions, specifically for the json output

### Agent Status

Human readable:
```
testflinger-cli --server=http://localhost:5000 agent-status test-agent-1
offline
```

JSON:
```
testflinger-cli --server=http://localhost:5000 agent-status test-agent-1 --json |jq
{
  "agent": "test-agent-1",
  "status": "offline",
  "queues": [
    "test"
  ]
}
```

### Queue Status

Human readable:
```
testflinger-cli --server=http://localhost:5000 queue-status test_queue0
Agents in queue: 2
Available:       0
Busy:            2
Offline:         0
Jobs waiting:    1
```

JSON:
```
testflinger-cli --server=http://localhost:5000 queue-status test_queue0 --json | jq
{
  "queue": "test_queue0",
  "jobs waiting": [
    "f445c4c8-48e6-4199-9587-403716cf539e"
  ],
  "agents": [
    {
      "name": "agent5",
      "state": "test"
    },
    {
      "name": "agent6",
      "state": "test"
    }
  ]
}
```

cli/testflinger_cli/__init__.py
```diff
@@ -1099,3 +1186,88 @@ def get_job_state(self, job_id):
...
             logger.debug("Unable to retrieve job state: %s", exc)
         return "unknown"
+
+    def get_agent_state(self, agent_name):
+        """Return the state for the specified agent.
+
+        :param str agent_name: Agent Name
+        :raises SystemExit: Exit with HTTP error code
+        :return Dict: Agent status which includes the state and the queues
+        """
```
pedro-avalos, line 1196: Can you move the type hints to the function definition?

```suggestion
    def get_agent_state(self, agent_name: str) -> dict:
        """Return the state for the specified agent.

        :param agent_name: Agent Name
        :raises SystemExit: Exit with HTTP error code
        :return: Agent status which includes the state and the queues
        """
```

cli/testflinger_cli/__init__.py
```diff
@@ -1099,3 +1186,88 @@ def get_job_state(self, job_id):
...
+            logger.debug("Unable to retrieve agent state: %s", exc)
+        return "unknown"
+
+    def get_queue_state(self, queue_name):
+        """Return the state of the agents from within a specified queue.
+
+        :param str queue_name: Queue name
+        :raises SystemExit: Exit with HTTP error code
+        :return Dict: Agent status from specified queue.
+        """
```
pedro-avalos, line 1227:
```suggestion
    def get_queue_state(self, queue_name: str) -> dict:
        """Return the state of the agents from within a specified queue.

        :param queue_name: Queue name
        :raises SystemExit: Exit with HTTP error code
        :return: Agent status from specified queue.
        """
```

cli/testflinger_cli/__init__.py
```diff
@@ -1099,3 +1186,88 @@ def get_job_state(self, job_id):
...
+            logger.debug("Unable to retrieve queue state: %s", exc)
+        return "unknown"
+
+    def get_queued_jobs(self, queue_name):
+        """Return the jobs waiting on a specified queue.
+
+        :param str queue_name: Queue name
+        :raises SystemExit: Exit with HTTP error code
+        :return list: List with the job ids for the waiting jobs
+        """
```
pedro-avalos, line 1253:
```suggestion
    def get_queued_jobs(self, queue_name: str) -> list:
        """Return the jobs waiting on a specified queue.

        :param queue_name: Queue name
        :raises SystemExit: Exit with HTTP error code
        :return: List with the job ids for the waiting jobs
        """
```

cli/testflinger_cli/client.py
```diff
@@ -139,6 +139,33 @@ def get_status(self, job_id):
...
         return data.get("job_state")
 
+    def get_agent_data(self, agent_name):
+        """Get all the data from a specified agent.
+
+        :param agent_name
+            Name of the agent to retrieve its data
+        :return
+            Dict containing all the data from the agent.
+        """
```
pedro-avalos, line 149:
```suggestion
    def get_agent_data(self, agent_name: str) -> dict:
        """Get all the data from a specified agent.

        :param agent_name: Name of the agent to retrieve its data
        :return: Dict containing all the data from the agent.
        """
```

cli/testflinger_cli/client.py
```diff
@@ -139,6 +139,33 @@ def get_status(self, job_id):
...
+        return json.loads(self.get(endpoint))
+
+    def get_agent_status_by_queue(self, queue):
+        """Get the status of the agents by a specified queue.
+
+        :param queue
+            Name of the queue to retrieve its agent status
+        :return
+            Dict with the name and status for each agent in queue.
+        """
```
pedro-avalos, line 160:
```suggestion
    def get_agent_status_by_queue(self, queue: str) -> dict:
        """Get the status of the agents by a specified queue.

        :param queue: Name of the queue to retrieve its agent status
        :return: Dict with the name and status for each agent in queue.
        """
```

cli/testflinger_cli/client.py
```diff
@@ -291,3 +318,13 @@ def get_agents_on_queue(self, queue):
         endpoint = f"/v1/queues/{queue}/agents"
         data = self.get(endpoint)
         return json.loads(data)
+
+    def get_jobs_on_queue(self, queue):
+        """Get the list of jobs assigned to a queue.
+
+        :param queue
+            String with the queue name where to list the jobs
+        """
```
pedro-avalos, line 327:
```suggestion
    def get_jobs_on_queue(self, queue: str) -> dict:
        """Get the list of jobs assigned to a queue.

        :param queue: The queue name to list the jobs for
        """
```

cli/testflinger_cli/__init__.py
```diff
@@ -400,6 +427,66 @@ def status(self):
...
+        else:
+            agents = Counter()
+            for agent in queue_state:
+                status = (
+                    agent["state"]
+                    if agent["state"] in ("waiting", "offline")
+                    else "busy"
+                )
+                agents[status] += 1
+
```
pedro-avalos, line 480: I think this can be simplified

```suggestion
            agents = Counter(
                agent["state"]
                if agent["state"] in ("waiting", "offline")
                else "busy"
            )
```

pedro-avalos (changes requested):
I realize a lot of the functions don't have type-hints, but let's add those to any new functions.

cli/testflinger_cli/__init__.py
```diff
@@ -1099,3 +1186,88 @@ def get_job_state(self, job_id):
...
+            if exc.status == HTTPStatus.NOT_FOUND:
+                sys.exit(
+                    "Received 404 error from server. Are you "
+                    "sure this is a testflinger server?"
+                )
+        except (IOError, ValueError) as exc:
+            # For other types of network errors, or JSONDecodeError if we got
+            # a bad return from get_agent_status()
+            logger.debug("Unable to retrieve agent state: %s", exc)
+        return "unknown"
```
pedro-avalos, line 1219: Hm, The return type is `dict`, should this be:

`{"state": "unknown", queues: []}`

pedro-avalos (changes requested):
This looks great! Just one last issue I just noticed because of the type hints.

pedro-avalos (changes requested):
Last few issues with the different return types. I think we should either raise an error or `sys.exit` on these functions, let me know what you think

cli/testflinger_cli/__init__.py
```diff
@@ -1099,3 +1186,88 @@ def get_job_state(self, job_id):
...
+            if exc.status == HTTPStatus.NOT_FOUND:
+                sys.exit(
+                    "Received 404 error from server. Are you "
+                    "sure this is a testflinger server?"
+                )
+        except (IOError, ValueError) as exc:
+            # For other types of network errors or JSONDecodeError is we got
+            # a bad return from get_jobs_in_queue()
+            logger.debug("Unable to retrieve queue state: %s", exc)
+        return "unknown"
```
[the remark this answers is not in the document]
rene-oromtz: I'm okay raising the exception, In any case, if the "unknown" is found, we are just printing a message we can raise the exception and return that message

rene-oromtz:
@pedro-avalos I included a custom exception for handling retrieving status and applied it to all the new functions. Let me know that you think!

cli/testflinger_cli/__init__.py
rene-oromtz: Based on your suggestion, I included the following output:
testflinger-cli --server=http://localhost:5000 queue-status test_queue0 --json | jq
```
{
  "queue": "test_queue0",
  "jobs waiting": [
    "f445c4c8-48e6-4199-9587-403716cf539e"
  ],
  "agents": [
    {
      "name": "agent5",
      "state": "test"
    },
    {
      "name": "agent6",
      "state": "test"
    }
  ]
}
```

cli/testflinger_cli/errors.py
```diff
@@ -17,3 +17,15 @@ def __init__(self, path: PathLike[str]) -> None:
...
+class TFStatusError(Exception):
+    """Exception thrown when unable to retrieve status
+    information from TF Server.
+    """
+
+    def __init__(self, endpoint: str) -> None:
+        super().__init__(
+            f"Unable to retrieve {endpoint} status from the server, check "
+            "your connection or try again later."
+        )
```
pedro-avalos, line 31: Love this, can we rename it to `UnknownStatusError`, that name is more descriptive of what the exception actually is.

I also recommend rephrasing the docstring to make it fit within the first line.

```suggestion
class UnknownStatusError(Exception):
    """Exception thrown when unable to retrieve a status from the server."""

    def __init__(self, endpoint: str) -> None:
        super().__init__(
            f"Unable to retrieve {endpoint} status from the server, check "
            "your connection or try again later."
        )
```

pedro-avalos (changes requested):
Great! Just one last (I swear) nitpick on the naming of the error; with that, I'll give it a +1

rene-oromtz:
All suggestions are now included :P

pedro-avalos (dismissed):
Thank you so much for the time you put into this; this is fantastic work. +1 from me!

cli/testflinger_cli/client.py
```diff
@@ -139,6 +139,29 @@ def get_status(self, job_id):
         data = json.loads(self.get(endpoint))
         return data.get("job_state")
 
+    def get_agent_data(self, agent_name: str) -> dict:
+        """Get all the data from a specified agent.
+
+        :param agent_name: Name of the agent to retrieve its data
+        :return: Dict containing all the data from the agent.
```
boukeas, line 146: This is minor but I think that using "from" here may be misleading: we are retrieving the data "for" the agent (and we're actually retrieving it from the server). So my suggestion is to replace "from" with "for".

cli/testflinger_cli/client.py
```diff
@@ -139,6 +139,29 @@ def get_status(self, job_id):
...
+        """
+        endpoint = f"/v1/agents/data/{agent_name}"
+        return json.loads(self.get(endpoint))
+
+    def get_agent_status_by_queue(self, queue: str) -> list[dict]:
```
boukeas, line 151: I am sorry but I really really think using "status" and "state" interchangeably is confusing. If this is not deliberate (if there isn't a reason behind it) I would really like to use just one of the two, preferably the "legacy" option, i.e. the one that is already in use. This applies throughout the PR, there are multiple places where this occurs.

cli/testflinger_cli/__init__.py
```diff
@@ -1099,3 +1184,88 @@ def get_job_state(self, job_id):
...
+            ]
+            return jobs_waiting
+        except client.HTTPError as exc:
+            if exc.status == HTTPStatus.NO_CONTENT:
+                return []
+            if exc.status == HTTPStatus.NOT_FOUND:
+                sys.exit(
+                    "Received 404 error from server. Are you "
+                    "sure this is a testflinger server?"
+                )
```
boukeas, line 1266: Why is a 404 from the server an indication that this might not be a Testflinger server? What would the server return if there is simply no such queue?

cli/testflinger_cli/__init__.py
```diff
@@ -1099,3 +1184,88 @@ def get_job_state(self, job_id):
...
+        :return: Agent status from specified queue.
+        """
+        try:
+            return self.client.get_agent_status_by_queue(queue_name)
+        except client.HTTPError as exc:
+            if exc.status == HTTPStatus.NO_CONTENT:
+                sys.exit(
+                    "Invalid queue specified. Check the queue name "
+                    "to make sure its correct."
+                )
```
boukeas, line 1233: If I contact an endpoint to request the statuses of the agents serving a specific queue and the result is `HTTPStatus.NO_CONTENT` I will assume that there are no agents serving the queue. On the other hand, if I get a `HTTPStatus.NOT_FOUND`, I will assume there is no such queue. If this is indeed how our endpoint behaves then I think this needs to be addressed in the server.

cli/testflinger_cli/__init__.py
```diff
@@ -1099,3 +1184,88 @@ def get_job_state(self, job_id):
             # a bad return from get_status()
             logger.debug("Unable to retrieve job state: %s", exc)
         return "unknown"
+
+    def get_agent_state(self, agent_name: str) -> dict:
```
boukeas, line 1188: So right now we have e.g. `TestflingerCli.agent_status` which calls `TestflingerCli.get_agent_state` which calls `Client.get_agent_data`. My question is: do we really need the intermediate abstraction layer, i.e. do we need `TestflingerCli.get_agent_state`? I am not suggesting we necessarily do away with it, I am just trying to understand if it is necessary or if it's functionality could be integrated into the two other methods. @pedro-avalos what do you think?

boukeas (changes requested):

cli/testflinger_cli/__init__.py
rene-oromtz: Yes, right now I believe we would only get the NO_CONTENT response from the server. 
This makes sense, I would first need to make necessary changes on the server to return 404 (NOT FOUND) if the queue is non existent. 
Will work on this, should be a very small PR.

cli/testflinger_cli/__init__.py
rene-oromtz: The current design for other CLI commands use this abstraction layer and that's why I continue using it for consistency purposes. 

I took the `status` command as an example:

TesflingerCLI.status: Prints output. This layer gets the status from TestflingerCLI.get_job_state()
TestflingerCLI.get_job_state: Gets the jobs state and handles errors. Gets the job state from client.get_status()
Client.get_status: Gets the job state from the server. 

In most CLI commands the intermediate layer is being used for handling any potential errors during data retrieval and the first layer is just a print function but I don't see why the first two steps can't be joined together. 

If we agree this is the way to go, I can remove the intermediate layer and integrate all from a single function. 

Thoughts @pedro-avalos?

pedro-avalos: I think the intermediate layer makes sense if, for example, we need to get the agent state in multiple different functions. With the `get_job_state`, I think we use it for the `status` command and for the `jobs` command, so that makes sense. If we are not using the agent state in multiple functions like this, we can combine them for now

rene-oromtz:
After rebasing with the latest changes and by removing the intermediate layer, this PR is ready for a review:

### Agent Status
Human readable:
```
testflinger-cli --server=http://localhost:5000 agent-status agent1
test
```

JSON format:
```
testflinger-cli --server=http://localhost:5000 agent-status agent1 --json
{"agent": "agent1", "status": "test", "queues": ["test_queue8"]}
```

If agent does not exists
```
testflinger-cli --server=http://localhost:5000 agent-status agent12
Specified agent does not exists.
```

### Queue status
Human readable:
```
testflinger-cli --server=http://localhost:5000 queue-status test_queue8
Agents in queue: 2
Available:       0
Busy:            2
Offline:         0
Jobs waiting:    1
```

JSON format:
```
testflinger-cli --server=http://localhost:5000 queue-status test_queue8 --json
{"queue": "test_queue8", "jobs waiting": ["472ab479-455f-463e-8107-e87b62708e84"], "agents": [{"name": "agent0", "status": "test"}, {"name": "agent1", "status": "test"}]}
```

Queue does not exists
```
testflinger-cli --server=http://localhost:5000 queue-status test_queue1000
Specified queue does not exists
```

If no agent serving to queue:
```
testflinger-cli --server=http://localhost:5000 queue-status test_queue2
No agent serving to specified queue
```

bladernr:
Slight Nit Picks based on your sample output above:

> Specified queue does not exists

That should be "does not exist"

> No agent serving to specified queue

Would make more sense saying "No agent is listening to that queue"

rene-oromtz:
@bladernr 
For the first comment, do you mean just if the queue or agent does not exist just return "Does not exist"? As follows:
```
testflinger-cli --server=http://localhost:5000 agent-status agent12
Does not exists.
```

```
testflinger-cli --server=http://localhost:5000 queue-status test_queue1000
Does not exists
```

@pedro-avalos @boukeas 
Besides above comments, anything else I should modify from current code? It also needs a rebase so I'm planning to include all suggested changes in next commit.

server/src/testflinger/database.py
```diff
@@ -362,4 +362,4 @@ def queue_exists(queue: str) -> bool:
     :param queue: Name of the queue to validate.
     :return: True if exists, False otherwise
     """
-    return mongo.db.agents.find_one({"queues": queue}) is not None
+    return mongo.db.queues.find_one({"name": queue}) is not None
```
rene-oromtz, line 365: This needs to be discarded, I thought the queues collection will be a more direct approach but now I realized this collection only shows the "advertised queues". 

The current approach for validating the queue based on the agents is also valid.

pedro-avalos (changes requested):
Found the typos that Jeff was referring to. I have left some phrasings based on his suggestions.

Issue #289, opened 2024-06-18
Provide CLI to retrieve status of a job or reservation


fginther:
Testflinger allows for a reservation to be created for a device. However, it is the necessary to poll the job and watch the output to know when the device is ready and how to connect to it. This request is to proved that information via a status command:

```
$ testflinger status <job_id>
{
  "queue": "rpi3bp",
  "device": "rpi3bp-001",
  "ip": "1.2.3.4",
  "serial": 1.2.3.5:3000",
  "status": "READY",
  "reservation_remaining": 500
}
```

status could be used to indicate if the device is still "WAITING", is "PROVISIONING", "READY" for use, or the reservation has "TERMINATED" (or other values). Other values might be useful, but these are the ones that come to mind currently.

rene-oromtz:
As mentioned above, device information has been included in the `results` command. 
Right after provisioning, the agent name and the device IP will be available with this command. 
Reference: 
https://canonical-testflinger.readthedocs-hosted.com/en/latest/tutorial/index.html#check-test-output

Sample output:
```
$ testflinger-cli results 2bac1457-0000-0000-0000-15f23f69fd39

{
    "device_info": {
        "agent_name": "agent1",
        "device_ip": "10.10.10.10"
    },
    "job_state": "reserve"
}
```
Agents in TOR were just updated so on next job executions this information should be available. Please feel free to reopen this issue if you encounter any issues.

PR #682, opened 2025-05-22
Feat(cli) Add --json parameter to submit command.

marosg42:
## Description

New `--json` parameter for reservation jobs. Instead of polling job status externally, polling happens in testflinger-cli and prints JSON output without blocking the process when machine is deployed.

## Tests

```
# testflinger submit --json testflinger-racah.yaml | jq .
ca38a7b1-cb12-4a27-9f68-71dce13f2cf4
Job state changed: unknown -> waiting
Job state changed: waiting -> setup
Job state changed: setup -> provision
Job state changed: provision -> reserve
Found IP: 10.241.2.26
{
  "rc": 0,
  "job_id": "ca38a7b1-cb12-4a27-9f68-71dce13f2cf4",
  "ip": "10.241.2.26",
  "job_state": "reserve"
}

# testflinger submit --json testflinger-geminus.yaml | jq .
7d6e0b8d-d442-4d19-948c-9e1ce68a2285
Job state changed: unknown -> waiting
Job state changed: waiting -> provision
Job state changed: provision -> complete
{
  "rc": 1,
  "job_id": "7d6e0b8d-d442-4d19-948c-9e1ce68a2285",
  "ip": "",
  "job_state": "complete"
}

```

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -21,6 +21,7 @@ import inspect
 import json
 import logging
 import os
+import re
 import sys
 import tarfile
 import tempfile
@@ -376,6 +377,11 @@ class TestflingerCli:
             default=None,
             help="Secret key to be used with client id for authentication",
         )
+        parser.add_argument(
+            "--json",
+            action="store_true",
+            help="Output agent name, job id and reserved IP in JSON format",
+        )
         relative = parser.add_mutually_exclusive_group()
         relative.add_argument(
             "--attachments-relative-to",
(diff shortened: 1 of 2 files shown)


boukeas (commented):
@marosg42 I am quite conflicted about this. I think the title doesn't reflect what the PR really introduces. One would think that a `--json` argument would simply return the same response one would normally receive but in a JSON format. Instead, this actually performs polling, displays custom status change messages (on `stderr` so that they can be separated from the JSON) and then displays JSON data about the job that is extracted by parsing the captured output. My opinion:
- Testflinger needs to return the JSON data you need directly from an endpoint, so that the user doesn't have to parse the output to get it. This is definitely a functionality gap that should be addressed. If we are introducing a solution to this, it shouldn't depend on parsing the output.
- This PR provides a solution to a very specific problem by combining two or three component solutions. I am concerned about such an approach, I don't think a generic tool can address all the requirements of its individual users by introducing custom functionality like this.

@pedro-avalos I would very much appreciate your opinion on this.

pedro-avalos:
Hm, I would argue the right way to get this behavior would be to have `submit` no poll, then your script or user wraps/loops `results` until the status is reserved. If the `status` blob doesn't have the necessary information, that's what the PR should address

pedro-avalos:
I think #289 is what we should address

PR #685, opened 2025-05-22
feat(cli): Extract parts from __init__ to helpers and consts

pedro-avalos:
## Description

#668 introduced `helpers.py` and `consts.py`. This PR moves more elements out of the bloated `__init__.py` into these modules.

This PR also simplifies some code by:

- using the ternary operator
- using `contextlib.suppress` where we just ignore an error
- using f-strings in more places
- using `x in dictionary` instead of `x in dictionary.keys()`

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -851,7 +851,7 @@ class TestflingerCli:
             timestamp = datetime.fromtimestamp(
                 jobdata.get("submission_time"), tz=timezone.utc
             )
-            queue = jobdata.get("queue")
+            queue = jobdata["queue"]
             print(f"{job_id} {job_state:9} {timestamp:%a %b %d %H:%M} {queue}")
         print()
 
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -57,21 +57,14 @@ def cli():
     """Generate the TestflingerCli instance and run it."""
     try:
         tfcli = TestflingerCli()
-        configure_logging()
+        logging.basicConfig(
+            level=logging.WARNING,
+            format=consts.LOG_FORMAT,
+            datefmt=consts.LOG_DATE_FORMAT,
+        )
         tfcli.run()
     except KeyboardInterrupt:
         sys.exit("Received KeyboardInterrupt")
-
-
-def configure_logging():
-    """Configure default logging."""
-    logging.basicConfig(
-        level=logging.WARNING,
-        format=(
-            "%(levelname)s: %(asctime)s %(filename)s:%(lineno)d -- %(message)s"
-        ),
-        datefmt="%Y-%m-%d %H:%M:%S",
-    )
 
 
 class TestflingerCli:
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -56,13 +56,13 @@ if os.path.exists(os.path.join(basedir, "setup.py")):
 
 def cli():
     """Generate the TestflingerCli instance and run it."""
+    tfcli = TestflingerCli()
+    logging.basicConfig(
+        level=logging.WARNING,
+        format=consts.LOG_FORMAT,
+        datefmt=consts.LOG_DATE_FORMAT,
+    )
     try:
-        tfcli = TestflingerCli()
-        logging.basicConfig(
-            level=logging.WARNING,
-            format=consts.LOG_FORMAT,
-            datefmt=consts.LOG_DATE_FORMAT,
-        )
         tfcli.run()
     except KeyboardInterrupt:
         sys.exit("Received KeyboardInterrupt")
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -35,7 +35,14 @@ import argcomplete
 import requests
 import yaml
 
-from testflinger_cli import autocomplete, client, config, helpers, history
+from testflinger_cli import (
+    autocomplete,
+    client,
+    config,
+    consts,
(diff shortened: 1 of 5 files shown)


cli/testflinger_cli/__init__.py
```diff
@@ -923,21 +848,16 @@ def jobs(self):
...
-                    job_state,
-                    datetime.fromtimestamp(
-                        jobdata.get("submission_time"), tz=timezone.utc
-                    ).strftime("%a %b %d %H:%M"),
-                    jobdata.get("queue"),
-                )
+            timestamp = datetime.fromtimestamp(
+                jobdata.get("submission_time"), tz=timezone.utc
             )
+            queue = jobdata.get("queue")
```
boukeas, line 854: If a job should always have a queue then I would prefer we used `jobdata["queue"]` here. Otherwise, using `get` will not fail when it should.

cli/testflinger_cli/__init__.py
```diff
@@ -953,43 +873,38 @@ def list_queues(self):
...
-        queue = self.args.queue or self._get_queue(queues)
-        if queue not in queues.keys():
-            print(
-                "WARNING: '{}' is not in the list of known queues".format(
-                    queue
-                )
-            )
+        queue = self.args.queue or helpers.prompt_for_queue(queues)
+        if queue not in queues:
+            print(f"WARNING: '{queue}' is not in the list of known queues")
```
boukeas, line 888: Shouldn't we be logging instead of printing here?

cli/testflinger_cli/__init__.py
```diff
@@ -953,43 +873,38 @@ def list_queues(self):
...
         if (
             not image.startswith(("http://", "https://"))
-            and image not in images.keys()
+            and image not in images
         ):
             sys.exit(
-                "ERROR: '{}' is not in the list of known "
-                "images for that queue, please select "
-                "another.".format(image)
+                f"ERROR: '{image}' is not in the list of known "
```
boukeas, line 900: Also would prefer logging here. Generally, if a message starts with a ERROR or WARNING, it's a good hint :grin:

cli/testflinger_cli/helpers.py
```diff
@@ -51,3 +51,82 @@ def parse_filename(
     ):
         raise SnapPrivateFileError(path)
     return path
+
+
+def prompt_for_image(images: dict[str, str]) -> str:
```
boukeas, line 56: I like how you renamed these functions but my (possibly controversial) opinion is that not only should they not be moved to a different module, they should be turned into (static or class) methods of `TestflingerCLI` because that's where their functionality belongs conceptually.

cli/testflinger_cli/helpers.py
```diff
@@ -51,3 +51,82 @@ def parse_filename(
...
+    """Prompt the user to select an image from a list.
+
+    :param images: A mapping of image name to image job line to choose from.
+    :return: The selected image.
+    """
+    input_msg = "\nEnter the image you wish to use"
+    if images and images[list(images)[0]].startswith("url:"):
```
boukeas, line 63: I am not 100% clear what this is trying to do but I am pretty sure it can be improved.

cli/testflinger_cli/consts.py
```diff
@@ -5,3 +5,17 @@
 SNAP_PRIVATE_DIRS = [
     "/tmp",  # noqa: S108
 ]
+
+LOG_FORMAT = (
+    "%(levelname)s: %(asctime)s %(filename)s:%(lineno)d -- %(message)s"
+)
+LOG_DATE_FORMAT = "%Y-%m-%d %H:%M:%S"
+
+TESTFLINGER_SERVER = "https://testflinger.canonical.com"
```
boukeas, line 14: I am wondering if this perhaps belongs to the common module.

cli/testflinger_cli/consts.py
```diff
@@ -5,3 +5,17 @@
...
+TESTFLINGER_ERROR_THRESHOLD = 3
+
+ADVERTISED_QUEUES_MESSAGE = (
+    "ATTENTION: This only shows a curated list of queues with "
+    "descriptions, not ALL queues. If you can't find the queue you want "
+    "to use, a job can still be submitted for queues not listed here.\n"
+)
```
boukeas, line 21: To be honest, I really don't see why this needs to be moved here. It's just a constant used in a single module. If anything, you'll be forcing whoever needs to find it or change it to search for it in a different module than the one that uses it.

cli/testflinger_cli/__init__.py
pedro-avalos: I agree; I wanted to make minimal changes in this PR, but I can change that to `logging`

cli/testflinger_cli/__init__.py
```diff
@@ -50,84 +58,16 @@ def cli():
     """Generate the TestflingerCli instance and run it."""
     try:
         tfcli = TestflingerCli()
-        configure_logging()
```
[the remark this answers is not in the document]
pedro-avalos: Hm, I personally think that if the entire purpose of the function is to just wrap a single function call then the `configure_logging` function should not exist

cli/testflinger_cli/helpers.py
pedro-avalos: I've simplified it and make it work in more cases. You're right, that was a confusing line

cli/testflinger_cli/consts.py
pedro-avalos: I've removed this const by creating a single `do_list_queues` function that will log this warning properly as a logging warning and return the queues.

boukeas (approved):

PR #690, opened 2025-05-27
oem_autoinstall: to verify oem iso before provisioning

Artur-at-work:
## Description
After this change, oem_autoinstall will check the content of ISO to verify oem images. If iso doesn't match criteria, it will exit.
Another minor change is not using the TLS certificate when downloading ISO from cache server, because it is not required anymore.

## Resolved issues
Closes https://warthogs.atlassian.net/browse/OEX86-701

## Tests
Run with local agent to provision two Somerville machines in hotlab and cert lab.
1) DUT boot from alloem bootstrap image
2) DUT already provisioned, and run re-provisioning

Provision OEM ISO -> pass
Provision Ubuntu Desktop CDI ISO -> fail
Provision non-iso file -> fail

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/data/oem_autoinstall/provision-image.sh
+++ b/device-connectors/src/testflinger_device_connectors/data/oem_autoinstall/provision-image.sh
@@ -158,6 +158,58 @@ wget_iso_on_dut() {
     fi
 }
 
+is_valid_iso_on_dut() {
+    local addr=$1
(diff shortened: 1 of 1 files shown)


Artur-at-work:
hi @kevinyehk could you review this please? Thanks.

kevinyehk (approved):
LGTM +1. With this change, we can prevent provisioning the oem_autoinstall machine with the wrong image.

PR #693, opened 2025-05-29
feat(devices): Use uv-build for device-connectors

pedro-avalos:
## Description

Since this package has a version identifier, we can use `uv-build` for it


rene-oromtz (approved):
LGTM!

PR #695, opened 2025-05-30
infra: Update workflows for more testing

pedro-avalos:
## Description

This updates the workflows to do the following:

- Run tests on any PR anywhere
- CLI snap is built and uploaded as artifact on any PR/workflow-dispatch (useful for testing)
  - The snap is not published unless the workflow is run on the `main` branch
- Ensure workflows are run when the workflow file is changed
- Add `name` fields to make the GH logs easier to read

### DIFF
--- a/.github/workflows/agent-host-charm-check-libs.yml
+++ b/.github/workflows/agent-host-charm-check-libs.yml
@@ -3,8 +3,7 @@ permissions:
   contents: read
 on:
   pull_request:
-    branches:
-      - main
+    branches: [main]
     paths:
       - .github/workflows/agent-host-charm-check-libs.yml
       - agent/charms/testflinger-agent-host-charm/**
--- a/.github/workflows/agent-tox.yml
+++ b/.github/workflows/agent-tox.yml
@@ -10,7 +10,6 @@ on:
       - agent/**
       - .github/workflows/agent-tox.yml
   pull_request:
-    branches: ["**"]
     paths:
       - common/** # testflinger-agent depends on testflinger-common
       - agent/**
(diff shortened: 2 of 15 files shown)


.github/workflows/agent-tox.yml
```diff
@@ -10,7 +10,7 @@ on:
       - agent/**
       - .github/workflows/agent-tox.yml
   pull_request:
-    branches: [main]
+    branches: ["**"]
```
boukeas, line 13: Is this different than not specifying branches at all?

.github/workflows/agent-host-charm-check-libs.yml
```diff
@@ -3,8 +3,7 @@ permissions:
   contents: read
 on:
   pull_request:
-    branches:
-      - main
+    branches: [main]
```
boukeas, line 6: I don't mind either way but is this change simply a matter of stylistic preference? Some people prefer the "itemized list" version that YAML files offer because it's easier to review changes when items are added or removed.

boukeas (commented):
I am perfectly fine with the essence of the PR, I just have a couple of questions.

.github/workflows/agent-host-charm-check-libs.yml
pedro-avalos: Yes, I noticed we had a mix of how we define the `branches` field. If you prefer the itemized list, we can go with that, but I don't think we will be changing these very often.

.github/workflows/agent-tox.yml
pedro-avalos: Hm, perhaps not, I can remove it

boukeas (approved):

PR #700, opened 2025-06-03
Add Metrics Endpoint to Testflinger Agents

val500:


agent/pyproject.toml
```diff
@@ -11,6 +11,7 @@ authors = []
 requires-python = ">=3.10"
 dependencies = [
     "influxdb>=5.3.2",
+    "prometheus-client>=0.22.1",
```
pedro-avalos, line 14: Could you update the agent lockfile? I think you can run `uv lock` to update it.

agent/src/testflinger_agent/metrics.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+        self.total_failures = Counter(
+            "failures",
+            "Total job failures since last agent restart",
+            ["test_phase"],
+        )
+        try:
+            start_http_server(port)
+        except Exception:
+            logger.error("Unable to start metrics endpoint")
```
pedro-avalos, line 58: Should this be a silent error? I feel like if we are adding metrics, we should ensure that the metrics is started.

Also, let's not do `except Exception`.

agent/src/testflinger_agent/__init__.py
```diff
@@ -201,4 +202,11 @@ def parse_args():
         default="testflinger-agent.conf",
         help="Testflinger agent config file",
     )
+    parser.add_argument(
+        "--metrics_port",
+        "-p",
+        default="8000",
+        type=int,
+        help="Port to expose metrics endpoint on",
+    )
```
pedro-avalos, line 211: I love the push for metrics, but should this be a hard requirement by default in the agent code?

I wonder if we should make this default to `None` and only start a metrics handler if a port is provided. We can make it the default in the charm, though, by editing the Supervisor template.

agent/src/testflinger_agent/agent.py
```diff
@@ -99,6 +100,9 @@ def __init__(self, client):
         signal.signal(signal.SIGUSR1, self.restart_signal_handler)
         self.set_agent_state("waiting")
         self._post_initial_agent_data()
+        self.metrics_handler = PrometheusHandler(
+            self.client.config.get("metrics_endpoint_port")
+        )
```
pedro-avalos, line 105: Depending on whether we want to enforce metrics at this level, this may need to have an if or we may want a mock metrics endpoint, whatever makes most sense to you

pedro-avalos (changes requested):
I have some questions regarding whether we want to enforce metrics at the agent level or the charm level.

I definitely think that if we are trying to enforce metrics, a failure to initialize the metrics endpoint should be a noisier (even exit) error.

agent/src/testflinger_agent/metrics.py
val500: start_http_server doesn't really return any sensical exceptions - with the toggle, I think it makes sense for this to not be silent anymore.

pedro-avalos (approved):
This looks good. Thanks for addressing the comments!

PR #709, opened 2025-06-12
fix(server): Add noprovision provision_data schema

pedro-avalos:
## Description

This PR allows `provision_data` to be null; this is used for the `noprovision` provision method.

## Web service API changes

Allow `provision_data` to be empty/null.

## Tests

Added unit test for a `noprovision` job with a null `provision_data`.


p-gentili:
Can you please update the reference page https://canonical-testflinger.readthedocs-hosted.com/en/latest/reference/device-connector-types.html#noprovision ?

agent/src/testflinger_agent/job.py
```diff
@@ -63,23 +63,23 @@ def run_test_phase(self, phase, rundir):
...
-            return 0, None, None
+            if not self.job_data.get(f"{phase}_data"):
+                logger.info(
+                    "No %s_data defined in job data, skipping...", phase
+                )
+                return 0, None, None
+            if self.job_data.get(f"{phase}_data", {}).get("skip", False):
+                logger.info(
+                    "Skipping %s phase as requested in job data", phase
+                )
```
p-gentili, line 81: this makes `skip` applicable to any phase though. Why don't you instead handle `skip` inside the noprovision connector's `provision` function?

pedro-avalos: Only for phases which we know we could potentially skip.

Most of these phases have defined schemas. I think this is the fastest way to hotfix/patch the feature.

The device connector doesn't have a `provision` function, and it might take some more investigation to get that implementation to work.

pedro-avalos: Done

p-gentili (approved):
TYVM!

PR #710, opened 2025-06-13
oem_autoinstall: add ubuntu to group adm and sudo

xanthein:
## Description

Add ubuntu to group adm and sudo. Group adm is used for system monitoring tasks. Add user to group adm to make user able to read the journal logs. Member in group sudo  can execute any command with sudo or pkexec.


KaiChuan-Hsieh (approved):
LGTM

PR #712, opened 2025-06-13
feat(server): Modify server and agents table to allow comments on agent state

rene-oromtz:
## Description
This is a prerequisite for the admin status handling for the agent. Whenever an agent is set to be offline or in maintenance, the reason should be indicated and displayed in the UI for visualization. 

This PR does two things:
1. Allow the `comments` key in the `AgentIn/Out` Schema
2. Modify UI to display an informational tooltip if there are any messages available for the agent. 

## Resolved issues
Resolves [CERTTF-650 ](https://warthogs.atlassian.net/browse/CERTTF-650)

## Tests

Tested in local deployment. 
Screenshot from 2025-06-13 12-49-12

### DIFF
--- a/server/src/testflinger/api/schemas.py
+++ b/server/src/testflinger/api/schemas.py
@@ -53,6 +53,7 @@ class AgentIn(Schema):
     provision_type = fields.String(required=False)
     queues = fields.List(fields.String(), required=False)
     state = fields.String(required=False)
+    comments = fields.String(required=False)
 
 
 class AgentOut(Schema):
@@ -64,6 +65,7 @@ class AgentOut(Schema):
     location = fields.String(required=False)
     provision_type = fields.String(required=False)
     job_id = fields.String(required=False)
+    comments = fields.String(required=False)
 
 
 class ActionIn(Schema):
--- a/server/src/testflinger/templates/_agents_table.html
+++ b/server/src/testflinger/templates/_agents_table.html
@@ -51,15 +51,14 @@
           <td data-heading="Location">{{ agent.location or "unknown" }}</td>
           <td class="has-overflow" data-heading="State">
             {% if agent.get('comments') %}
-              <span> {{ agent.state }} </span>
+              <span>{{ agent.state }}</span>
               <span class="p-tooltip--top-left" aria-describedby="tooltip">
                 <i class="p-icon--information"></i>
-                <span class="p-tooltip__message" 
-                      role="tooltip" 
+                <span class="p-tooltip__message"
+                      role="tooltip"
                       id="tooltip"
-                      style="white-space: nowrap; text-align: left">
-                      {{ agent.comments }}
-                </span>
+                      style="white-space: nowrap;
+                             text-align: left">{{ agent.comments }}</span>
               </span>
             {% else %}
               {{ agent.state }}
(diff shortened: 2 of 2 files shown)


server/src/testflinger/api/schemas.py
```diff
@@ -53,6 +53,7 @@ class AgentIn(Schema):
     provision_type = fields.String(required=False)
     queues = fields.List(fields.String(), required=False)
     state = fields.String(required=False)
+    comments = fields.String(required=False)
```
pedro-avalos, line 56: Nitpick, should be `comment`.

`comments` might imply a list, rather than just a single string

server/src/testflinger/api/schemas.py
```diff
@@ -64,6 +65,7 @@ class AgentOut(Schema):
     location = fields.String(required=False)
     provision_type = fields.String(required=False)
     job_id = fields.String(required=False)
+    comments = fields.String(required=False)
```
pedro-avalos, line 68: Same as other comment

server/src/testflinger/templates/_agents_table.html
```diff
@@ -49,7 +49,21 @@
...
+                <i class="p-icon--information"></i>
+                <span class="p-tooltip__message"
+                      role="tooltip"
+                      id="tooltip"
+                      style="white-space: nowrap;
+                             text-align: left">{{ agent.comments }}</span>
+              </span>
+            {% else %}
+              {{ agent.state }}
+            {% endif %}
```
pedro-avalos, line 65: Hm, I wonder, would this work?

```html
<span>{{ agent.state }}</span>
{% if agent.get('comment') %}
...
{% endif %}
```

I see the `agent.state` bit is used in both cases, maybe we can simplify it?

server/src/testflinger/templates/_agents_table.html
```diff
@@ -49,7 +49,21 @@
...
+            {% if agent.get('comments') %}
+              <span>{{ agent.state }}</span>
+              <span class="p-tooltip--top-left" aria-describedby="tooltip">
+                <i class="p-icon--information"></i>
+                <span class="p-tooltip__message"
+                      role="tooltip"
+                      id="tooltip"
+                      style="white-space: nowrap;
+                             text-align: left">{{ agent.comments }}</span>
+              </span>
```
pedro-avalos, line 62: I think we have other tooltips (or may want to add more), could you test if we can use an `id` that is a little more unique?

```suggestion
              <span class="p-tooltip--top-left" aria-describedby="state-tooltip">
                <i class="p-icon--information"></i>
                <span class="p-tooltip__message"
                      role="tooltip"
                      id="state-tooltip"
                      style="white-space: nowrap;
                             text-align: left">{{ agent.comments }}</span>
              </span>
```

pedro-avalos (changes requested):
This is a great addition.

I've added some nitpicks.

rene-oromtz:
Thanks for the review @pedro-avalos 

Tested again the local changes and is working as expected. 
Screenshot from 2025-06-13 14-45-33

pedro-avalos (dismissed):
Great, thanks!

rene-oromtz:
@pedro-avalos 
Minor typo found before merging, can you please help me with re-reviewing?

pedro-avalos (approved):
Good catch

PR #717, opened 2025-06-18
oem_autoinstall: Add zapper related keys in provisioning data

Artur-at-work:
## Description
1. Add zapper_iso_type and zapper_iso_url in provisioning data keys. When zapper_iso_type is present, device connector will call zapper to flash zapper_iso_url via typecmux.

2. Update provision-image.sh script:
 - after provisioning completed, set the BootOrder via efibootmgr to USB. When zapper USB is empty, then next device will be selected to boot. When zapper USB flashed with ISO, then this ISO will boot (i.e bootstrap image)
 - cleaned up the logging to reduce noise/warnings

## Resolved issues
Closes https://warthogs.atlassian.net/browse/OEX86-726

## Documentation
WIP

## Web service API changes
Add zapper_iso_type and zapper_iso_url in schema


kevinyehk (commented):
This PR looks mostly reasonable to me but it requires input from @p-gentili, I'll leave it to him.

Artur-at-work:
> Let's define first the server side, than we can adjust here what we need.

Server side logic was merged. Could you review this please? Thanks.

p-gentili:
Can you confirm this is not a new connector but rather just an extension of oem_autoinstall?

PR #724, opened 2025-06-24
 feat: update agent's behavior to only accept restricted queues if any

diohe0311:
## Description

This PR updates client.check_jobs() behavior according to the discussion in the [spec](https://docs.google.com/document/d/1EUZxKwsuQf-VkcYWFEb4lvOvYqE_wmyhV8RAyR6lrD8):

- Add `get_agent_data()` to query `/v1/agents/data/<agent_id>` endpoint.
- If any of the agent’s queues are marked as restricted, the agent will only listen on those restricted queues.
- Added unit tests:
    - `test_check_jobs_with_restricted_queues`
    - `test_check_jobs_with_unrestricted_queues`
    - `test_get_agent_data`

## Resolved issues

Resolves [LM-1934](https://warthogs.atlassian.net/browse/LM-1934)

## Tests
Added test cases and passed the tests:
```
(testflinger-agent) dio@dio-zenbook:~/Documents/testflinger/agent$ pytest tests/test_client.py
================================================================= test session starts ==================================================================
platform linux -- Python 3.10.12, pytest-8.3.5, pluggy-1.5.0
rootdir: /home/dio/Documents/testflinger/agent
configfile: pyproject.toml
plugins: cov-6.1.1, timeout-2.3.1, mock-3.14.0, requests-mock-1.12.1, xdist-3.6.1
collected 17 items                                                                                                                                     

tests/test_client.py .................                                                                                                           [100%]

================================================================== 17 passed in 0.08s ==================================================================
```


agent/src/testflinger_agent/client.py
```diff
@@ -220,6 +235,26 @@ def get_result(self, job_id):
...
+            logger.error(
+                "Unable to get agent data from: %s (error: %d)",
+                url,
+                response.status_code,
+            )
+            return {}
+        if response.content:
+            return response.json()
+        else:
+            return {}
```
rene-oromtz, line 256: I think this can be simplified if we expect to return {} upon any failure.
```suggestion
        try:
            response = self.session.get(agent_data_url, timeout=30)
            response.raise_for_status()
            return response.json()
        except (requests.exceptions.RequestException, ValueError) as exc:
            logger.error("Failed to retrieve agent data: %s", exc)
            return {}
```

This suggestion comes from one of my PR where I actually want to do the same thing, you can refer to this [link](https://github.com/canonical/testflinger/blob/f715257f1d535b143aa71f043de6932e4c0aeacb/agent/src/testflinger_agent/client.py) to the check the changes I propose to this specific file. Let me know your thoughts and we can maybe agree on something that works best for both of us. 

Could you please also add typehints?

agent/src/testflinger_agent/client.py
```diff
@@ -91,11 +91,26 @@ def _configure_influx(self):
...
+            ]
+
+            if restricted_queues:
+                queue_list = restricted_queues
+            else:
+                queue_list = all_queues
+
             job_uri = urljoin(self.server, "/v1/job")
-            queue_list = self.config.get("job_queues")
             logger.debug("Requesting a job")
```
rene-oromtz, line 114: This comes from the "old" code but I think we can move all this logic before the `try` as those lines won't be raising an exception of type `requests.exceptions.RequestException` which is what we are catching here. The only piece of code that can raise this exception is `self.session.get()`

I also don't think any of this will be raising an exception that's worth adding as you are handling it properly with the `.get()`

agent/src/testflinger_agent/client.py
```diff
@@ -91,11 +91,26 @@ def _configure_influx(self):
...
+        If the agent has restricted queues, only accept jobs from those queues.
+
         :return: Dict with job data, or None if no job found
         """
         try:
+            agent_id = self.config.get("agent_id")
+            agent_data = self.get_agent_data(agent_id)
+
+            all_queues = self.config.get("job_queues", [])
+            restricted_to = agent_data.get("restricted_to", {})
```
rene-oromtz, line 103: There is a slight problem with this approach as we are getting the "restricted_to" key from the /agents/data/<agents_name>. 

I forgot that endpoint `/agents/data/<agents_name>` and `/agents/data/` are not directly related. We already merged the modifications done to the second endpoint (which is useful for the UI) but if we want to retrieve the same data from API for a single agent, we will need to create another PR that does the same process with an specific agent only.

rene-oromtz (changes requested):
Thanks for taking a look at this. I added some suggestions based on another PR I'm working on that is doing similar modifications to the agent code. 

Also, we might need to create a spinoff PR to allow the `/agents/data/<agent_name>` to retrieve the restricted_queues. As mentioned in one of the comments, your initial PR did the modification to the `/agents/data` endpoint but those two retrieve the agent(s) data different so we need to add your changes also this endpoint. I'm sorry I forgot about this detail. 

Once we cover all this, I would like also to test those changes in staging before merging just to be safe.

agent/src/testflinger_agent/client.py
diohe0311: Thanks for the feedback! I just double-checked the code from main branch, and it looks like the /agents/data/<agent_name> GET endpoint already includes the logic for restricted queues, same as what I changed in the previous PR.
https://github.com/canonical/testflinger/blob/486c38541acdf81d170398c6c0668cc9bf5f856a/server/src/testflinger/api/v1.py#L687

Or I'm missing something here?

rene-oromtz: Ah, that was also on me. I think I tried a different revision of the branch where you made this modification (wasn't testing it on main in my local). 

But yes, just doubled checked the code. All good, sorry about that

rene-oromtz (commented):
Thanks for adding the suggestions! This looks good to me. 

I also added an additional [comment](https://docs.google.com/document/d/1EUZxKwsuQf-VkcYWFEb4lvOvYqE_wmyhV8RAyR6lrD8/edit?disco=AAABmuI-C9s) to the spec is just a consideration on which queues we should add to the restricted list. 

As mentioned there, I think we should be only adding queues by agent_name or canonical id and not customized queues. 

Also, will test this on staging on Monday before approving/merging just to make sure the behavior is as expected.

rene-oromtz (approved):

diohe0311:
> I'm a bit confused as to why the mapping between queue and owner is stored within the agent data. Agents and queues aren't strictly coupled in Testflinger - wouldn't it make more sense to store this info in the restricted queues collection? Then you could just check on the server side whether any of the requested queues are restricted and to ignore the other ones. I could be missing something though.

Thanks for raising this, Varun.

The mapping between queues and client IDs (owners) is not stored in the Testflinger DB (i.e., it’s not part of any collection). It's simply included as part of the `/agents/data` API response, which is consumed by C3 to render the web UI. This behavior follows what was defined in the spec I shared earlier.

The decision to surface this data through the agent data API was based on keeping changes minimal across the Testflinger, C3, and Lab Manager stacks. Since C3 already fetches agent data via `/agents/data`, it felt natural to extend this response rather than introduce a new endpoint or modify server-side storage.

Regarding your suggestion to move this logic to the server and check queue restrictions there — If I understand correctly, the decision to accept or ignore job queues happens on the agent side, not the server. So the server itself wouldn’t have the context to filter or ignore queues during job submission.

That said, if there’s a longer-term direction where queue ownership metadata should live in a dedicated collection (currently this data is spread across `client_permissions`, `restricted_queues`, and `agents`), I’m happy to revisit and discuss further. For now, this approach aligns with the existing architecture and keeps the implementation relatively simple.

val500:
Thanks for the explanation, I'm good with merging it once the test passes.

PR #729, opened 2025-06-27
fix: add attachments to oem autoinstall schema

pedro-avalos:
## Description

Adds `attachments` to the `oem_autoinstall` provision data schema.

## Tests

Added unit test to ensure this field is tested on the schema


boukeas (approved):
Looks good. Thanks for fixing this instantly! :star:

PR #732, opened 2025-06-30
Fix: encoded user data should be a str

p-gentili:
## Description

After encoding bytes to b64, the expectation was to move around a str, not bytes again (which is the return type of `b64encode` for some reason).

## Resolved issues

N/A

## Documentation

N/A

## Web service API changes

N/A

## Tests

Tested locally with
```
# job.json
{
    "provision_data": {
      "url": "<tel-cache-iso>",
      "robot_tasks": [
        "hp/Z2/G9/secure_boot/disable.robot",
        "hp/Z2/G9/boot/boot_from_usb.robot",
        "common/grub/autoinstall/http/configure_autoinstall_http.robot",
        "hp/boot/wait_reboot.robot"
        ],
      "autoinstall_oem": true
    }
}

```


pedro-avalos (approved):
Ah, didn't realize that was bytes either. This looks good

PR #733, opened 2025-06-30
Change: pass part of TF configuration to provision API

p-gentili:
## Description

The agent name is quite useful on the service side, independently  from the selected method. This API moves part of the arguments definition to the base class, also adding "agent-name" and "cid" to the mix.

## Resolved issues

Part of [ZAP-1235](https://warthogs.atlassian.net/browse/ZAP-1235)

## Documentation

N/A

## Web service API changes

N/A

## Tests

There's nothing on the service side making use of these values yet. Only tested with unit tests.


pedro-avalos (changes requested):
I think this should be an okay way to access the CID for the connector (our internal tooling adds that field automatically). I just left one minor comment. @boukeas or @val500 what do you think?

PR #739, opened 2025-07-02
(CI): Set limits for coverage reports

rene-oromtz:
## Description
This PR sets limits for codecov. By default, all components are set to 70% for project and patch checks. 
An exemption was made to device-connector as this is the lowest rated component. 
Since the project seems to be calculated based on each component, the passing grade for device-connector is 50% (its current rating) and patch is set to 60%. 

Reference: 
https://docs.codecov.com/docs/components#getting-started-with-components

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -11,6 +11,9 @@ component_management:
     - component_id: agent
       name: Agent
       paths: [agent/**]
+      statuses:
+        - type: project
+          target: 69%
     - component_id: cli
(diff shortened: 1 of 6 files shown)


codecov.yaml
pedro-avalos, line 1: Let's remember to bump these up to the 70% as we see the coverage go up. But this is better than nothing for a starting point.

pedro-avalos (approved):

PR #744, opened 2025-07-08
feat(cli): Add Admin module

rene-oromtz:
## Description
This PR add a new Admin Module where all privileged command will be executed upon authentication. 
This module is built on top of current "main" CLI but it has been separated as a different module to release burden on `__init__.py` file and to allow scalability of admin commands. 
Authorisation is made with a decorator approach that can be reused by future commands.

## Resolved issues
Resolves [CERTTF-646](https://warthogs.atlassian.net/browse/CERTTF-646)

## Tests
Added additional Unit Test that validates admin cli functionality, in this case, they cover the only command available "set agent-status"

### DIFF
--- a/cli/testflinger_cli/admin.py
+++ b/cli/testflinger_cli/admin.py
@@ -97,7 +97,7 @@ class TestflingerAdminCLI:
             and not self.main_cli.args.comment
         ):
             sys.exit(
-                "Commment is required when setting agent status to offline."
+                "Comment is required when setting agent status to offline."
             )
 
         for agent in agents:
--- a/cli/testflinger_cli/admin.py
+++ b/cli/testflinger_cli/admin.py
@@ -1,0 +1,128 @@
+# Copyright (C) 2020-2022 Canonical
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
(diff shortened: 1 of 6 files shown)


pedro-avalos (changes requested):

cli/testflinger_cli/admin.py
```diff
@@ -0,0 +1,134 @@
...
+        client_id = self.main_cli.auth.client_id
+
+        # Creating dictonary to define formmated comments
+        comment_dict = {
+            "waiting": lambda _, __: "",
+            "offline": lambda user,
+            comment: f"Set to offline by {user}. Reason: {comment}",
+            "maintenance": lambda user,
+            _: f"Set to offline by {user} for lab related task.",
+        }
```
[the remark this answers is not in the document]
rene-oromtz: Ah, this is way better! Thanks for the suggestion!

cli/testflinger_cli/admin.py
```diff
@@ -0,0 +1,137 @@
...
+            try:
+                self.main_cli.client.set_agent_status(agent, status, comment)
+                if agent_status in test_status:
+                    print(
+                        f"Agent {agent} processing job. "
+                        f"Status {status} deferred until job completion"
+                    )
+                else:
+                    print(f"Agent {agent} status is now: {status}")
+            except client.HTTPError as exc:
```
pedro-avalos, line 135: Should we be re-raising the exception if it's not `NOT_FOUND`?

I'm okay with not re-raising it since we're looping through and the rest might be okay. Mind at least adding a message to avoid any confusion about whether an agent failed:

```python
print(f"Exception raised when setting status for {agent}: {exc}")
```

pedro-avalos (changes requested):
This looks good. I just left one last comment regarding the HTTP error handling.

cli/testflinger_cli/admin.py
rene-oromtz: Yeah, I think might be best to at least log the failure (if any besides 404). Let me add that right away, thanks for pointing this out.

PR #746, opened 2025-07-09
feat(submit): support authentication in submit action

k-s-0:
## Description

This adds support for submitting jobs with authentication (`--client_id`, `--secret_key`) via the `submit` action by adding `client-id` and `secret-key` optional inputs and passing them to the invocation of `testflinger submit` if they are both provided.

## Resolved issues

N/A

## Documentation

The API table in the README.md associated with the `submit` action has been updated to include the `client-id` and `secret-key` optional inputs.

## Web service API changes

N/A

## Tests

I created a simple test repo with a workflow that invokes the submit action from my fork, providing a client_id and secret_key (stored as repository secrets) associated with the queue used. test_job.yaml declares `job_priority: 10` and `timeout: 259200`, both of which require elevated privileges via authentication, else the job is rejected. See completed workflow runs in the test repo [here](https://github.com/canonical/tf-submit-auth-test/actions).

### DIFF
--- a/.github/actions/submit/action.yaml
+++ b/.github/actions/submit/action.yaml
@@ -156,14 +156,20 @@ runs:
         JOB: ${{ steps.create-job-file.outputs.job }}
       run: |
         echo "::group::Submit job to the Testflinger server"
-        if [ -n "$RELATIVE_TO" ]; then
-          RELATIVE="--attachments-relative-to $RELATIVE_TO"
-        fi
-        if [[ -n "$CLIENT_ID" && -n "$SECRET_KEY" ]]; then
-          JOB_ID=$(testflinger --server $SERVER submit --quiet $RELATIVE --client_id "$CLIENT_ID" --secret_key "$SECRET_KEY" "$JOB")
-        else
-          JOB_ID=$(testflinger --server $SERVER submit --quiet $RELATIVE "$JOB")
-        fi
+        AUTH_ARGS_COUNT=0
+        [[ -n $CLIENT_ID ]] && ((++AUTH_ARGS_COUNT))
+        [[ -n $SECRET_KEY ]] && ((++AUTH_ARGS_COUNT))
+        case $AUTH_ARGS_COUNT in
+            0) AUTH_ARGS=() ;;
+            2) AUTH_ARGS=( --client_id "$CLIENT_ID" --secret_key "$SECRET_KEY" ) ;;
+            1) echo "::error::Both 'client-id' and 'secret-key' must be provided or neither"; exit 1 ;;
+        esac
+        JOB_ID=$(\
+          testflinger --server "$SERVER" submit --quiet \
+          ${RELATIVE_TO:+--attachments-relative-to "$RELATIVE_TO"} \
+          "${AUTH_ARGS[@]}" \
+          "$JOB" \
+        )
         echo "id=$JOB_ID" >> $GITHUB_OUTPUT
         echo "::endgroup::"
         echo "::notice::Submitted job $JOB_ID"


.github/actions/submit/action.yaml
```diff
@@ -145,13 +151,19 @@ runs:
...
         echo "::group::Submit job to the Testflinger server"
         if [ -n "$RELATIVE_TO" ]; then
           RELATIVE="--attachments-relative-to $RELATIVE_TO"
         fi
-        JOB_ID=$(testflinger --server $SERVER submit --quiet $RELATIVE "$JOB")
+        if [[ -n "$CLIENT_ID" && -n "$SECRET_KEY" ]]; then
+          JOB_ID=$(testflinger --server $SERVER submit --quiet $RELATIVE --client_id "$CLIENT_ID" --secret_key "$SECRET_KEY" "$JOB")
+        else
+          JOB_ID=$(testflinger --server $SERVER submit --quiet $RELATIVE "$JOB")
+        fi
```
boukeas, line 166: I'm thinking maybe when only one of the authorization arguments are provided we shouldn't be ignoring that silently. I thought about this a little and this is the best I could come up with:
```
AUTH_ARGS_COUNT=0
[[ -n $CLIENT_ID ]] && ((AUTH_ARGS_COUNT++))
[[ -n $SECRET_KEY ]] && ((AUTH_ARGS_COUNT++))

case $AUTH_ARGS_COUNT in
    0) AUTH_ARGS= ;;
    2) AUTH_ARGS="--client_id $CLIENT_ID --secret_key $SECRET_KEY" ;;
    1) echo "::error::Both 'client-id' and 'secret-key' must be provided or neither"; exit 1 ;;
esac

JOB_ID=$(\
  testflinger --server $SERVER submit --quiet \
  "${RELATIVE_TO:+--attachments-relative-to $RELATIVE_TO}" \
  "$AUTH_ARGS" \
  "$JOB"
)
```
Note this also involves _removing_ this block:
```
if [ -n "$RELATIVE_TO" ]; then
  RELATIVE="--attachments-relative-to $RELATIVE_TO"
fi
```
If you agree, would you mind testing this in your workflow?

boukeas (changes requested):
Thank you so much for addressing this issue, much appreciated. I think that the call to testflinger is becoming a little convoluted now that additional command-line arguments are added, so I've made a suggestion that (I hope) simplifies it a little bit. Let me know what you think and please do test again if you're in agreement.

.github/actions/submit/action.yaml
```diff
@@ -145,13 +151,25 @@ runs:
...
-        if [ -n "$RELATIVE_TO" ]; then
-          RELATIVE="--attachments-relative-to $RELATIVE_TO"
-        fi
-        JOB_ID=$(testflinger --server $SERVER submit --quiet $RELATIVE "$JOB")
+        AUTH_ARGS_COUNT=0
+        [[ -n $CLIENT_ID ]] && ((++AUTH_ARGS_COUNT))
+        [[ -n $SECRET_KEY ]] && ((++AUTH_ARGS_COUNT))
+        case $AUTH_ARGS_COUNT in
+            0) AUTH_ARGS=() ;;
+            2) AUTH_ARGS=( --client_id "$CLIENT_ID" --secret_key "$SECRET_KEY" ) ;;
```
boukeas, line 164: I might be missing something here but what's the advantage of using an array here instead of a simple string? We're not really using `$AUTH_ARGS` as an array anywhere.

k-s-0: Ah sorry, I meant to add a comment discussing my amendments to your suggestion. 

In the original:

`AUTH_ARGS="--client_id $CLIENT_ID --secret_key $SECRET_KEY"`

the expansion of `$SECRET_KEY` isn't quoted, so bash will try to interpret symbols like `*`, `&`, etc.. Using an array allows $SECRET_KEY to be directly quoted while associated with a variable ($AUTH_ARGS).

I also updated:

```
[[ -n $CLIENT_ID ]] && ((AUTH_ARGS_COUNT++))
[[ -n $SECRET_KEY ]] && ((AUTH_ARGS_COUNT++))
```
to use pre-fix incrementation, e.g., `++AUTH_ARGS_COUNT`, to avoid the expression evaluating to `TRUE && FALSE` when `AUTH_ARGS_COUNT=0` (returned before post-fix incrementation).

and

`"${RELATIVE_TO:+--attachments-relative-to $RELATIVE_TO}"`

to

`${RELATIVE_TO:+--attachments-relative-to "$RELATIVE_TO"}`

In the first case, the entire string becomes a single argument (the flag and it's value are quoted together). In the latter, they aren't quoted together, and inner quotes are added to `$RELATIVE_TO` to ensure dirnames with whitespace are not treated as split.

---

Thanks for the quick reviews and feedback! I completely agree with the suggestion to not silently ignore cases where only one of the auth args is passed. I've tested these changes and confirmed the auth args work as described. I'm finishing up testing with `--attachments-relative-to` just for sanity, since the new changes also affect how it is propagated to the `testflinger submit` command invocation.

boukeas: > the expansion of $SECRET_KEY isn't quoted, so bash will try to interpret symbols like *, &, etc.

I am not so sure about that, the entire value of `AUTH_ARGS` is quoted and `$SECRET_KEY` is within that value. Can you provide a minimal example where that would fail?

boukeas: > they aren't quoted together, and inner quotes are added to $RELATIVE_TO

:+1:

k-s-0: > > the expansion of $SECRET_KEY isn't quoted, so bash will try to interpret symbols like *, &, etc.
> 
> I am not so sure about that, the entire value of `AUTH_ARGS` is quoted and `$SECRET_KEY` is within that value. Can you provide a minimal example where that would fail?

[I created a branch](https://github.com/k-s-0/testflinger/blob/quoting_ex/.github/actions/submit/action.yaml) with the original / non-array version of `AUTH_ARGS`. Here's the relevant error output from a [workflow run](https://github.com/canonical/tf-submit-auth-test/actions/runs/16179648794/job/45673068665) in my test repo that uses this branch.

```
2025-07-09T20:36:13.3972136Z + AUTH_ARGS='--client_id *** --secret_key ***'
2025-07-09T20:36:13.3972900Z ++ testflinger --server https://testflinger.canonical.com submit --quiet --attachments-relative-to ./.github '--client_id *** --secret_key ***' .github/test_job.yaml
2025-07-09T20:36:14.4525983Z usage: testflinger-cli [-h] [-c CONFIGFILE] [-d] [--server SERVER]
2025-07-09T20:36:14.4527274Z                        {artifacts,cancel,config,jobs,list-queues,poll,poll-serial,reserve,status,agent-status,queue-status,results,show,submit}
2025-07-09T20:36:14.4528157Z                        ...
2025-07-09T20:36:14.4528555Z testflinger-cli: error: unrecognized arguments: .github/test_job.yaml
```

---

Oh... after trying to replicate with a more simplified script, I think the use of quotes when `"$AUTH_ARGS"` is used in the `testflinger` command is actually the issue:

```
#!/bin/bash

FOO='abc&*123^a[b]c'
ARGS="--foo $FOO"
ARGS_ARR=( --foo "$FOO" )

echo "scalar, unquoted:"
echo "ARGS: " $ARGS
./test.py $ARGS file.txt

echo
echo "scalar, quoted"
echo "ARGS: $ARGS"
./test.py "$ARGS" file.txt

echo
echo "array, individually quoted FOO"
echo "ARGS: ${ARGS_ARR[@]}"
./test.py "${ARGS_ARR[@]}" file.txt

```

test.py
```
#!/usr/bin/env python3
import argparse

parser = argparse.ArgumentParser()
parser.add_argument('--foo')
parser.add_argument('file')
args = parser.parse_args()
print("args:", args)

```

`$ ./test.sh`:
```
scalar, unquoted:
ARGS:  --foo abc&*123^a[b]c
args: Namespace(foo='abc&*123^a[b]c', file='file.txt')

scalar, quoted
ARGS: --foo abc&*123^a[b]c
usage: test.py [-h] [--foo FOO] file
test.py: error: unrecognized arguments: file.txt

array, individually quoted FOO
ARGS: --foo abc&*123^a[b]c
args: Namespace(foo='abc&*123^a[b]c', file='file.txt')

```

---

After creating [another branch](https://github.com/k-s-0/testflinger/blob/quoting_ex2/.github/actions/submit/action.yaml) with `$AUTH_ARGS` instead of `"$AUTH_ARGS"`, it appears to [work](https://github.com/canonical/tf-submit-auth-test/actions/runs/16179980300/job/45674136816)!

---

Though... there is an exceedingly unlikely case where the above *would* cause issues, where the unquoted expansion matches a local file via globbing (e.g., FOO='fi*le', file on disk named "fiXle" or similar, expansion incorrectly matches the filename via globbing and "fiXle" is passed instead of "fi*le").

test.sh:
```
#!/bin/bash

FOO='fi*le'
ARGS="--foo $FOO"
ARGS_ARR=( --foo "$FOO" )

echo "scalar, unquoted:"
echo "ARGS: " $ARGS
./test.py $ARGS file.txt

echo
echo "scalar, quoted"
echo "ARGS: $ARGS"
./test.py "$ARGS" file.txt

echo
echo "array, individually quoted FOO"
echo "ARGS: ${ARGS_ARR[@]}"
./test.py "${ARGS_ARR[@]}" file.txt
```

`$ touch fiXle; ./test.sh`:
```
scalar, unquoted:
ARGS:  --foo fiXle  # this is not what we wanted to pass!
args: Namespace(foo='fiXle', file='file.txt')

scalar, quoted
ARGS: --foo fi*le
usage: test.py [-h] [--foo FOO] file
test.py: error: unrecognized arguments: file.txt

array, individually quoted FOO
ARGS: --foo fi*le
args: Namespace(foo='fi*le', file='file.txt')

```

So, scalar with quoted usage fails because it prevents splitting on whitespace (treated as one grouped string), scalar with unquoted usage works but may fail if a local file happens to match what `$SECRET_KEY` may glob to (the secret keys may contain `*` - ours does; extremely unlikely but technically possible issue), and the array approach allows correct splitting on whitespace and prevents globbing on `*`.

Thoughts? Apologies for the long response :sweat_smile:.

boukeas: That's extremely thorough, I really appreciate the detail you went into.

PR #749, opened 2025-07-10
feat(server): Added logic for enforcing endpoint restriction

rene-oromtz:
## Description
Added two decorators to handle authentication/authorisation
1. authenticate: Will return the permissions given to a user (No permissions if user not authenticated)
2. require_role: Will determine if user is able to do a request on endpoint base on role. 

The logic for unauthenticated user is preserved. If attempting to submit a job it will authenticate the user. If not authenticated, need to follow default limitations. If authenticated, need to have the right privileges. 

Note: Need to also add "role" from the `client_credentials_admin` script. Will do in a separate PR. For now, if authenticated, will default to contributor. 

## Resolved issues
Resolves [CERTTF-665](https://warthogs.atlassian.net/browse/CERTTF-665)

## Tests
For authenticate decorator: All current Unit Test passed
For require_role: This decorator is unused but required by @diohe0311. I tested locally but in his PR #730 there should be unit tests that ensure only admin users are able to add/remove restricted queues. 

[CERTTF-665]: https://warthogs.atlassian.net/browse/CERTTF-665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/src/testflinger/api/auth.py
```diff
@@ -185,21 +207,75 @@ def check_token_reservation_timeout(
...
+            return func(*args, **kwargs)
+
+        # If there is a token available, attempt to retrieve information
+        secret_key = os.environ.get("JWT_SIGNING_KEY")
+        decoded_jwt = decode_jwt_token(auth_token, secret_key)
+        permissions = decoded_jwt.get("permissions", {})
+
+        # Store auth state if decoding was successful
+        g.client_id = permissions["client_id"]
+        g.role = permissions.get("role", "contributor")
```
pedro-avalos, line 251: Hm, just trying to familiarize myself more with the auth module. Someone can have a token, but not have a role? If so, this line makes sense I suppose.

Do we have an enum of possible roles that we could use here instead of a magic string?

server/src/testflinger/api/auth.py
```diff
@@ -185,21 +207,75 @@ def check_token_reservation_timeout(
...
+        g.permissions = permissions
+        g.is_authenticated = True
+        return func(*args, **kwargs)
+
+    return decorator
+
+
+def require_role(*roles):
```
pedro-avalos, line 259: codecov is complaining that this function has no coverage, do we not have any tests for API requests that require a role?

I think we haven't implemented any; if so, we can ignore these codecov warnings.

pedro-avalos: Should've read lower in the PR description :facepalm:

We can safely ignore the coverage warnings as this function is unused

server/pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "sentry-sdk>= 2.0.1",
     "urllib3>=2.2.1",
     "marshmallow-oneofschema>=3.1.1",
+    "strenum>=0.4.15",
```
pedro-avalos, line 28: Too bad the official [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum) is not available until Python 3.11

pedro-avalos (approved):
Thanks for adding the enum!

PR #752, opened 2025-07-14
refactor(agent): Modify agent status handling and remove tmp files

rene-oromtz:
## Description
The purpose of this PR is to improve the status handling for an agent. This is being done by removing the necessity of the creation/deletion of temporary files that can only be available by accessing the agent Charm Unit. 

The agent will now be able to get its state directly by querying the MongoDB by using server recent endpoint addition `GET /agents/data/<agent_name>` and process jobs depending on the state retrieved. 

Non breaking changes are introduced to perform the follows:
1. Whenever setting an agent state, the reason for the state is displayed in the UI. 
2. Agent is now able to detect if state is either `offline` or `maintenance` directly from the DB, in any case, the agent will stop looking for available jobs. 
3. Restart is also handled directly by temporarily setting a `restart` status on the agent. 
4. If at any point an offline/restart is detected, the new AgentStatusHandler will be able to carry over the reason for the offline/restart and only perform the action after the agent is not processing any job. 

## Resolved issues
Resolves [CERTTF-645](https://warthogs.atlassian.net/browse/CERTTF-645)

## Documentation
Documention needs to be updated to reflect how the agent is set to offline/restart/online, however, I'm making those modification once I completed the PR for the CLI changes. 

## Tests
This was tested and validated in Staging

Agent Offline:
Screenshot from 2025-06-17 14-44-28

Attempting to submit job
```
testflinger-cli --server=https://testflinger-staging.canonical.com submit -p ~/Desktop/staging.yaml
ERROR: No online agents available for queue 202409-35503. If you want to wait for agents to become available, use the --wait-for-available-agents option.
```

If at any point the restart is detected, a note will be shown with the current phase state:
Screenshot from 2025-06-17 15-02-14

Agent will continue processing the job and restart only when its safe to do so:
```
[25-06-17 20:48:50]    INFO: (agent.py:259)| Starting job 0268999f-01b0-4512-8736-b02a2312a8c0
[25-06-17 20:48:52]    INFO: (job.py:87)| Running setup_command: tf-setup
[25-06-17 20:49:07]    INFO: (job.py:87)| Running provision_command: tf-provision
[25-06-17 21:01:51]    INFO: (job.py:64)| No firmware_update_command configured, skipping...
[25-06-17 21:01:51]    INFO: (job.py:87)| Running test_command: tf-test
[25-06-17 21:04:34]    INFO: (job.py:74)| No allocate_data defined in job data, skipping...
[25-06-17 21:04:36]    INFO: (job.py:87)| Running reserve_command: tf-reserve
[25-06-17 21:05:33]    INFO: (job.py:87)| Running cleanup_command: tf-cleanup
[25-06-17 21:05:49]    INFO: (client.py:252)| Submitting job outcome for job: 0268999f-01b0-4512-8736-b02a2312a8c0
[25-06-17 21:05:50]    INFO: (agent.py:186)| Restarting agent
[25-06-17 21:05:53]   ERROR: (client.py:73)| InfluxDB host undefined
[25-06-17 21:05:53]    INFO: (__init__.py:138)| Checking jobs
```

[CERTTF-645]: https://warthogs.atlassian.net/browse/CERTTF-645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/src/testflinger_agent/agent.py
+++ b/agent/src/testflinger_agent/agent.py
@@ -26,7 +26,7 @@ from testflinger_common.enums import AgentState, JobState, TestEvent, TestPhase
 from testflinger_agent.config import ATTACHMENTS_DIR
 from testflinger_agent.errors import TFServerError
 from testflinger_agent.event_emitter import EventEmitter
-from testflinger_agent.handlers import RestartHandler
+from testflinger_agent.handlers import AgentStatusHandler
 from testflinger_agent.job import TestflingerJob
 from testflinger_agent.metrics import PrometheusHandler
 
@@ -99,7 +99,7 @@ class TestflingerAgent:
     def __init__(self, client):
         self.client = client
         self.agent_id = self.client.config.get("agent_id")
-        self.restart_handler = RestartHandler()
+        self.status_handler = AgentStatusHandler()
         signal.signal(signal.SIGUSR1, self.restart_signal_handler)
         self.set_agent_state(AgentState.WAITING)
         self._post_initial_agent_data()
@@ -155,20 +155,23 @@ class TestflingerAgent:
             agent_state = AgentState.OFFLINE
         return (agent_state, comment)
 
-    def check_offline(self) -> bool:
-        """Determine if the agent is offline.
+    def check_offline(self) -> tuple:
+        """Determine if agent should be taken offline.
 
-        :return: True if agent is offline, False otherwise.
+        :return: True or False along with the comment if any.
         """
-        agent_state, _ = self.get_agent_state()
+        agent_state, comment = self.get_agent_state()
 
+        # Offline set by server
         if (
             agent_state == AgentState.OFFLINE
             or agent_state == AgentState.MAINTENANCE
         ):
-            return True
-        else:
-            return False
+            return (True, comment)
+        # Offline deferred and handled by status handler
+        if self.status_handler.marked_for_offline():
+            return (True, self.status_handler.get_comment())
+        return (False, comment)
 
     def check_restart(self) -> tuple:
         """Determine if the agent requires a restart.
@@ -177,15 +180,15 @@ class TestflingerAgent:
         """
         agent_state, comment = self.get_agent_state()
 
-        # Restart requested by server
+        # Restart set by server
         if agent_state == AgentState.RESTART:
             return (True, comment)
-        # Restart requested by Signal
+        # Restart deferred and requested by Signal
         if (
-            self.restart_handler.marked_for_restart()
+            self.status_handler.marked_for_restart()
             and agent_state != AgentState.OFFLINE
         ):
-            return (True, self.restart_handler.get_comment())
+            return (True, self.status_handler.get_comment())
         return (False, comment)
 
     def restart_agent(self, comment: str = "") -> None:
@@ -195,6 +198,13 @@ class TestflingerAgent:
         logger.info("Restarting agent")
         self.set_agent_state(AgentState.OFFLINE, comment)
         sys.exit("Restart Requested")
+
+    def offline_agent(self, comment: str = "") -> None:
+        """Perform the offline action if device is not busy
+        and requested by user.
+        """
+        logger.info("Taking agent offline")
+        self.set_agent_state(AgentState.OFFLINE, comment)
 
     def unpack_attachments(self, job_data: dict, cwd: Path):
         """Download and unpack the attachments associated with a job."""
@@ -246,11 +256,26 @@ class TestflingerAgent:
         # First, see if we have any old results that we couldn't send last time
         self.retry_old_results()
 
-        # Before picking up jobs, validate restart is not needed.
-        needs_restart, comment = self.check_restart()
-        self.restart_handler.update(needs_restart, comment)
-        if self.restart_handler.marked_for_restart():
-            self.restart_agent(self.restart_handler.get_comment())
+        # Before picking up jobs, validate offline and restart are not needed.
+        needs_offline, offline_comment = self.check_offline()
+        needs_restart, restart_comment = self.check_restart()
+
+        # Update status handler, if offline is needed, will prioritize it
+        if needs_offline:
+            self.status_handler.update(
+                offline=needs_offline, comment=offline_comment
+            )
+        elif needs_restart:
+            self.status_handler.update(
+                restart=needs_restart, comment=restart_comment
+            )
+
+        # Offline or restart agent if needed
+        if self.status_handler.marked_for_offline():
+            self.offline_agent(self.status_handler.get_comment())
+            return
+        if self.status_handler.marked_for_restart():
+            self.restart_agent(self.status_handler.get_comment())
 
         job_data = self.client.check_jobs()
         while job_data:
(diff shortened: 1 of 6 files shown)


agent/src/testflinger_agent/agent.py
```diff
@@ -124,66 +127,86 @@ def _post_initial_agent_data(self):
...
+        agent_state, comment = self.get_agent_state()
 
-    def mark_device_offline(self):
-        # Create the offline file, this should work even if it exists
-        open(self.get_offline_files()[0], "w").close()
+        # Offline set by server
+        if (
+            agent_state == AgentState.OFFLINE
+            or agent_state == AgentState.MAINTENANCE
+        ):
```
pedro-avalos, line 169: Might need formatting, but this would be easier to manage in case we ever add more "offline" states

```suggestion
        if agent_state in (
            AgentState.OFFLINE,
            AgentState.MAINTENANCE,
        ):
```

agent/src/testflinger_agent/agent.py
```diff
@@ -124,66 +127,86 @@ def _post_initial_agent_data(self):
...
+            return (True, self.status_handler.get_comment())
+        return (False, comment)
+
+    def restart_agent(self, comment: str = "") -> None:
+        """Perform the restart action if device is not busy
+        and requested by user.
+        """
+        logger.info("Restarting agent")
+        self.set_agent_state(AgentState.OFFLINE, comment)
```
pedro-avalos, line 199: Is this supposed to be `AgentState.RESTART`?

If not, could you add a short explanation as to why `OFFLINE` is used when marking the agent for restart?

agent/src/testflinger_agent/agent.py
```diff
@@ -124,66 +127,86 @@ def _post_initial_agent_data(self):
...
+        sys.exit("Restart Requested")
+
+    def offline_agent(self, comment: str = "") -> None:
+        """Perform the offline action if device is not busy
+        and requested by user.
+        """
+        logger.info("Taking agent offline")
+        self.set_agent_state(AgentState.OFFLINE, comment)
+        # Need to set the offline flag to False to allow recovery
+        self.status_handler.update(offline=False, comment=comment)
```
pedro-avalos, line 209: Are the `offline_agent` and `restart_agent` functions flipped?

agent/src/testflinger_agent/agent.py
```diff
@@ -311,7 +372,9 @@ def process_jobs(self):
                         # exit code 46 is our indication that recovery failed!
                         # In this case, we need to mark the device offline
                         if exit_code == 46:
-                            self.mark_device_offline()
+                            comment = "Set to offline by agent. Recovery "
+                            f"Failed found during {job.job_id} execution."
```
pedro-avalos, line 376: I think this phrasing might be a little easier to read

```suggestion
                            comment = (
                                "Set to offline by agent. Recovery failed"
                                f" during job '{job.job_id}' execution."
                            )
```

agent/src/testflinger_agent/handlers.py
```diff
@@ -31,3 +31,57 @@ def __init__(self, log_file: str):
...
+        """Initialize handler with default values."""
+        self.needs_restart = False
+        self.needs_offline = False
+        self.comment = ""
+
+    def update(
+        self,
+        comment: str,
+        restart: bool | None = None,
+        offline: bool | None = None,
```
pedro-avalos, line 49: Since both of these can be `None` (which is considered the same as `False` in `if` statements by Python) perhaps you should have the following lines before the `if` statements:

```python
if restart is None and offline is None:
    raise ValueError("One of `restart` or `offline` required")
```

agent/src/testflinger_agent/handlers.py
```diff
@@ -31,3 +31,57 @@ def __init__(self, log_file: str):
...
+        """
+        if restart and not self.needs_restart:
+            self.needs_restart = True
+            if not self.needs_offline:
+                self.comment = comment
+        if offline and not self.needs_offline:
+            self.needs_offline = True
+            self.comment = comment
+        # Clear the flag if received an offline False
+        elif not offline and self.needs_offline:
```
pedro-avalos, line 65: Follow-up from the `None` comment I made earlier. If the caller does not specify `offline` (i.e. it's `None`), then should this line execute? Or, in other words, does `offline=None` / not providing `offline` imply that the agent should not be offline?

If that's the case, then I'm not quite sure why `restart` and `offline` are optional/`None`, maybe they should default to `False`, depending on the expected behavior

agent/src/testflinger_agent/handlers.py
```diff
@@ -31,3 +31,57 @@ def __init__(self, log_file: str):
...
+        :return: True if a offline is neeeded, False otherwise.
+        """
+        return self.needs_offline
+
+    def get_comment(self) -> str:
+        """Retrieve the comment from the status handler.
+
+        :return: Preserved comment if an agent status was modified.
+        """
+        return self.comment
```
pedro-avalos, line 87: I feel like these functions and "public" fields can be confusing. If what you want is only "getters" for `needs_restart` and `needs_offline`, I think you may want to use properties:

```python
class AgentStatusHandler:
    def __init__(self):
        self._needs_restart = False
        self._needs_offline = False
        self._comment = ""

    @property
    def needs_restart(self) -> bool:
        """Indicate the current restart state."""
        return self._needs_restart

    @property
    def needs_offline(self) -> bool:
        """Indicate the current offline state."""
        return self._needs_offline

    @property
    def comment(self) -> str:
        """Retrieve the comment on the state."""
        return self._comment
```

Another option is to allow using the instance variables `needs_offline`, `needs_restart`, and `comment` directly, but I do not think that's desirable, as updating them needs some extra logic (thus the `update` function)

pedro-avalos (changes requested):
I have left a few comments regarding the code. There were 2 functions that either need comments or accidentally got flipped (`offline_agent` and `restart_agent`). I also left some questions regarding:

- one minor efficiency improvement
- style suggestions for parameters and class structure

agent/src/testflinger_agent/agent.py
rene-oromtz: Both are correct, restart will terminate agent code with sys.exit and offline will only stop processing jobs (it do needs to clear the flag from the handler, otherwise there is no way to recover the agent without manual intervention)

agent/src/testflinger_agent/handlers.py
rene-oromtz: Ah this is tricky. 
Initially I have them both as `False` however, there is an edge case where if during testing the agent is marked for offline and then, it also receives a restart it will only preserve the restart as the handler will clear the offline flag. The reason is this piece of code:
```
        # Clear the flag if received an offline False
        elif not offline and self._needs_offline:
            self._needs_offline = False
```

This is needed to clear the offline flag after effectively take the agent as offline. 

Another option I can do (if setting both to False is preferred) is that whenever calling the status handler for restart I should also be specifying the current offline flag. Something as:
```
self.status_handler.update(
    restart=needs_restart, 
    offline=self.status_handler.needs_offline, 
    comment=restart_comment
)
```

Let me try this scenario to see if it look cleaner

agent/src/testflinger_agent/handlers.py
rene-oromtz: Thanks for the recommendation! I like this approach and implemented in the last commit

rene-oromtz:
Thanks for the review! I added the suggestions and reformatted a bit for defaulting `offline` and `restart` to False. I tested on Staging and found couple of bugs which were also addressed. 
In case anything else is needed, please let me know!

pedro-avalos (dismissed):
LGTM

Thanks for the changes! I'm so glad to see the `/tmp` files go away

rene-oromtz:
@pedro-avalos I was about to merge this when I realize about a bug in a edge case. 
The scenario is the following:
1. Agent processing job. 
2. Received offline signal
3. Offline successfully deferred until job is no longer processing. 
4. Agent successfully set to offline
5. Agent successfully set back to online/waiting  from the CLI
6. On the next job the UI will display the preserved comment in the status handler as this wasn't cleared!

This is not affecting the agent as it will process the job and the comment will be cleared after completing this second job... But I found it too annoying as the comment is now displayed in the UI. 

Pushing the fix dismissed your review automatically, can you please help me with a re-review ?

pedro-avalos (approved):

PR #755, opened 2025-07-16
docs: Small tutorial update

fhost:
## Description

In the tutorial, there were no reference to the `poll` command, nor to "reserve" jobs.
Adding them with a bit of update (jammy -> noble) and format clean up.

### DIFF
--- a/docs/tutorial/index.rst
+++ b/docs/tutorial/index.rst
@@ -236,7 +236,13 @@ This can be particularly relevant during the ``reserve`` stage as it will provid
       "job_state": "reserve"
   }
 
+You can also view job status and results for a specific queue at ``https://testflinger.example.com/queues/test-queue-1``, where you can select the corresponding job under the “Jobs” list.
 
+You can even monitor the live output of your job using:
+
+.. code-block:: shell
+
+   $ testflinger-cli poll 2bac1457-0000-0000-0000-15f23f69fd39
 
 ---------
 
--- a/docs/tutorial/index.rst
+++ b/docs/tutorial/index.rst
@@ -132,13 +132,13 @@ Test jobs are YAML or JSON files that define the configurations and instructions
 
 A test job might contain a very complex command workflow that includes provisioning a system image onto the device, updating the firmware, executing a series test and more. In this tutorial, you will start with a simple test job.
 
-The following example shows a test job, written in YAML, that provisions an Ubuntu Jammy system image on a MAAS-provisioned device and then prints the distribution information:
+The following example shows a test job, written in YAML, that provisions an Ubuntu Noble system image on a MAAS-provisioned device and then prints the distribution information:
 
 .. code-block:: yaml
 
   job_queue: example-queue-1
   provision_data:
-    distro: jammy
+    distro: noble
   test_data:
     test_cmds: |
       ssh -t ubuntu@$DEVICE_IP lsb_release -a
--- a/docs/tutorial/index.rst
+++ b/docs/tutorial/index.rst
@@ -1,5 +1,5 @@
 Get started with Testflinger CLI
-=================================
+================================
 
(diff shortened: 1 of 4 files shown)


docs/tutorial/index.rst
```diff
@@ -236,14 +242,61 @@ This can be particularly relevant during the ``reserve`` stage as it will provid
       "job_state": "reserve"
   }
 
+You can also view job status and results for a specific queue at ``https://testflinger.example.com/queues/test-queue-1``, where you can select the corresponding job under the “Jobs” list.
+
+Reserve job
```
boukeas, line 247: I am wondering if perhaps this needs to be in one of the separate how-to guides. A new one, just for making reservations. I think perhaps this will make it easier to discover, as users will see it in the left-hand side index.

image

docs/tutorial/index.rst
```diff
@@ -236,14 +242,61 @@ This can be particularly relevant during the ``reserve`` stage as it will provid
...
+-----------
+
+Often, your work will involve manually interacting with a live system rather than simply batch-submitting automated jobs and awaiting results.
+In these cases, we can use Testflinger "reserve" functionality to grant us SSH access (using the key of your LaunchPad or GitHub account) to a test machine for a specified period of time.
```
boukeas, line 251: There is an existing section documenting the reserve phase [here](https://canonical-testflinger.readthedocs-hosted.com/en/latest/reference/test-phases.html#reserve). Please link to that and align descriptions and terminology.

docs/tutorial/index.rst
```diff
@@ -126,35 +126,35 @@ Alternatively, you can also visit the Web UI of this server at ``https://testfli
...
 
-The following example shows a test job, written in YAML, that provisions an Ubuntu Jammy system image on a MAAS-provisioned device and then prints the distribution information:
+The following example shows a test job, written in YAML, that provisions an Ubuntu Noble system image on a MAAS-provisioned device and then prints the distribution information:
 
 .. code-block:: yaml
 
   job_queue: example-queue-1
   provision_data:
-    distro: jammy
+    distro: noble
```
boukeas, line 141: I don't mind this modification but can I ask what is the point of it? How does the example change if it provisions noble instead of jammy?

Also (you may have already done this) can you please make sure this example actually works, i.e. submit it? I seem to remember that `lsb_release` may not be preinstalled on noble.

docs/tutorial/index.rst
```diff
@@ -186,8 +186,14 @@ This command provides you with brief information about the job's current status,
 
 The above output implies that the test job is going through the provisioning phase. If the job is completed, the returned status shows ``complete``.
 
+You can also monitor your job status using:
```
boukeas, line 189: Polling isn't really relevant to job status. It allows the user to monitor the *live output* of the job, as it is running. So I'd say this paragraph doesn't belong in this section. In fact, there's a [separate how-to](https://canonical-testflinger.readthedocs-hosted.com/en/latest/how-to/submit-job.html) that I think should be extended and linked to from the main tutorial.

It's also useful to mention the `--poll` flag that is available when submitting a job. This starts polling right away:
```
testflinger-cli submit --poll test-job.yaml
```

Finally, this is also a common pattern worth including:
```
JOB_ID=$(testflinger submit -q job.yaml)
testflinger poll $JOB_ID
```
Storing the job ID in an environment variable allows us to reuse it without copying-pasting it, e.g. `testflinger results $JOB_ID`, etc

boukeas (changes requested):

docs/tutorial/index.rst
boukeas: I think you are referring to "job status" more loosely here and my objection is that "status" means a very specific thing for Testflinger and this is not it. What polling does is allowing you to monitor the live output from the job so this line should be changed to:

> You can also monitor the live output of your job by using

boukeas (approved):

PR #762, opened 2025-07-23
Modify CLI to support new logging scheme

val500:
## Description
This PR introduces two changes:
1. Agent changes
- The output log handler and serial log handler is moved to be class variables of TestflingerJob. This is so the fragment numbers don't reset between phases.

2. CLI changes
- The logic for poll/poll serial was combined to not duplicate code.
    - `get_output` and `get_serial_output` combined into `get_logs`
- Polling modified to use new logging endpoint
    - Since logs aren't deleted after polling, the CLI needs to keep track of the latest fragment number retrieved.
- Logic to convert json logs retrieved from server to one combined log
- Several new arguments to poll
    - start_fragment, start_timestamp, phase, and json

## Resolved issues
https://warthogs.atlassian.net/browse/CERTTF-619

## Tests
Tests were added to test_cli and test_client


cli/testflinger_cli/__init__.py
```diff
@@ -883,36 +904,113 @@ def artifacts(self):
...
+        log_type: LogType,
+    ):
         """Poll for output from a running job and print it while it runs.
 
         :param str job_id: Job ID
+        :param LogType log_type: Enum representing serial or agent output.
         """
+        start_fragment = self.args.start_fragment
+        start_timestamp = self.args.start_timestamp
+        job_id = self.args.job_id
```
rene-oromtz, line 1005: `job_id` is now removed from do_poll() however, submit/reserve actions does not have `self.args.job_id`
I believe the more reusable way for this method is to keep the `job_id` as a required param but please let me know your thoughts on this.

cli/testflinger_cli/__init__.py
```diff
@@ -925,10 +1023,14 @@ def do_poll(self, job_id):
                         prev_queue_pos = int(queue_pos)
                         print(f"Jobs ahead in queue: {queue_pos}")
                 time.sleep(10)
-                output = ""
-                output = self.get_latest_output(job_id)
-                if output:
-                    print(output, end="", flush=True)
+                last_fragment_number, log_data = self._get_combined_log_output(
+                    job_id, log_type, phase, start_fragment, start_timestamp
```
rene-oromtz, line 1027: I think this is: 
```suggestion
                    job_id, log_type, phase, cur_fragment, start_timestamp
```

rene-oromtz (changes requested):
I added couple enhancements and also pointed some possible bug specifically when calling `do_poll()` please let me know your thought in case I'm missing something.

cli/testflinger_cli/__init__.py
val500: Yeah, that makes sense. Will add that back

cli/testflinger_cli/__init__.py
```diff
@@ -883,36 +904,105 @@ def artifacts(self):
...
+        log_type: LogType = LogType.STANDARD_OUTPUT,
+    ):
         """Poll for output from a running job and print it while it runs.
 
         :param str job_id: Job ID
+        :param LogType log_type: Enum representing serial or agent output.
         """
+        start_fragment = self.args.start_fragment
+        start_timestamp = self.args.start_timestamp
+        job_id = self.args.job_id
```
rene-oromtz, line 997: This is no longer needed as `job_id`  is already included as a required param
```suggestion
```

rene-oromtz (changes requested):
Thanks for the changes and for the additional tests. Overall this looks good to me but seems there is a variable definition that needs to be removed as its no longer needed in `do_poll()`

cli/testflinger_cli/__init__.py
```diff
@@ -883,36 +904,105 @@ def artifacts(self):
...
+        phase = self.args.phase
```
rene-oromtz, line 998: Based on my other comment I realized about the self.args! Submit/reserve don't have this arguments defined, one option could be to include it as params in the `do_poll` but I think this would work just fine:
```suggestion
        start_fragment = getattr(self.args, 'start_fragment', 0)
        start_timestamp = getattr(self.args, 'start_timestamp', None)
        phase = getattr(self.args, 'phase', None)
```

Please let me know what you think.

rene-oromtz (approved):
LGTM, thanks for addressing the required changes

PR #765, opened 2025-07-29
Add common module to testflinger server

val500:


val500:
> Changes look good to me, if any, I would probably also include the changes in the `server/docker-compose.yml` file as otherwise this will be unusable while I'm not sure if anyone besides me uses that file

@rene-oromtz I believe the normal docker compose in the server pulls the already built docker image so I don't think that needs extra references to the common repo.

rene-oromtz:
@val500 I think testflinger dev image for docker compose is built locally as there is no image for `testflinger:latest` in Docker hub.  

Looking at the charm, seems the TF image is in GHCR but this is not being referenced by the docker compose file: `ghcr.io/canonical/testflinger:main`

rene-oromtz (approved):
LGTM! Thanks for including the changes on regular docker compose file

PR #770, opened 2025-07-31
Modify agent charm configuration of grafana

val500:
## Description
Explicitly specified metrics path and localhost for grafana agent scrape configs.

## Tests
Tested in Staging


rene-oromtz (approved):
LGTM

PR #771, opened 2025-08-01
feat(server): add client permissions related endpoints

rene-oromtz:
## Description
This PR introduces the required changes in the API to create, edit, delete, show client_permissions from the API. 
Both the output and the input are schema validated. To avoid potential collisions, the `allow_queues` logic is removed from the allowed input, this should be taken care of by using the `restricted-queues` endpoint. 

## Resolved issues
Resolves [CERTTF-679](https://warthogs.atlassian.net/browse/CERTTF-679)

## Web service API changes
The following endpoints and methods are introduced to support the client_permissions creation directly from the API:
 
- `GET /v1/client-permissions`
- `GET /v1/client-permissions/<client_id>`
- `POST /v1/client-permissions`
- `PUT /v1/client-permissions/<client_id>`
- `DELETE /v1/client-permissions/<client_id>`

All above endpoints were documented in `API.md` along with the restricted-queues endpoints which were also missing from docs. 

## Tests
Added unit tests to support this logic and also validated locally

[CERTTF-679]: https://warthogs.atlassian.net/browse/CERTTF-679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/src/testflinger/api/auth.py
+++ b/server/src/testflinger/api/auth.py
@@ -285,3 +285,25 @@ def require_role(*roles):
         return wrapper
 
     return decorator
+
+
+def check_role_hierarchy(user_role: str, target_role: str) -> bool:
+    """
+    Check role of current user to validate if authorized to perform action.
+
+    :param user_role: Role of the user making the request.
+    :param target_role: Role being assigned or modified.
+    :return: True if allowed, False otherwise.
+    """
+    role_levels = {
+        ServerRoles.USER: 1,
+        ServerRoles.CONTRIBUTOR: 2,
+        ServerRoles.MANAGER: 3,
+        ServerRoles.ADMIN: 4,
+    }
+
+    current_level = role_levels.get(user_role, 0)
+    target_level = role_levels.get(target_role, 0)
+
+    # Users can modify roles at their level or below
+    return current_level >= target_level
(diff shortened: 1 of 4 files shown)


server/API.md
```diff
@@ -444,3 +444,230 @@ Example:
 curl http://localhost:8000/v1/oauth2/token \
   -X GET --header "Authorization: Basic ABCDEF12345"
 ```
+
+## `[GET] /v1/restricted-queues`
+
+Retrieve the list all agents restricted queues.
```
diohe0311, line 450:
```suggestion
Retrieve the list of all restricted queues and their owners.
```

diohe0311 (changes requested):
Hi Rene, thanks a lot for updating the API doc, looks really helpful! Appreciate you jumping in to patch this up

I just did a quick read and everything looks great overall. A couple of minor nitpicks below:

diohe0311 (commented):
Hi Rene, I think it's generally clean and well-structured. Just one security concern:

Since both `ADMIN` and `MANAGER` roles can access the `/client-permissions` POST and PUT endpoints, it’s seems possible for a `MANAGER` to assign the `ADMIN` role to a client. This could lead to privilege escalation, so we might want to prevent that.

rene-oromtz:
Ah thanks a lot for pointing this out @diohe0311! I just added this restriction on the last commit along with some Unit Tests for this added functionality, now only manager account can edit/create client_ids within its level or lower level.

PR #777, opened 2025-08-09
fix(agent): Bug fix for unhandled exception

rene-oromtz:
## Description
Agents are showing as EXITED due to an unhandled exception when processing a job. 

During any test phase, the agent is resilient enough to handle HTTP errors and continue job processing, however, when getting to the cleanup phase, this forces the agent to raise a TFServerError Exception that is not handled properly. 

Traceback: 
```
requests.exceptions.RetryError: HTTPSConnectionPool(host='testflinger.canonical.com', port=443): Max retries exceeded with url: /v1/result/1dd65a8a-e1a0-4c89-8990-7ac7f8479156

  File "/srv/testflinger-venv/lib/python3.10/site-packages/testflinger_agent/__init__.py", line 143, in start_agent
    agent.process_jobs()
  File "/srv/testflinger-venv/lib/python3.10/site-packages/testflinger_agent/agent.py", line 400, in process_jobs
    exit_code, _, _ = job.run_test_phase(TestPhase.CLEANUP, rundir)
  File "/srv/testflinger-venv/lib/python3.10/site-packages/testflinger_agent/job.py", line 128, in run_test_phase
    self.client.post_result(self.job_id, device_info)
  File "/srv/testflinger-venv/lib/python3.10/site-packages/testflinger_agent/client.py", line 198, in post_result
    raise TFServerError("other exception") from exc
testflinger_agent.errors.TFServerError: HTTP Status: other exception
```

The `client.post_result()` in `run_test_phase()` was not wrapped in a try-except block and the only reason why the other phases work are due to the general exception... Cleanup is handled differently in a `finally` block

Additionally to handling the TFServerError, this PR also addressed the following when the server is unreachable:

Before job:
The agent will simply go to sleep until the server is reachable again, in the meantime, will log that the server couldn't be reached.

During job:
Since now at every stage it attempts to get the status of the agent, if its unable to do it, it won't assume is offline, it will get an "unknown" status that is never sent to the server, this is just for agent resiliency, otherwise the agent will be marked as offline and will require manual intervention.
Additionally, the job will not be able to "finish" unless its able to clear the job_id from the database and send its status to the server. This to avoid any inconsistency.

## Resolved issues
Resolves [CERTTF-689](https://warthogs.atlassian.net/browse/CERTTF-689)

## Tests

[CERTTF-689]: https://warthogs.atlassian.net/browse/CERTTF-689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/src/testflinger_agent/__init__.py
+++ b/agent/src/testflinger_agent/__init__.py
@@ -127,7 +127,7 @@ def start_agent():
     agent = TestflingerAgent(client)
     while True:
         # Only check agent status and process jobs if server is available
-        if client.is_server_reacheable():
+        if client.is_server_reachable():
             is_offline, offline_comment = agent.check_offline()
             if is_offline:
                 logger.error(
--- a/agent/src/testflinger_agent/__init__.py
+++ b/agent/src/testflinger_agent/__init__.py
@@ -126,23 +126,28 @@ def start_agent():
     client = TestflingerClient(config)
     agent = TestflingerAgent(client)
     while True:
-        is_offline, offline_comment = agent.check_offline()
-        if is_offline:
-            logger.error(
-                "Agent %s is offline, not processing jobs! "
-                "Reason: %s\n"
-                "Please contact Tesflinger Admin if you require assistance.",
-                config.get("agent_id"),
-                offline_comment or "Unknown",
-            )
-            while agent.check_offline()[0]:
-                time.sleep(check_interval)
-        # Refresh the updated_at timestamp on advertised queues
-        client.post_advertised_queues()
-        logger.info("Checking jobs")
-        agent.process_jobs()
-        logger.info("Sleeping for %d", check_interval)
-        time.sleep(check_interval)
+        # Only check agent status and process jobs if server is available
+        if client.is_server_reacheable():
+            is_offline, offline_comment = agent.check_offline()
+            if is_offline:
+                logger.error(
+                    "Agent %s is offline, not processing jobs! "
+                    "Reason: %s\n"
+                    "Please contact TF Admin if you require assistance.",
+                    config.get("agent_id"),
+                    offline_comment or "Unknown",
+                )
+                while agent.check_offline()[0]:
+                    time.sleep(check_interval)
+            # Refresh the updated_at timestamp on advertised queues
+            client.post_advertised_queues()
+            logger.info("Checking jobs")
+            agent.process_jobs()
+            logger.info("Sleeping for %d", check_interval)
+            time.sleep(check_interval)
+        else:
+            # Waiting until we restore connectivity
+            client.wait_for_server_connectivity()
 
 
 def load_config(configfile):
--- a/agent/src/testflinger_agent/__init__.py
+++ b/agent/src/testflinger_agent/__init__.py
@@ -127,27 +127,25 @@ def start_agent():
     agent = TestflingerAgent(client)
     while True:
         # Only check agent status and process jobs if server is available
-        if client.is_server_reachable():
-            is_offline, offline_comment = agent.check_offline()
-            if is_offline:
-                logger.error(
-                    "Agent %s is offline, not processing jobs! "
-                    "Reason: %s\n"
-                    "Please contact TF Admin if you require assistance.",
-                    config.get("agent_id"),
-                    offline_comment or "Unknown",
-                )
-                while agent.check_offline()[0]:
-                    time.sleep(check_interval)
-            # Refresh the updated_at timestamp on advertised queues
-            client.post_advertised_queues()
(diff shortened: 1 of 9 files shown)


agent/src/testflinger_agent/job.py
```diff
@@ -125,7 +125,16 @@ def run_test_phase(self, phase, rundir):
...
+            try:
+                # raises TFServerError upon HTTP failure
+                self.client.post_result(self.job_id, device_info)
+            except TFServerError as exc:
+                logger.warning(
+                    "Failed to post device info during %s for job %s: %s",
+                    phase,
+                    self.job_id,
+                    exc,
+                )
```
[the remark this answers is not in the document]
rene-oromtz: After checking, seems that all 5xx were already retried in the definition of the client Session:
```
        retry = Retry(
            total=retries,
            read=retries,
            connect=retries,
            backoff_factor=0.3,
            status_forcelist=(500, 502, 503, 504),
            allowed_methods=False,  # allow retry on all methods
        )
```
But I did ended up adding another approach to handle this scenario. Will update the description and add another comment about this.

agent/src/testflinger_agent/client.py
```diff
@@ -481,3 +481,40 @@ def post_status_update(
...
+            if self.is_server_reachable():
+                logger.info("Testflinger server connectivity restored")
+                break
+            else:
+                logger.warning(
+                    "Testflinger server unreachable, waiting for connectivity"
+                )
+                # Exponentially increase interval between rechecks
+                interval = min(interval * (2**retry_count), max_interval)
+                retry_count += 1
```
pedro-avalos, line 520: How about we move the `time.sleep` to the end of the loop?

```suggestion
            if self.is_server_reachable():
                logger.info("Testflinger server connectivity restored")
                break
            else:
                logger.warning(
                    "Testflinger server unreachable, waiting for connectivity"
                )
                time.sleep(interval)
                # Exponentially increase interval between rechecks
                interval = min(interval * (2**retry_count), max_interval)
                retry_count += 1
```

agent/src/testflinger_agent/agent.py
```diff
@@ -417,6 +415,13 @@ def process_jobs(self):
                 results_basedir = self.client.config.get("results_basedir")
                 shutil.move(rundir, results_basedir)
 
+            # Complete cleanup only if server is reachable
+            if not self.client.is_server_reachable():
+                self.client.wait_for_server_connectivity()
```
pedro-avalos, line 420: The `wait_for_server_connectivity` alone would be able to check this, right?

```suggestion
            self.client.wait_for_server_connectivity()
```

pedro-avalos (changes requested):
This looks great and resilient. Thank you for the changes!

I have left a suggestion that could cut a couple lines.

agent/src/testflinger_agent/client.py
```diff
@@ -481,3 +481,39 @@ def post_status_update(
...
+            if self.is_server_reachable():
+                break
+            else:
+                logger.warning(
+                    "Testflinger server unreachable, waiting for connectivity"
+                )
+                time.sleep(interval)
+                # Exponentially increase interval between rechecks
+                interval = min(interval * (2**retry_count), max_interval)
+                retry_count += 1
```
pedro-avalos, line 519: I missed this last time, but we can get rid of the `else` since the `if` block `break`s

```suggestion
        while True:
            if self.is_server_reachable():
                break
            logger.warning(
                "Testflinger server unreachable, waiting for connectivity"
            )
            time.sleep(interval)
            # Exponentially increase interval between rechecks
            interval = min(interval * (2**retry_count), max_interval)
            retry_count += 1
```

agent/src/testflinger_agent/__init__.py
```diff
@@ -126,12 +126,15 @@ def start_agent():
...
+        # Only check agent status and process jobs if server is available
+        client.wait_for_server_connectivity()
+
         is_offline, offline_comment = agent.check_offline()
         if is_offline:
             logger.error(
                 "Agent %s is offline, not processing jobs! "
                 "Reason: %s\n"
-                "Please contact Tesflinger Admin if you require assistance.",
+                "Please contact TF Admin if you require assistance.",
```
pedro-avalos, line 137: Nitpick: I think this should still be `Testflinger` in my opinion. Feel free to ignore if you think otherwise

pedro-avalos (changes requested):
Some minor nitpicks, I will approve after these

pedro-avalos (approved):
Thanks for this!

PR #780, opened 2025-08-13
feat(cli): extend cli client module with additional methods

rene-oromtz:
## Description
As the server has now extended functionality, it now accepts the following methods:
`PUT`
`DELETE`

Adding this methods for sending correct requests via CLI. Also modified current `POST` (which was using `put`) to match the proper request. 

Added also an improvement while retrieving error messages from Schema validation (UNPROCESSABLE_ENTITY) and `apiflask.abort()` this should help to make clearer the failure reason while executing command. 

Sample response retrieve from server due to schema validation:
`{"detail": {"json": {"max_priority": ["Missing data for required field."], "max_reservation_time": ["Missing data for required field."]}}, "message": "Validation error"}`

Sample response retrieved from any other non 422 error aborted by server with apiflask:
`{"detail": {}, "message": "Specified action requires role: admin"}`

## Resolved issues
Related to #779 
Related to [CERTTF-683](https://warthogs.atlassian.net/browse/CERTTF-683)

## Tests
Added additional unit tests to client test module. 
DELETE and PUT methods will be implemented in #779 . 

[CERTTF-683]: https://warthogs.atlassian.net/browse/CERTTF-683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/testflinger_cli/client.py
+++ b/cli/testflinger_cli/client.py
@@ -21,6 +21,7 @@ import json
 import logging
 import sys
 import urllib.parse
+from http import HTTPStatus
 from pathlib import Path
 
 import requests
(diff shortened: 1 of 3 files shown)


cli/testflinger_cli/client.py
```diff
@@ -66,16 +70,47 @@ def get(self, uri_frag, timeout=15, headers=None):
                     exc,
                 )
             raise
-        if req.status_code != 200:
-            raise HTTPError(req.status_code)
+        if req.status_code != HTTPStatus.OK:
+            try:
```
val500, line 74: Is this the same error status parsing as the PUT/POST? If so, should it be put into a separate function?

rene-oromtz: Yes, this can be used also by the `get` because of the apiflask abort is also present in some of the endpoints that uses this method. I agree with you, since all methods use this, I can add a separate function for this logic.

PR #786, opened 2025-08-25
fix(oem_autoinstall): Default stock default-user-data to match raid storage

Artur-at-work:
## Description
Changes to default user-data configs in oem_autoinstall connector:
 - stock installer from media to match raid storage first
 - fix autologin
 - change location of set_usb_boot.sh to keep for later use

## Tests
Run provision from local agent with updated user-data. Verify stock installer can complete on platform with RAID1


KaiChuan-Hsieh (approved):
LGTM

PR #788, opened 2025-08-26
fix(server): fix a typo and expected client_id type

diohe0311:
@rene-oromtz just spotted a small typo, sorry I didn't catch it earlier.


rene-oromtz (approved):
Thank you for noticing this and for addressing this typo!

PR #792, opened 2025-08-29
feat: add OAuth2 refresh and revoke token endpoints

diohe0311:
This PR resolves https://warthogs.atlassian.net/browse/CERTTF-684

## Description
This PR introduces support for OAuth2 refresh tokens, allowing clients to obtain new access tokens without re-authenticating, and enables administrators to revoke or extend refresh tokens when necessary. 

## Implementation
- Updated /v1/oauth2/token, now returns both `access_token` and `refresh_token`
- Added /v1/oauth2/refresh, exchange a valid refresh token for a new access token
- Added /v1/oauth2/revoke, revoke a refresh token to prevent further use
- Added database handling for refresh tokens (add, get, edit and delete)
- Added test cases
- Updated API.md

## Web service API changes
[CR123 - Testflinger OAuth2 Integration](https://docs.google.com/document/d/17oS3oy-ZWu79PyvJ8vokTJgUdln9VeXj4J6knl3_SHM/edit?tab=t.0)

## Tests
Tested locally


server/src/testflinger/api/v1.py
```diff
@@ -827,9 +830,62 @@ def retrieve_token():
     allowed_resources = auth.validate_client_key_pair(client_id, client_key)
     if allowed_resources is None:
         return "Invalid client id or client key", 401
+
+    secret_key = os.environ.get("JWT_SIGNING_KEY")
+    access_token = auth.generate_access_token(allowed_resources, secret_key)
+
+    refresh_token = auth.generate_refresh_token(client_id)
```
rene-oromtz, line 837: Another thing I noticed, is that it by default will set the 30 days expiration which needs to be `None` for the required LabManager/Admin roles. 

I'm thinking maybe we can pass optional data to this endpoint to set a specified expiration (default/max 30days)
`--data '{"refresh_token_expires_in": null}'`

This of course will need to double check that the authenticated user can have a non expiry token an either reject the request if not authorized to set this or limit it to 30 days. 

If we need to inspect the `client_id`, we may optionally do it simpler by just getting the role of the user and set a None for admin/manager roles automatically. Since revocation is already in place, I don't see any major concern by doing this automatically. 

What do you think?

diohe0311:
@rene-oromtz, you’re right, I forgot about the non-expiring tokens. `last_accessed` make sense. I’ve added the TTL accordingly.

rene-oromtz (approved):
Thanks for adding the TTL, this now looks good to me! 

From my perspective, this can be merged but Charm can only be promoted to stable until CLI implement the required changes in [CERTTF-706](https://warthogs.atlassian.net/browse/CERTTF-706). I'm currently working on it based on the work in this PR so hope to got it merged early next week.

PR #798, opened 2025-09-09
Fix: default-user-data for stock to enable autologin in late-commands

Artur-at-work:
## Description
This change in user-data moves updating the /etc/gdm3/custom.conf from runcmd to late-commands because first boot after install doesn't load the updated custom.conf (only after gdm3 restart or reboot it becomes effective). Thus we need to edit the custom.conf before the first boot. 

## Tests
Option 1:
1. Inject updated user-data to stock installer 24.04.3 and verify autologin enabled on first boot
 
Option 2:
1. Update default-user-data and run agent locally
4. Run stock provisioning and verify autologin enabled on first boot


KaiChuan-Hsieh (approved):
LGTM

PR #801, opened 2025-09-17
[CERTTF-441][CERTTF-444] Server handling of job containing secrets

boukeas:
## Description

This PR extends the `/job` POST and GET endpoints to support secrets for the `test` phase.

- The `/job` POST endpoint is used by clients to submit jobs. Its input schema is extended so that the `test_data` can contain a `secrets` field, mapping the names of environment variables to secrets paths (example to follow). When a client submits a job that contains secrets, the job is only accepted if all the secrets specified are accessible in the secrets store.
- The `/job` GET endpoint is used by agents to request the next job they will execute. When the job that will be assigned to an agent contains secrets, the server retrieves these secrets and resolves them in the job (replaces the paths with the corresponding values) before returning the job to the agent.

> [!IMPORTANT]
> At the moment, this can only be tested locally. In order for the new endpoints to be functional in staging or
production, a Vault instance would need to be deployed alongside the Testflinger server (see [CERTTF-387](https://warthogs.atlassian.net/browse/CERTTF-387)). However, note that no assumption has been made around the existence of a Vault instance in the environment, i.e. this is a non-breaking change and all other Testflinger functionality should remain unaffected by merging this change.

## Resolved issues

Resolves CERTTF-441. Partially resolves CERTTF-444 (no CLI changes).

## Documentation

The secrets-related functionality will be fully documented once the whole of CERTTF-314 epic is resolved.

## Web service API changes

Detailed above.

## Tests

Aside from the unit tests introduced, the following actions were performed to test the endpoint _locally_:

```shell
# cleanup of any existing containers
$ docker compose down

# build and run service containers
$ docker compose up -d --build

# populate the database with the data of a single client
# e.g. select "Create client", use "certification" for the client ID and client secret and 
# answer "no" to the rest of the questions
$ docker exec -it testflinger-server client_credentials_admin

# create the test job
$ cat << EOF > test-job-secrets.yaml
job_queue: myqueue
test_data:
  secrets:
    SUPERSECRET: path/to/secret
  test_cmds: |
    echo $SUPERSECRET
    [ "$SUPERSECRET" = "your-secret-value" ] && echo "Secret confirmed"
EOF

# attempt to submit the job: failing because the secret is inaccessible
$ testflinger-cli --server http://localhost:5000/ --poll test-job-secrets.yaml --client-id certification --secret-key certification
Unexpected error status from testflinger server: [422] Inaccessible secret paths: path/to/secret

# retrieve the authentication token
$ TOKEN=$(curl -s -X POST -H "Content-Type: application/json" -u "certification:certification" "http://localhost:5000/v1/oauth2/token" | jq -r '.access_token')

# send a request to the new endpoint to store a secret
$ curl -s -X PUT \
-H "Authorization: Bearer ${TOKEN}" \
-d '{"value":"your-secret-value"}' \
http://localhost:5000/v1/secrets/certification/path/to/secret
OK

# resubmit the job
$ testflinger-cli --server http://localhost:5000/ submit test-job-secrets.yaml --client-id certification --secret-key certification
Job submitted successfully!
job_id: 855fd37f-b4cf-4670-9f20-e54b917de5fd
This job is waiting on a node to become available.

# request a job from the queue as an agent would
# confirm that the secret has been resolved and included in the response
curl -s -X GET \
http://localhost:5000/v1/job/?queue=myqueue | jq -rc '.test_data.secrets'
{"SUPERSECRET":"your-secret-value"}
```

[CERTTF-387]: https://warthogs.atlassian.net/browse/CERTTF-387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/src/testflinger/api/v1.py
+++ b/server/src/testflinger/api/v1.py
@@ -119,6 +119,8 @@ def validate_secrets(data: dict):
         )
 
     # check that all secrets paths correspond to stored secrets
+    # (i.e. a job containing secrets cannot be submitted unless all its secrets
+    # are accessible.)
     inaccessible_paths = []
     for secret_path in secrets.values():
         try:
@@ -201,7 +203,15 @@ def job_get():
 
 
 def retrieve_secrets(data: dict) -> dict | None:
-    """Retrieve all secrets from the secrets store."""
+    """
+    Retrieve all secrets from the secrets store.
+
+    Any secrets that are not accessible at the time of retrieval will be
+    resolved to the empty string, instead of the retrieval failing.
+    It is the responsibility of the consumer of the secrets to account for
+    this possibility. This is a design decision and it mirrors how undefined
+    secrets are handled in other platforms such as GitHub.
+    """
     try:
         secrets = data["test_data"]["secrets"]
     except KeyError:
--- a/server/src/testflinger/api/v1.py
+++ b/server/src/testflinger/api/v1.py
@@ -82,6 +82,9 @@ def job_post(json_data: dict):
         job_queue = ""
     if not job_queue:
         abort(422, message="Invalid data or no job_queue specified")
+
+    validate_secrets(json_data)
+
     try:
         job = job_builder(json_data)
     except ValueError:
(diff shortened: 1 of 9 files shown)


server/src/testflinger/api/v1.py
```diff
@@ -149,12 +192,48 @@ def job_get():
...
+
+def retrieve_secrets(data: dict) -> dict | None:
+    """
+    Retrieve all secrets from the secrets store.
+
+    Any secrets that are not accessible at the time of retrieval will be
+    resolved to the empty string, instead of the retrieval failing.
+    It is the responsibility of the consumer of the secrets to account for
+    this possibility. This is a design decision and it mirrors how undefined
+    secrets are handled in other platforms such as GitHub.
```
rene-oromtz, line 213: Note for the future, we need to add this to the official docs once secrets are fully supported. Sometimes we fall behind in the docs and this is definitively a good warning for the end user.

rene-oromtz (commented):
Just added a note for the future, besides that, the code looks good. 
Looking forward for the agent/device logic, that definitively will be interesting

server/src/testflinger/api/v1.py
boukeas: I've added a note to the relevant section of the API.md file.

rene-oromtz (approved):
Thanks for the changes. LGTM!

PR #803, opened 2025-09-21
Simplify the argparser setup in testflinger-cli

plars:
Hello from the dark side! (still looking for those cookies)
I was feeling nostalgic and decided to go tidy some things up. Feel free to take this or not. I'm really happy to see how nicely everything seems to still be progressing with this project! :)

## Description

There's a lot of boilerplate and extra functions needed to setup argparse, due to all the subcommands and different args needed. This allows us to define them up front in a JSON blob that gets used to setup argparse.

## Resolved issues

Mainly it just makes it easier to read and modify, and removes a lot of boilerplate as mentioned above. 

## Documentation
N/A

## Web service API changes
N/A

## Tests
The tests all still pass, and to be clear, this doesn't intend to change the way the cli args work at all. It just cleans it up a bit and makes it easier to work with.

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -59,6 +59,186 @@ from testflinger_cli.errors import (
 
 logger = logging.getLogger(__name__)
 
+# Define all args we'll allow with argparse
+ARGUMENTS = {
+    "artifacts": {
+        "help": "Download a tarball of artifacts saved for a specified job",
+        "func": "artifacts",
+        "args": [
+            {
+                "name": "--filename",
+                "type": helpers.parse_filename,
+                "default": "artifacts.tgz",
+            },
+            {"name": "job_id", "completer": "job_ids"},
+        ],
+    },
+    "cancel": {
+        "help": "Tell the server to cancel a specified JOB_ID",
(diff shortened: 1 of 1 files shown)


ajzobro:
I would like to see tests that go along with something like this as every argument is being touched.
Furthermore, this abstracts things from their use by creating magic lists (e.g. ```"args": [
            {
                "name": ["--oneshot", "-o"],
                "action": "store_true",
                "help": "Get latest output and exit immediately",
            },
            {"name": "job_id", "completer": "job_ids"},
        ]```

With no supporting documentation as to how they relate or will be parsed.

ajzobro:
@plars Thanks for all the work you did on this project, I hope that you are content with the direction it is heading in.

We are undergoing some changes with respect to the CLI including developing a python API and refactoring the CLI to make use of it.

plars:
No worries! feel free to close it if it's not helpful. It was something I was experimenting with and thought it might have some utility but it looks like some addition arg parsing changes shortly after I proposed it that might be harder to model like this. I haven't had time to go look into it further, but if it was only intended as a possible quality of life improvement. If it doesn't help, then there's no reason to keep it.

PR #804, opened 2025-09-21
Agent tests reduce sleeps

plars:
## Description

Some of the unit tests unfortunately require running something with the agent that does a sleep, in order to test timeouts, etc. This is also impacted by the 10s output polling delay, because if you sleep for <10s then the thing the agent runs will exit before it gets a chance to timeout.  This allows adjusting that output polling delay through a config option, which you may want to do anyway if you don't like that default.

And no, this isn't something you can just get around with things like freezegun. It's relying on a long running subprocess kicked off by the agent.

It also improves the time it takes to run the unit tests that require sleeps.

## Resolved issues

No real issue, just trying to speed things up a bit

## Documentation

Added docs to the agent reference section for the new output_polling_interval option, and adjusted the one for "polling_interval" to make it clear that it was about polling the server for new jobs. Hopefully this makes the difference pretty clear.

## Web service API changes
N/A

## Tests
This allowed adjusting the unit tests, so that was the point :)


ajzobro:
PR #988 adds freezegun which can be further used in time-sensitive testing.

ajzobro (dismissed):
Thanks for the contribution, @plars . Looks harmless enough.  Can you merge in the latest `main` and we can take this.

agent/src/testflinger_agent/job.py
```diff
@@ -67,10 +67,17 @@ def __init__(self, job_data, client):
         )
 
     def get_runner(self, rundir: str, phase: TestPhase):
+        output_polling_interval = self.client.config.get(
+            "output_polling_interval", 10.0
+        )
```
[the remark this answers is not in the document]
ajzobro: @copilot apply changes based on [this feedback](https://github.com/canonical/testflinger/pull/804#discussion_r3174009935)

sane lower bound to be 0.1 in order to enable testing

agent/tests/test_job.py
ajzobro:
```suggestion
        runner = CommandRunner(tmp_path, env={}, output_polling_interval=0.1)
```

agent/src/testflinger_agent/job.py
```diff
@@ -67,10 +67,17 @@ def __init__(self, job_data, client):
...
         try:
```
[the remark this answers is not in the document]
rene-oromtz: @ajzobro copilot may be partially right on this one (not on the suggested change but on the schema part. 

Since this now can be configurable, we may want to define it in [schema.py](https://github.com/canonical/testflinger/blob/main/agent/src/testflinger_agent/schema.py) with float defaulting to 10.0

agent/tests/test_job.py
```diff
@@ -131,9 +129,7 @@ def test_no_output_timeout_in_provision(
         timeout_str = "complete\n"
         logfile = tmp_path / "provision.log"
         fake_job_data = {"output_timeout": 1, "provision_data": {"url": "foo"}}
-        self.config["provision_command"] = (
-            "bash -c 'sleep 12 && echo complete'"
-        )
+        self.config["provision_command"] = "bash -c 'sleep 2 && echo complete'"
```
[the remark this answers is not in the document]
rene-oromtz: @ajzobro this test is also missing the new option as per copilot finding.

rene-oromtz:
@ajzobro there are a couple of copilot suggestions that may make sense, let me know what you think! For the fractional comment I don't really think that is much needed as the schema should also be self explanatory but let me know what you think!

agent/src/testflinger_agent/job.py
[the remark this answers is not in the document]
ajzobro: Agreed: should be defaulted at the API not the client.

ajzobro:
> @ajzobro there are a couple of copilot suggestions that may make sense, let me know what you think! For the fractional comment I don't really think that is much needed as the schema should also be self explanatory but let me know what you think!

I think that both suggested fixes are not right/needed.

1) default value -> belongs in the API not the client
2) I think the documentation explanation was fine -- fractional values are more for testing than for standard use, though they will work we don't need to showcase that.

agent/src/testflinger_agent/job.py
```diff
@@ -67,10 +67,17 @@ def __init__(self, job_data, client):
         )
 
     def get_runner(self, rundir: str, phase: TestPhase):
+        output_polling_interval = self.client.config.get(
+            "output_polling_interval", 10.0
```
ajzobro: @rene-oromtz  The default is established here.

agent/src/testflinger_agent/runner.py
```diff
@@ -48,13 +48,19 @@ class CommandRunner:
     known event types are defined in RunnerEvents.
     """
 
-    def __init__(self, cwd: Optional[str], env: Optional[dict]):
+    def __init__(
+        self,
+        cwd: Optional[str],
+        env: Optional[dict],
+        output_polling_interval: float = 10.0,
```
ajzobro, line 55: @rene-oromtz  default also established here.

agent/src/testflinger_agent/job.py
rene-oromtz: Yes, but this is taken from `self.client.config` and based on Description:

>  This allows adjusting that output polling delay through a config option, which you may want to do anyway if you don't like that default.

Without the schema, we will not be able to change the value via config (not saying that we would change it in our configs).

ajzobro: @copilot the `agent/src/testflinger_agent/schema.py` file needs to have this new `output_polling_interval` added as an optional parameter with a default of 10.

agent/tests/test_job.py
ajzobro, line 133:
```suggestion
        self.config["output_polling_interval"] = 0.1
        self.config["provision_command"] = "bash -c 'sleep 2 && echo complete'"
```

agent/src/testflinger_agent/job.py
ajzobro: Schema validation ensures a default will be available.
```suggestion
        output_polling_interval = self.client.config["output_polling_interval"]
```

PR #805, opened 2025-09-22
[IQA-3167] Add flag for full agent data JSON output

wctaylor:
## Description

The CLI queries the `agents/data/<agent>` endpoint but discards some of the information. This PR introduces the `--all` flag for the `agent-status` command to expose the full information when combined with the `--json` flag.

## Resolved issues

Resolves: IQA-3167

The Solutions QA team makes use of Testflinger in some of our deployments, and we make use of the `"provision_streak_count"` and `"provision_streak_type"` keys in the agent data output. These are not exposed by the CLI currently, so we query the API for this information, and then otherwise rely on the CLI for everything else. Since it would be nice to stick with one approach, I wanted the CLI to be able to expose the full information as well. 

I'm a bit curious why the command discards some information currently, but since that choice was already made, I thought the best way to handle things would be to add an additional option, since that way, no existing usage would be affected. 

## Tests

I added a test for the plain `--json` flag, which otherwise didn't seem tested, and I also added a test for the combination of `--json` and `--all`.

### DIFF
--- a/cli/testflinger_cli/tests/test_cli.py
+++ b/cli/testflinger_cli/tests/test_cli.py
@@ -658,9 +658,8 @@ def test_agent_status_json(capsys, requests_mock):
         "provision_streak_type": "pass",
     }
     requests_mock.get(URL + "/v1/agents/data/" + fake_agent, json=fake_return)
-    sys.argv = ["", "agent-status", fake_agent]
+    sys.argv = ["", "agent-status", fake_agent, "--json"]
     tfcli = testflinger_cli.TestflingerCli()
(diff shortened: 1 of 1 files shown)


boukeas (changes requested):
I think the consensus is that there's no need for the `--json` output to be limited to only a couple of fields. So just do away with `--all` and make the `--json` output complete (i.e. to coincide with `--all`). 

Unfortunately, in the existing behaviour for `--json`, the `name` field is actually renamed to `agent`. Please keep to that, to avoid making this a breaking change.

PR #806, opened 2025-09-22
feat(server): add general oidc integration

rene-oromtz:
## Description
This PR is meant as an initial approach for OpenID Connect integration for SSO. 
More changes will be defined once this initial work is completed and we can request required parameters to set Ubuntu One in staging and production via IS ticket. 

For development purposes, a generic identity provider was added by using [Dex](https://github.com/dexidp/dex). This IdP allows a quick and lightweight setup to configure OIDC and create a sample user. 

> [!NOTE]
> In dev environment, an entry to `/etc/hosts` must be added to allow pointing localhost to dex container and callback redirection to succeed. 
>```
>127.0.0.1 localhost dex
>```

The current approach should not break current application as the OIDC routes will only be enabled upon setting all required variables.

Upon successful user authentication, if it is a new user it will register it in the database, otherwise, will only update the last_accessed entry when it logs in again. 

## Resolved issues
Resolves [CERTTF-707](https://warthogs.atlassian.net/browse/CERTTF-707)

## Documentation
No documention changes for now, once Ubuntu One is added as provider, will document login procedure in public docs. 

## Web service API changes
If OIDC is configured, the following routes will be available from web app:
```
/auth/login
/auth/logout
/auth/callback
```

## Tests
Added new module for oidc related tests. This is using pytest_iam to mock the OIDC integration.

This was also tested locally, when any of the variables is missing, it won't register the app with OIDC and then headers will not be shown. 

https://github.com/user-attachments/assets/1e0d4686-271e-43d5-b61b-bbde2dac335c

[CERTTF-707]: https://warthogs.atlassian.net/browse/CERTTF-707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pedro-avalos (approved):
Thanks for the changes. This looks good to me

PR #808, opened 2025-09-23
fix(cli): resolve login issues caused by stored token

### LINKED ISSUE #807: Login bug when using multiple contexts
Problem description
----

When using multiple environments e.g. dev, production, staging, the cli will try to use by default the stored refresh token. 
Since this refresh token is unique per environment, the stored refresh token needs to be updated to match the environment.

The `login` command should be able to clear the refresh token and switch context, however, due to the fact that the auth module is initialized during CLI constructor, it will take two `login` attempts to succeed. 
First attempt will fail because the token was invalid but it will effectively remove the refresh token from SNAP_USER_DATA. 
Second attempt will then succeed as the refresh token was removed on earlier request.

On another similar issue, you can be using mainly one set of credentials e.g `testflinger-admin` and when trying to use personal credentials via CLI or environment variables, those will be ignored and the refresh token will be used instead (which is still referring to `testflinger-admin`)

Proposed solution
----

Auth module should only be initialized when needed. Also, if authentication is needed, it will need to match the following order:
1. Explicit credentials via CLI args
2. Refresh Token
3. Environment Variables

rene-oromtz:
## Description
This PR solves an issue that can happen when multiple context are used e.g production, staging, dev or multiple users. For context, please refer to linked github issue. 

Auth module is now removed from CLI constructor and initialized only whenever required by an authenticated command. 

Authentication is now made with the following priority:
1. Explicit cli arguments (which updates the refresh token as well)
2. Environment file `.env` 
3. Refresh Token (persistent login to avoid passing credentials constantly via CLI or store it in insecure env file)

Also added Base Class for all auth related issues in the `errors` module

## Resolved issues
Resolves #807
Resolves [CERTTF-716](https://warthogs.atlassian.net/browse/CERTTF-716)

## Tests
Current tests are passing, only few tests needed modification because auth is no longer made during constructor and also to match current authentication priority.

### Local testing
1. Create two users with different roles by running:
`docker exec -it testflinger-server client_credentials_admin`

2. Login as user with admin role:
```
uv run testflinger-cli --server http://localhost:5000 login --client-id 'test_admin' --secret-key 'password'
Successfully authenticated as user 'test_admin'
```

3. Run admin only command, at this point, refresh token should be available so no need to specify credentials:
```
uv run testflinger-cli --server http://localhost:5000 admin get client-permissions
[{"client_id": "test_admin", "max_priority": {}, "allowed_queues": [], "max_reservation_time": {}, "role": "admin"}, {"client_id": "test_user", "max_priority": {}, "allowed_queues": [], "max_reservation_time": {}, "role": "contributor"}]
```

4. Login as regular user. explicit credentials will be prioritized and refresh token ignored (and cleared):
```
uv run testflinger-cli --server http://localhost:5000 login --client-id 'test_user' --secret-key 'password'
Successfully authenticated as user 'test_user'
```

5. Attempt admin command with this user:
```
uv run testflinger-cli --server http://localhost:5000 admin get client-permissions
Authorization Error: Command requires role: admin, manager
```

6. Validate client_id is now tied to refresh token:
Note: When using  `uv`, config_home is set to `~/.config` instead of $SNAP_USER_DATA 
```
cat ~/.config/testflinger-cli-auth.conf 
[AUTH]
client_id = test_user
refresh_token = m1z3MicPZ2YpghtHq5tL4g6dDmWFdn9xJFeyHVGNi5HKPZjO
```

[CERTTF-716]: https://warthogs.atlassian.net/browse/CERTTF-716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -76,6 +76,8 @@ def cli():
         tfcli.run()
     except KeyboardInterrupt:
         sys.exit("Received KeyboardInterrupt")
+    except CredentialsError as exc:
+        sys.exit(exc)
 
 
 class TestflingerCli:
--- a/cli/testflinger_cli/auth.py
+++ b/cli/testflinger_cli/auth.py
@@ -37,12 +37,7 @@ from testflinger_cli.errors import (
 class TestflingerCliAuth:
     """Class to handle authentication and authorisation for Testflinger CLI."""
 
-    def __init__(
-        self,
-        client_id: str,
-        secret_key: str,
-        tf_client,
-    ):
+    def __init__(self, client_id: str, secret_key: str, tf_client):
         self.client_id = client_id
         self.secret_key = secret_key
         self.client = tf_client
@@ -87,6 +82,7 @@ class TestflingerCliAuth:
             response = self.client.authenticate(
                 self.client_id, self.secret_key
             )
+            # Store refresh token for persistent login
             self.store_refresh_token(response["refresh_token"])
             return response["access_token"]
         except client.HTTPError as exc:


boukeas (approved):
Thanks for the fix Rene and for addressing the comments. I am approving since this looks exactly as I would expect it but, just to be on the safe side, before we merge, please test this on a local deployment and describe the test in the PR description.

PR #809, opened 2025-09-25
feat(server): introduce secrets store based on CSFLE in MongoDB

boukeas:
## Resolved issues

This PR doesn't resolve an issue itself, but it will allow testing for CERTTF-443 from the Testflinger staging environment (instead of just locally).

### DIFF
--- a/server/src/testflinger/secrets/__init__.py
+++ b/server/src/testflinger/secrets/__init__.py
@@ -15,24 +15,48 @@
 
 """Default setup for the Testflinger secrets store."""
 
+import base64
+import logging
 import os
+from contextlib import suppress
 
 import requests
+from bson.codec_options import CodecOptions
 from hvac import Client
+from pymongo import MongoClient
+from pymongo.encryption import ClientEncryption
+from pymongo.errors import EncryptionError
 
+from testflinger.database import get_mongo_uri
 from testflinger.secrets.exceptions import StoreError
+from testflinger.secrets.mongo import MongoStore
 from testflinger.secrets.store import SecretsStore
 from testflinger.secrets.vault import VaultStore
 
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+handler = logging.StreamHandler()
+formatter = logging.Formatter(
+    "[%(asctime)s] [%(levelname)s] %(name)s: %(message)s",
+    datefmt="%Y-%m-%dT%H:%M:%S %z",
+)
+handler.setFormatter(formatter)
+logger.addHandler(handler)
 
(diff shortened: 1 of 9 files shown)


server/src/testflinger/secrets/__init__.py
```diff
@@ -44,6 +68,101 @@ def setup_secrets_store() -> SecretsStore | None:
...
+    master_key_b64 = os.environ.get("TESTFLINGER_SECRETS_MASTER_KEY")
+    if not master_key_b64:
+        return None
+
+    try:
+        master_key = base64.b64decode(master_key_b64)
+    except Exception as error:
+        raise StoreError(
+            "Environment variables with Mongo credentials are incorrect"
+        ) from error
```
rene-oromtz, line 104: I guess this is the cleaner way to avoid catching all possible exceptions (binascii.Error, ValueError, TypeError). Can you please add a comment just mentioning this exceptions for clarity?
```suggestion
    except Exception as error:
        # binascii.Error, ValueError and TypeError can be raised during master_key decoding
        raise StoreError(
            "Environment variables with Mongo credentials are incorrect"
        ) from error
```

server/src/testflinger/secrets/mongo.py
```diff
@@ -0,0 +1,119 @@
...
+        self,
+        client: MongoClient,
+        cipher: ClientEncryption,
+        data_key_name: str,
+    ):
+        """Initialize MongoStore with client, cipher, and data key name."""
+        self.database = client.get_default_database()
+        self.cipher = cipher
+        self.data_key_name = data_key_name
+        self.algorithm = Algorithm.AEAD_AES_256_CBC_HMAC_SHA_512_Random
```
rene-oromtz, line 45: Any particular reason why to set this as an attribute instead of as a constant or directly on the write method?

server/src/testflinger/secrets/mongo.py
```diff
@@ -0,0 +1,119 @@
...
+            raise StoreError(
+                f"Failed to decrypt value for '{key}' under '{namespace}'"
+            ) from error
+
+    def write(self, namespace: str, key: str, value: str):
+        """Write the `value` for `key` under `namespace`."""
+        try:
+            encrypted_value = self.cipher.encrypt(
+                value=value.encode("utf-8"),
+                algorithm=self.algorithm,
```
rene-oromtz, line 85: If algorithm is not used by anything else, could this make more sense? Or as a constant if you prefer, what do you think?
```suggestion
                algorithm=Algorithm.AEAD_AES_256_CBC_HMAC_SHA_512_Random,
```

rene-oromtz (approved):

PR #810, opened 2025-09-29
ci: update workflows to new zizmor recommendations

pedro-avalos:
## Description

- This PR resolves the issues brought up by the `zizmor` update.
- The following changes were made:
  - Added notes to `permissions` changes with explanation for the elevation
  - Updated workflows that were pinned to a major version, pinned to minor version now, to avoid issues with `zizmor` complaining the SHA was pointing to a minor version
  - Updated GH action versions

## Tests

- Tested with `zizmor` locally
- `zizmor` workflow should pass
- Existing workflows that do not push/release should automatically run with this PR to verify that our CI is not broken by any changes.


rene-oromtz (approved):
LGTM!

PR #816, opened 2025-10-01
feat(cli/server): improve creation and update of client permissions

rene-oromtz:
## Description
This PR is meant to make more user friendly the creation/update of client permissions. Currently, the server differentiates between `POST` and `PUT` request for creation and update respectively and also requires some set of required parameters in order to perform the necessary action. 

This PR proposed to simplify and just use `PUT` for both operations. Parameters are also more relaxed and if not provided, it will be defaulted on server side during client_id creation. 

This modification simplifies the CLI component and allows to specify single argument for update as other values will be defaulted or retrieved from existing id on server side. 

### Changes
CLI:
1. client POST removed to match server
2. `set` for creation will now only require `client_secret` and `client_id` if provided via args or json.
3. `set` for update can also take single argument to be modified. Update now works by sending just the required modification on request body. Server will get the current permissions and just update the provided in request.

Server:

1. Remove POST endpoint, PUT will now create and update client permissions. 
2. Secret can now also be included in the `PUT` request during client permissions update. Password is hashed and stored in the db. 
3. Add PUT method restriction to `testflinger-admin`. This client-id is not meant to be modified from the API, permissions are granted to personal client-id. Since secret can also be modified from the API now, this restriction allows only source for modifying the admin account is via Charm.    

## Resolved issues
Resolves [CERTTF-725](https://warthogs.atlassian.net/browse/CERTTF-725)

## Documentation
Minor modification to the docs to be more clear on update

## Tests
Needed to add a couple more unit tests to improve coverage
### Server Setup
```
$ docker compose down
$ docker compose up -d --build
# populate the database with the data of a single client 
$ docker exec -it testflinger-server client_credentials_admin
{'client_id': 'test_user', 'role': 'admin', 'max_priority': {}, 'allowed_queues': [], 'max_reservation_time': {}}
```

### Test CLI changes
```
# First authenticate with test_user
uv run testflinger-cli --server http://localhost:5000 login
Successfully authenticated as user 'test_user'

# Create new user with only client-id and client-secret
uv run testflinger-cli --server http://localhost:5000 admin set client-permissions --testflinger-client-id 'userA' --testflinger-client-secret 'super-secret-key'
Created permissions for client 'userA'


# Validate user was created
uv run testflinger-cli --server http://localhost:5000 admin get client-permissions --testflinger-client-id 'userA'
{"client_id": "userA"}

# Attempt to edit existing client with set
uv run testflinger-cli --server http://localhost:5000 admin set client-permissions --testflinger-client-id 'userA' --max-reservation '{"q1": 10000}'
Updated permissions for client 'userA'

# Modify role for userA
uv run testflinger-cli --server http://localhost:5000 admin set client-permissions --testflinger-client-id 'userA' --role manager
Updated permissions for client 'userA'

# Validate permissions changed
uv run testflinger-cli --server http://localhost:5000 admin get client-permissions --testflinger-client-id 'userA'
{"client_id": "userA", "max_reservation_time": {"q1": 10000}, "role": "manager"}

# Attempt to create new user without password
uv run testflinger-cli --server http://localhost:5000 admin set client-permissions \
--json '{"client_id": "userB", "max_priority": {}, "max_reservation_time": {}, "role": "contributor"}'
ERROR: 2025-10-06 11:37:46 admin.py:309 -- Error: Missing client_secret in request body for new client

# Create userB providing the password
uv run testflinger-cli --server http://localhost:5000 admin set client-permissions \
--json '{"client_id": "userB", "client_secret": "tmp-password", "max_priority": {}, "max_reservation_time": {}, "role": "contributor"}'
uv run testflinger-cli --server http://localhost:5000 admin get client-permissions --testflinger-client-id 'userB'

# Change the password for this userB
uv run testflinger-cli --server http://localhost:5000 admin set client-permissions \
--testflinger-client-id 'userB' --testflinger-client-secret 'insecure-password'
Updated permissions for client 'userB'

# Validate other permissions were not modified
uv run testflinger-cli --server http://localhost:5000 admin get client-permissions --testflinger-client-id 'userB'
{"client_id": "userB", "max_priority": {}, "max_reservation_time": {}, "role": "contributor"}

# Attempt to login with another-new-user with initial password
uv run testflinger-cli --server http://localhost:5000 login --client-id 'userB' --secret-key 'tmp-password'
Authentication with Testflinger server failed. Check your client id and secret key

# Attempt to login with modified password
uv run testflinger-cli --server http://localhost:5000 login --client-id 'userB' --secret-key 'insecure-password'
Successfully authenticated as user 'userB'

```

[CERTTF-725]: https://warthogs.atlassian.net/browse/CERTTF-725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/testflinger_cli/admin.py
+++ b/cli/testflinger_cli/admin.py
@@ -290,11 +290,8 @@ class TestflingerAdminCLI:
             sys.exit("Error: client_id cannot be empty")
 
         try:
-            # Response include if we are creating or updating
-            print(
-                self.main_cli.client.set_client_permissions(
-                    auth_header, json_data
-                )
+            response = self.main_cli.client.set_client_permissions(
+                auth_header, json_data
             )
         except client.HTTPError as exc:
             if exc.status in [
@@ -314,6 +311,9 @@ class TestflingerAdminCLI:
                     exc.msg,
                 )
             return
+
+        # Response indicates if we are creating or updating
+        print(response)
 
     @require_role(ServerRoles.ADMIN, ServerRoles.MANAGER)
     def get_client_permissions(self):


boukeas (changes requested):
Rene, I think you are "forced" to separate the logic because of the way the corresponding server API currently works. I believe this needs to be challenged: my suggestion is to have a single `client-permissions/<client-id>` PUT endpoint for both creating and updating permissions. And also handle all unspecified permissions on the server side with reasonable defaults. This would also simplify things substantially on the CLI end as well.

server/src/testflinger/api/v1.py
```diff
@@ -1094,97 +1099,78 @@ def get_client_permissions(client_id) -> list[dict]:
...
-    json_data.pop("client_id", None)
-    json_data.pop("client_secret", None)
-
-    # Retrieve non null values
-    update_fields = {
-        key: value for key, value in json_data.items() if value is not None
-    }
+    # Define permissions to set for client_id
+    client_permissions = current_permissions.copy()
+    client_permissions.update(json_data)
```
boukeas, line 1161: Why do we need to copy `current_permissions` into `client_permissions`? Why can't we just update `current_permissions` and use that?

boukeas: Also, `current_permissions` might be identical to `default_permissions` (if the client id doesn't exist in the database. Unless I am missing something, this means that all the defaults will be written in the database. Is that correct? Do we need an entry for e.g. `allowed_queues` in the database if that field isn't included in the request?

cli/testflinger_cli/admin.py
```diff
@@ -274,45 +273,58 @@ def set_client_permissions(self):
...
-                auth_header, tf_client_id
-            )
-            # Client exists, so edit it
-            print(f"Updating existing client '{tf_client_id}'...")
-            self.main_cli.client.edit_client_permissions(
-                auth_header, json_data
+            client_exist = bool(
+                self.main_cli.client.get_client_permissions(
+                    auth_header, tf_client_id
+                )
```
boukeas, line 297: Personally I don't see why we need to perform this additional request, just to display a message. Besides, if you really want to know if the PUT request resulted in a Create or an Update you could return that message in the response to the PUT request (instead of a simple OK).

boukeas (changes requested):
Thanks Rene. I've left a few comments here and there that may need changing or clarifying.

boukeas (approved):

PR #817, opened 2025-10-03
refactor(server): Specify an ordering between roles

boukeas:
## Description

This PR removes the `auth.check_role_hierarchy` function that would compare two instances of `ServerRoles` and refactors its functionality into a "less-than" operator in the `ServerRoles` class. This allows code that previously used `auth.check_role_hierarchy` to become more readable and explicit, e.g.:

```
    if not auth.check_role_hierarchy(g.role, current_role):
```

now becomes:

```
    if ServerRoles(g.role) < ServerRoles(current_role):
```

## Resolved issues

None, this is a minor untracked modification.

## Tests

Introduced unit tests for the comparison operators, The tests in `test_v1_authorization` for the `client-permissions` endpoints remain unchanged and still pass.

### DIFF
--- a/server/src/testflinger/api/auth.py
+++ b/server/src/testflinger/api/auth.py
@@ -288,28 +288,6 @@ def require_role(*roles):
     return decorator
 
 
-def check_role_hierarchy(user_role: str, target_role: str) -> bool:
-    """
-    Check role of current user to validate if authorized to perform action.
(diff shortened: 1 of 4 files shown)


rene-oromtz (approved):
Thank you for taking a look at this. LGTM

PR #825, opened 2025-10-14
fix: Restore network settings in default-user-data for stock

Artur-at-work:
## Description
Previously, network was disabled to speed up provisioning when ubuntu archive was timing out in lab. This change will restore network in stock installer to follow default settings and fix issue when questing quoka does not re-enable networking in the provisioned system.  

## Tests
1. Run TF agent locally with updated default-user-data
2. Provision Stock Ubuntu 25.10 on DUT and verify ssh connection works.
3. Provision Stock Ubuntu 24.04 on DUT and verify ssh connection works.


KaiChuan-Hsieh (approved):
LGTM

PR #839, opened 2025-12-02
feat(server): display new log format in job view

rene-oromtz:
## Description
The new logging redesign is using new collections `logs` to store the results at each stage. However, the job view was using only the data that was coming from the `jobs` collection. 

This PR reuses the transformation that was already made for converting logs into results via the API. 

New method `format_logs_as_results` is being now used by both API and web view to retrieve/display properly the logs.

Additionally, it add Tabs to the job detail view 

## Resolved issues
[CERTTF-760](https://warthogs.atlassian.net/browse/CERTTF-760)

## Tests
Tested on local environment after manually created two jobs via testflinger-cli

Submit results in old format:
```
curl -X POST http://localhost:5000/v1/result/4ef3b744-ea86-4b70-b11e-46d3ca456cc6 \
    -d '{
      "provision_status": 0,
      "provision_output": "legacy provision output",
      "test_status": 1,
      "test_output": "legacy test output",
      "job_state": "complete"
    }'
OK
```

Validate results are visible in the UI
Screenshot from 2025-12-02 09-20-13

Submit results in the new format:
```
curl -X POST http://localhost:5000/v1/result/4ed1e09a-22f9-4984-8f78-28b4b8df887e \
    -d '{
      "status": {"provision": 0, "test": 1},
      "device_info": {"agent_name": "agent1", "device_ip": "1.1.1.1"},
      "job_state": "complete"
    }'
OK
```

```
curl -X POST http://localhost:5000/v1/result/4ed1e09a-22f9-4984-8f78-28b4b8df887e/log/output \
    -d '{
      "fragment_number": 0,
      "timestamp": "2024-12-02T16:00:00Z",
      "phase": "provision", 
      "log_data": "Provision log content"
    }'
OK
```

```
curl -X POST http://localhost:5000/v1/result/4ed1e09a-22f9-4984-8f78-28b4b8df887e/log/output \
    -d '{
      "fragment_number": 0,
      "timestamp": "2024-12-02T16:01:00Z", 
      "phase": "test",
      "log_data": "Test log content"
    }'    
OK
```

Validate results are also visible in the UI
Screenshot from 2025-12-02 09-20-50

[CERTTF-760]: https://warthogs.atlassian.net/browse/CERTTF-760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/src/testflinger/logs.py
+++ b/server/src/testflinger/logs.py
@@ -141,7 +141,14 @@ class MongoLogHandler(LogHandler):
         )
 
     def format_logs_as_results(self, job_id: str, result_data: dict) -> dict:
-        """Format logs stored in MongoDB into the result_data structure."""
+        """Format logs stored in MongoDB into the result_data structure.
+
+        side-effect: The provided result_data is also modified in-place.
+
+        :param job_id: The job identifier.
+        :param result_data: The original result_data structure.
+        :return: The modified result_data structure.
+        """
         # Reconstruct result format with logs and phase statuses
         result_logs = {
             f"{phase}_{log_type}": log_data
--- a/server/src/testflinger/templates/job_detail.html
+++ b/server/src/testflinger/templates/job_detail.html
@@ -26,10 +26,9 @@
     </tbody>
   </table>
   <h2 class="p-muted-heading" id="job-definition">
-    Job Definition
     <a href="#job-definition"
-       class="p-icon--anchor"
-       aria-label="Link to Job Definition section">#</a>
+       class="p-link--anchor-heading"
+       aria-label="Link to Job Definition section">Job Definition</a>
   </h2>
   <div class="p-tabs">
     <div class="p-tabs__list" role="tablist" aria-label="Job Definition tabs">
@@ -110,10 +109,9 @@
   {% set phases_with_results = phases | selectattr("output") | selectattr("status", "!=", none) | list %}
   {% if phases_with_results %}
     <h2 class="p-muted-heading" id="phase-results">
-      Phase Results
       <a href="#phase-results"
-         class="p-icon--anchor"
-         aria-label="Link to Phase Results section">#</a>
+         class="p-link--anchor-heading"
+         aria-label="Link to Phase Results section">Phase Results</a>
     </h2>
     <div class="p-tabs">
       <div class="p-tabs__list" role="tablist" aria-label="Phase Results tabs">


server/src/testflinger/logs.py
```diff
@@ -139,3 +139,27 @@ def retrieve_log_fragments(
...
+                log_data := self.retrieve_logs(job_id, log_type, phase)[
+                    "log_data"
+                ]
+            )
+        }
+        phase_status = result_data.pop("status", {})
+        result_status = {
+            f"{phase}_status": status
+            for phase in TestPhase
+            if (status := phase_status.get(phase)) is not None
```
ajzobro: the "is not None" was added -- this will permit empty strings, is this intended?

rene-oromtz: Yes, that's intended, the reason is that the number 0 gets evaluated to False so its not added as `"{phase}_status": 0`

server/pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "uv_build"
 name = "testflinger"
 description = "Testflinger Server"
 readme = "README.md"
-version = "1.4.1"
+version = "1.4.3"
```
rene-oromtz: Please ignore this bump from x.x.1 to x.x.3. This is temporary as staging already has 1.4.2 and for some reason seems it won't pickup HTML changes if this is not bumped. 

I will revert to 1.4.2 once I do a final test in staging

rene-oromtz:
[Screencast from 2025-12-02 14-58-16.webm](https://github.com/user-attachments/assets/4a86f899-6cb6-4e6c-a913-30e8d2e00633)

@ajzobro @pedro-avalos 
Idded the anchor link but I also removed any reference to the live output for now... Turns out there is an additional modification that is needed on agent side as the status is only being sent at the end of the job. Because of that, even when adding the running icon works, since there is no status code I can't mark the status icon live.

I'm using the original (current) behavior to only show the Phase Results section AFTER the job completes. 

I'll then create a different Jira card so we can take care of live output more properly and in a dedicated PR

server/src/testflinger/templates/job_detail.html
```diff
@@ -25,59 +25,133 @@ <h1 class="p-heading--3">Job Detail - {{ job.job_id }}</h1>
...
-  <h2 class="p-muted-heading">Job Definition</h2>
-  <div class="p-code-snippet">
-    <div class="p-code-snippet__header">
-      <h5 class="p-code-snippet__title">Provision Data</h5>
+  <h2 class="p-muted-heading" id="job-definition">
+    Job Definition
+    <a href="#job-definition"
+       class="p-icon--anchor"
+       aria-label="Link to Job Definition section">#</a>
+  </h2>
```
pedro-avalos: The class should not be `p-icon--anchor`, but `p-link--anchor-heading`. Additionally, the `<a>` element should contain the text. This will give us the Vanilla anchor heading style (normal text, but adds a link icon on hover)


```suggestion
  <h2 class="p-muted-heading" id="job-definition">
    <a href="#job-definition"
       class="p-link--anchor-heading"
       aria-label="Link to Job Definition section">Job Definition</a>
  </h2>
```

server/src/testflinger/logs.py
```diff
@@ -139,3 +139,27 @@ def retrieve_log_fragments(
             )
             for fragment in log_collection.find(query).sort("fragment_number")
         )
+
+    def format_logs_as_results(self, job_id: str, result_data: dict) -> dict:
+        """Format logs stored in MongoDB into the result_data structure."""
```
pedro-avalos: Since this function doesn't just return the dictionary but modify the `result_data` argument, could you document the side effect on the docstring? Maybe something like:


```suggestion
        """Format logs stored in MongoDB into the result_data structure.

        :param job_id: The job identifier.
        :param result_data: The original result_data structure. This will be modified.
        :return: The modified result_data structure.
        """
```

pedro-avalos (changes requested):
Thanks for updating the UI. Here are some comments to make the UI more consistent with the Vanilla Framework and our other sites/applications.

I also added a suggestion to document the new helper function (it gets at the comment that @ajzobro made earlier)

PR #843, opened 2025-12-09
feat(agent): expose additional metrics for testflinger agent monitoring

rene-oromtz:
## Description
This PR add couple of additional metrics to be exposed by the agent for later visualization in Grafana

1. Add a Counter with Total Recovery Failures. This is of relevance as this helps to monitor and identify which agents are failing to recover at any stage as this failure leads to an agent to be set to offline automatically. 
2. Add a Histogram metric so we can track the time the agent spends at each phase.  
3. Also add a label (`agent_id`) to each of the metrics so it can be queried easier 

## Resolved issues
[CERTTF-771](https://warthogs.atlassian.net/browse/CERTTF-771)

## Tests
Modified the existing unit test to match the newly added label and the duration of each phase
Also added 1 additional unit test for the recovery failures counter
Used the exposed metrics in staging to build a dashboard (WIP). For now, dashboard is NOT included as template here, as we want to monitor all agents in all datacenter and adding template here means having 1 dashboard per charm
Dashboard is divided in two sections, one view per datacenter and one view per agent, both can be selected from the variables
Screenshot from 2025-12-09 17-42-49

[CERTTF-771]: https://warthogs.atlassian.net/browse/CERTTF-771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/tests/test_agent.py
+++ b/agent/tests/test_agent.py
@@ -862,10 +862,11 @@ class TestClient:
 
     def test_agent_metrics(self, agent, requests_mock):
         """
-        Tests that total job and total job failures metrics are increased
-        when running a job.
+        Tests that total job, total job failures, and phase duration metrics
+        are tracked when running a job.
         """
         self.config["provision_command"] = "/bin/false"
+        agent_id = self.config["agent_id"]
         mock_job_data = {
             "job_id": str(uuid.uuid1()),
             "job_queue": "test",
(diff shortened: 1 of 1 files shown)


pedro-avalos (approved):
Looks great! I have just left one comment to clarify the choice of buckets, but that is not a blocker or request for changes necessarily.

PR #845, opened 2025-12-10
update readme for cli to demonstrate typical use of --poll argument

ajzobro:
## Description

Minor change to the README.md to provide an example of using --poll right from the start as is likely the most often uses use-case.

## Resolved issues

Added an example, see above.

## Documentation

Documentation change only.

## Web service API changes

Documentation change only.

## Tests

N/A

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -32,8 +32,7 @@ testflinger-cli poll d14d2e31-f239-41b6-9a74-32f538e71cde
 ```
 
 > [!TIP] 
-> If you are expecting to poll the job, you may wish to specify this from
-the beginning when submitting the job with the `--poll` flag:
+> If you are expecting to poll the job, you may wish to specify this from the beginning when submitting the job with the `--poll` flag:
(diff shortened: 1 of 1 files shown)


rene-oromtz (changes requested):
Thanks for helping to improve the CLI README! I added a comment for rendering properly in markdown and a suggestion on this note placement

cli/README.md
```diff
@@ -31,6 +31,14 @@ You can also watch the output of the job as it runs by polling it with:
 testflinger-cli poll d14d2e31-f239-41b6-9a74-32f538e71cde
 ```
 
+> [!TIP] 
+> If you are expecting to poll the job, you may wish to specify this from
+the beginning when submitting the job with the `--poll` flag:
```
ajzobro:
```suggestion
> If you are expecting to poll the job, you may wish to specify this from the beginning when submitting the job with the `--poll` flag:
```

PR #846, opened 2025-12-11
feat(agent): send phase status code upon phase completion

rene-oromtz:
## Description
This is a follow-up on the logging redesign, since logs are now NOT transmitted at the end of the job execution, we can also send the phase status upon phase completion. 

1. This PR removes storing output and serial output in the `testflinger-outcome.json`, in the logging redesign, they were removed anyway while submitting results (pop'ed). While not important as of now, but this reduces the size of this file as its not storing any log output at all.
2. Send status updates (exit code) once the phase is completed. With this server keeps track of each phase completion immediately showing the results for that phase even when the job is not completed. 
3. Other minor modification includes replacing some uses of os.path to Path instead. 

This is also a step forward to be able to show live output in the UI

## Resolved issues
Resolves [CERTTF-770](https://warthogs.atlassian.net/browse/CERTTF-770)

## Tests
All current unit tests were passing, some needed minor modifications related to the usage of Pathlib. 
Phase related tests were modified to match this new behavior and added couple more. 

Tested as well on staging. 

Job at `test` phase. Provision logs are already available in the UI
Screenshot from 2025-12-10 17-53-29

Job at `reserve` phase. Provision and test logs are already available in the UI
Screenshot from 2025-12-10 17-54-55

[CERTTF-770]: https://warthogs.atlassian.net/browse/CERTTF-770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/src/testflinger_agent/job.py
+++ b/agent/src/testflinger_agent/job.py
@@ -192,7 +192,7 @@ class TestflingerJob:
             # Write serial log file generated in device connector to
             # the serial log endpoint if the file exists
             self.serial_output_handler.write_from_file(serial_log)
-            self._send_phase_results(
+            self._update_phase_results(
                 results_file,
                 phase,
                 exitcode,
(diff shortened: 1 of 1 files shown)


ajzobro (commented):
Makes sense.

PR #848, opened 2025-12-12
Feature/certtf 756

ajzobro:
## Description

Converted the list of queues displayed on the agent detail page into a table (matching the format of the queue_detail page exactly).

image

## Resolved issues

Resolves [CERTF-756](https://warthogs.atlassian.net/browse/CERTTF-756).

## Web service API changes

N/A

## Tests

Updated `server/devel/create_sample_data.py` to allow for a random number of queues to be attributed to each agent for better testing.

### DIFF
--- a/server/src/testflinger/templates/agent_detail.html
+++ b/server/src/testflinger/templates/agent_detail.html
@@ -55,64 +55,64 @@
           </tr>
         {% endfor %}
       </tbody>
-  </ul>
-  <h2 class="p-muted-heading">Provision History</h2>
-  <form action="" method="get" class="p-form--inline">
-    <div class="p-form__group">
-      <label for="start-date" class="p-form__label">Start Date</label>
-      <input type="date"
-             id="start-date"
-             name="start"
-             class="p-form__control"
-             value="{{ request.args.get('start', agent.start) }}">
+    </ul>
+    <h2 class="p-muted-heading">Provision History</h2>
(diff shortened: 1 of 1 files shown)


server/src/testflinger/views.py
```diff
@@ -91,6 +91,23 @@ def agent_detail(agent_id):
         for queue in agent_info.get("queues", [])
     }
 
+    queue_info = []
+    for queue_name in agent_info.pop("queues", []):
+        queue_data = mongo.db.queues.find_one({"name": queue_name})
```
rene-oromtz: Just note that the `queues` collection only reports "advertised queues" which does not accurately reflect the queues available. 

You can use `testflinger-cli list-queues` to take a look at what it is exposed in that collection in production...

So for this I would suggest to get the name of the queue directly from the `agent_info`

rene-oromtz (changes requested):
Thanks for taking a look at this! 
I just added couple comments to match the current practices we are following and an important note regarding the `queues` collection. 

Also, there is one test failing because the HTML lint. Can you run `djlint --reformat src`? That will do the formatting of this code automatically.

pedro-avalos:
One general comment @ajzobro, since you should have access; it would make it easier if you create branches on this repo itself rather than fork

ajzobro:
Moved to https://github.com/canonical/testflinger/pull/853

PR #850, opened 2025-12-16
[CERTTF-697]: Feature: adds exclude_agents field to job data to allow jobs to avoid specific agents

ajzobro:
## Description

Resolves [CERTTF-697]: https://warthogs.atlassian.net/browse/CERTTF-697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Added the ability to specify that select agents are not to be used to run a given job, even if they are in the specified queue.

Agents will now be required to POST to `/agents/data/<agent_name>` before they are able to GET a `/job` to work. The response message from registering the agent will return a cookie which will identify the agent on subsequent calls to fetch a job to work on.

## Resolved issues

Resolves Jira [CERTTF-697](https://warthogs.atlassian.net/browse/CERTTF-697)

## Documentation

Examples and references data updated in `docs/`. 

### Web service API changes

Job data may now include `exclude_agents` at the top level, e.g. alongside `job_queue`, for example:
```
job_queue: my_queue
exclude_agents:
    - agent_007
    - agent_002
provision_data:
  distro: noble
reserve_data:
  ssh_keys:
    - lp:bond
  timeout: 3600
```

Added a cookie in the response to POST `/agents/data/<agent_name>` which is required to then GET a `/job`. This is done to ensure that agents getting work are who they say they are and without introducing any new agent code to do so. The `requests` library is already configured to respond with cookies it it given and the agents are already configured to register before they look for work.

`exclude_agents` was added to the database job schema in support of this.

When there are excluded agents, they will be considered when checking against the relevant queues upon job scheduling and assignment so that jobs with no agents will be prevented from being scheduled and agents will not be blocked from doing other work for which they are not excluded from running.

### Agents

Due to the fact that the current agent code was not expected to respond to a `401 Unauthorized` the agent code was modified such that it will reregister in such an event. Agents will be restarted in advance of the server update so that they will be ready to reregister in such an event.

### Command Line

The `testflinger` CLI was updated to support checking the validity of the new field when attempting to schedule a job.

## Tests

For each function that changed, unit tests were added to ensure that unintended changes were not 
introduced and that the new behavior was as expected.

Further testing was done on the staging server to test the full interaction.

##### CLI
`$ cd cli && uvx tox`

Example job
```
job_queue: staging
exclude_agents:
  - multi
  - doesnt_exist
  - petilil___
  - audino
provision_data:
  distro: noble
reserve_data:
  ssh_keys:
    - lp:ajzobro
```

Tested against a queue where all agents on that queue were excluded; and then again where there was only one agent allowed.

##### Agent
`$ cd agent && uvx tox`

Added multiple agents on the staging server to a queue and verified that they were able to assign work and provision a machine.

##### Server
`$ cd server && uvx tox`

Initial tests performed via docker:

```shell
# clean up any existing environment
docker compose down

# build and start the server
docker compose build

### CODE REVIEW

File: cli/tests/test_secret.py [hunk 1]
```diff
@@ -94,7 +94,7 @@ def test_secret_write_http_error(auth_fixture, requests_mock):
 @pytest.mark.parametrize(
     "test_path",
     [
-        "path"
+        "path",
```
ajzobro: This appeared to be a mistake and was getting reformatted by `uvx ruff format`

server/tests/test_v1.py
```diff
@@ -741,8 +770,9 @@ def test_agents_post(mongo_app):
     }
     output = app.post(f"/v1/agents/data/{agent_name}", json=agent_data)
 
+    print(vars(output))
```
ajzobro: Remove the debug print.
```suggestion
```

docs/.custom_wordlist.txt
ajzobro: #852 has many of the same changes, but also changes webdav to WebDAV (appropriately)

cli/tests/test_cli.py
```diff
@@ -99,6 +99,63 @@ def test_submit(capsys, tmp_path, requests_mock):
...
+    requests_mock.get(
+        URL + "/v1/queues/fake/agents",
+        json=[
+            {"name": "fake_agent", "state": "waiting"},
+            {"name": "fake_agent_2", "state": "waiting"},
+        ],
+    )
+    sys.argv = ["", "submit", str(testfile)]
+    tfcli = testflinger_cli.TestflingerCli()
+    tfcli.submit()
```
rene-oromtz: We could also be using run as we are already using the submit argument, personally I preferred to use this but I don't mind if you decide to keep it as-is so I'll let you decide
```suggestion
    tfcli.run()
```

rene-oromtz (changes requested):
Thanks for taking care of this, I added couple comments but overall I like the proposed approach. 
Also, could you please test this in staging just to make sure everything is working as intended? Once tested could you please include the tests made in the PR description?

server/src/testflinger/api/v1.py
```diff
@@ -88,6 +88,24 @@ def job_post(json_data: dict):
     if not job_queue:
         abort(422, message="Invalid data or no job_queue specified")
 
+    exclude_agents = json_data.get("exclude_agents", [])
+    if not isinstance(exclude_agents, list):
+        abort(422, message="exclude_agents must be a list if specified")
```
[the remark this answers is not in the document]
ajzobro: Why isn't the code on line 88 also handled in the schema validation then?

server/src/testflinger/api/v1.py
```diff
@@ -178,6 +196,10 @@ def job_builder(data: dict):
     if has_attachments(data):
         data["attachments_status"] = "waiting"
 
+    # Ensure exclude_agents is always a list
+    if "exclude_agents" not in data:
+        data["exclude_agents"] = []
```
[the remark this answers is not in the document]
pedro-avalos: If you're going to use `load_default`, that should likely be the callable `list`, not `[]` because that can lead to some wonky behavior (technically you are using a global list that *can* be modified when using the literal `[]` in some cases and I think this is one of them)

`exclude_agents = fields.List(fields.String(), required=False, load_default=list)`

server/src/testflinger/api/v1.py
```diff
@@ -88,6 +88,24 @@ def job_post(json_data: dict):
...
+        # Make sure that there are at least some agents in the selected queue
+        # which can run this job.
+        agents_can_run = [
+            agent
+            for agent in database.get_agents_on_queue(job_queue)
+            if agent["name"] not in exclude_agents
+        ]
+        if not agents_can_run:
+            abort(
+                HTTPStatus.UNPROCESSABLE_ENTITY,
```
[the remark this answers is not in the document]
rene-oromtz: Yes, I understand that point. What I meant was that CLI can _still_ encounter this 422 error and I was assuming it wasn't catched but just found it should be covered by the else in cli/testflinger_cli/__init__.py lines 878-883:
```
                else:
                    # This shouldn't happen, so let's get more information
                    sys.exit(
                        "Unexpected error status from testflinger "
                        f"server: [{exc.status}] {exc.msg}"
                    )
```

server/src/testflinger/api/v1.py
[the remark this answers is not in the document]
ajzobro: Yes, I suppose `list` would have worked instead of `lambda: []` but the point you make is accurate, we don't want to use a mutable list that will be passed by reference that would would change over time.

I usually use `.get()` with dictionaries, but as has been suggested here we're relying on the decorators to already handle ensuring proper data typing and defaults, so I have reverted to using `[key]` notation instead.

server/src/testflinger/api/v1.py
```diff
@@ -198,7 +206,10 @@ def job_get():
     queue_list = request.args.getlist("queue")
     if not queue_list:
         return "No queue(s) specified in request", HTTPStatus.BAD_REQUEST
-    job = database.pop_job(queue_list=queue_list)
+    agent_name = request.cookies.get("agent_name")
+    if not agent_name:
+        return "Agent not identified", HTTPStatus.UNAUTHORIZED
```
rene-oromtz: ah I missed this one, for consistency with the other API endpoints could you please use `flask.abort()` instead? (if you want to make this same modification to the line 208 that would be great too!)

Also, I think the agent will need to do `job_request.raise_for_status()` in agent/src/testflinger_agent/client.py

ajzobro: OK wrt/`flask.abort()`

However, so far, I have not modified the agent code at all for this as I believe requests will automatically pass along the cookies that it received.  I have not tested this on staging, but we should (show me how).

Why do you think that the agent needs to be modified? Ought it not already do the right things?

rene-oromtz: Agents registers once upon restart or code update. In order for agents to get the cookie you will need to update first the server but if during the update, an agent attempts to get a job it will break as agent does not handle the HTTPError.

I think we need to add this so we can:
1. Update agent code to include this response (no cookie so far)
2. Update server. At this point, the agent can handle the HTTPError and not break, it will get the AuthorizationError but can retry after
3. Restart agents so they can register with the cookie

We can discuss this further in a call and I can show you how to test the full flow in staging

ajzobro: Actually, I looked over this and I believe that all the calls to get a job will not crash out of the loop and that there are many, many calls to the `post_agen_data` which will in turn declare what agent is connecting, such that when we get to requesting a job there is a cookie.

Every post_agent_data() call (including heartbeats) makes a POST to /v1/agents/data/{agent_id}, and the server should respond with Set-Cookie: agent_name=... in the response headers. The session automatically stores and re-uses that cookie.

agent/src/testflinger_agent/agent.py
```diff
@@ -253,6 +254,25 @@ def unpack_attachments(self, job_data: dict, cwd: Path):
...
                     del job_data[phase_str]
 
+    def get_job_data(self):
+        job_data = None
+        try:
+            job_data = self.client.check_jobs()
+        except HTTPError as exc:
+            # If we receive an error that we were not authenticated, try
+            # registering again and then try to get a job.
+            if exc.response.status_code == 401:
```
rene-oromtz: Let's use HTTPStatus instead
```suggestion
            if exc.response.status_code == HTTPStatus.UNAUTHORIZED:
```

agent/src/testflinger_agent/agent.py
```diff
@@ -253,6 +254,25 @@ def unpack_attachments(self, job_data: dict, cwd: Path):
...
+                self.client.post_agent_data({"job_id": ""})
+        if job_data is None:
+            # Try to get a job again; if we fail to get a job a second time,
+            # log the exception and skip the while loop, returning to the
+            # outer-most loop
+            try:
+                job_data = self.client.check_jobs()
+            except Exception as exc:  # pylint: disable=broad-except
+                logger.exception(exc)
+        return job_data
```
rene-oromtz: For this block, I think a simpler approach would be to just to return `job_data` (either valid or `None`) instead of retrying to get a job immediately. Agent will then wait the 10 seconds before automatically attempt to retry getting a job. 
This wait can be particularly useful in case during server upgrade, agent require a bit more of time to get the cookie from the server. (will be also removing this broad except)

If you prefer to attempt to get the a job immediately after registering the agent, I would move this inside the first try-except block as currently this is checking for jobs twice in a row if we don't encounter 401.

agent/src/testflinger_agent/agent.py
```diff
@@ -290,7 +310,12 @@ def process_jobs(self):
         if self.status_handler.needs_restart:
             self.restart_agent(self.status_handler.comment)
 
-        job_data = self.client.check_jobs()
+        rundir = None
+        job = None
+        event_emitter = None
```
rene-oromtz: I think this needs to be inside the while loop. In line 484, we immediately are checking for any queued job so the while loop will continue running and this variables won't be cleared for any subsequent jobs.

agent/tests/test_agent.py
```diff
@@ -985,3 +1000,82 @@ def test_server_wait_for_connectivity(self, agent, requests_mock, caplog):
...
+        mock_post_agent_data.assert_called()
+        # and verify that after we retried we got the expected data.
+        assert result == fake_job
+
+    def test_get_job_data_non_http_error_on_retry(self, agent):
+        """Test get_job_data handles non-HTTPError on retry."""
+        mock_response = Mock()
+        mock_response.status_code = 401
+        error_401 = HTTPError(response=mock_response)
+        generic_error = RuntimeError("Network connection failed")
```
rene-oromtz: Same nitpick comment in case you prefer to add it directly to the side_effect

cli/tests/test_cli.py
```diff
@@ -98,6 +98,63 @@ def test_submit(capsys, tmp_path, requests_mock):
...
+        "job_queue": "fake",
+        "provision_data": {"distro": "fake"},
+        "exclude_agents": "fake_agent",
+    }
+    testfile = tmp_path / "test.json"
+    testfile.write_text(json.dumps(fake_data))
+    fake_return = {"job_id": jobid}
+    requests_mock.post(URL + "/v1/job", json=fake_return)
+    requests_mock.get(
+        URL + "/v1/queues/fake/agents",
```
rene-oromtz: Nitpick, same comment about f-strings

rene-oromtz (changes requested):
Thanks for the changes in agent side! 
I added some comments and some other nitpicks in case you could consider them. 

I may have missed something but for tests, would it be possible to use `HTTPStatus` instead of error code and f-strings while building the URL?

server/src/testflinger/api/schemas.py
```diff
@@ -363,6 +363,9 @@ class Job(Schema):
     reserve_data = fields.Nested(ReserveData, required=False)
     job_status_webhook = fields.String(required=False)
     job_priority = fields.Integer(required=False)
+    exclude_agents = fields.List(
+        fields.String(), required=False, load_default=lambda: []
```
pedro-avalos: I think this should be `load_default=list` since `load_default` accepts callables and it's more readable than a lambda.

rene-oromtz (commented):
Thanks for addressing the changes! This looks good to me, there is just a formatting error that needs to be taken care of before merging. 

Also, would it be possible to test this on staging just for a sanity check, this indeed turned out bigger than expected! (agent + cli + server)

agent/src/testflinger_agent/client.py
```diff
@@ -124,14 +124,16 @@ def check_jobs(self):
...
         except requests.exceptions.RequestException as exc:
             logger.error(exc)
             # Wait a little extra before trying again
             time.sleep(60)
+        else:
+            job_request.raise_for_status()
+            if job_request.content:
+                return job_request.json()
+            else:
+                return None
```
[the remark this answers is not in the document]
ajzobro: done

rene-oromtz (dismissed):

rene-oromtz (approved):

PR #852, opened 2025-12-17
docs(server): add OpenAPI spec generation and validation workflow

tang-mm:
## Description
This pull request introduces the functionality to:
- (**Server**) export the server's OpenAPI specification into a JSON file stored in the repository (This file will be the same as the one in the web application at the `/docs` URL)
  - Created a minimal Flask app entry using `TESTING` mode, so that Flask can be used for spec generation without MongoDB
  - Added `server/scripts/generate_openapi_schema.py` script to call APIFlask's native method `.spec()`  retrieve the spec in Python dictionary form    
  - Added the initial/current version of API spec at `server/openapi.json`
- (**docs**) serve the API reference within Read the Docs documentation, making it publicly accessible 
- (**CI check**) keep the OpenAPI specification file in sync with the server API code
  - Added `schema` tox environment to generate new spec, and `check-schema` that compares the spec file in the working directory against a newly generated version
  - Added CI validation step in `server-tox.yml` workflow

## Resolved issues

Resolves CERTTF-243

## Documentation

- Added `docs/reference/openapi.rst` with Swagger UI integration. The JSON file is copied from the `server/` directory to `docs/`  during doc build, and the copy only exists in the build output. (Same changes as in https://github.com/canonical/hardware-api/pull/456)
- Updated a few titles and section descriptions in the Reference section for clarity (page URLs are unchanged)
- Updated Contributing guide with instructions about how to check and update the API spec.

Preview: https://canonical-testflinger--852.com.readthedocs.build/852/reference/openapi/#/

## Web service API changes

(No API changes introduced. This PR adds infrastructure to document the existing API.)

## Tests

- PR check will fail if the current spec file cannot be generated, or is not consistent with the server API code
- Local test:
    ```shell
    # generate the spec file
    cd server/
    uvx --with tox-uv tox run -e schema

    # Check if update needed
    uvx --with tox-uv tox run -e check-schema
    ```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -104,13 +104,13 @@ same pull request. The CI check will fail if the spec is out of sync.
 To check if the specification is up-to-date, run:
 
 ```shell
-uvx --with tox-uv tox -e check-schema
+uvx --with tox-uv tox run -e check-schema
 ```
 
 If the check fails, regenerate the spec from the `server/` directory:
 
 ```shell
-uvx --with tox-uv tox -e schema
+uvx --with tox-uv tox run -e schema
 ```
 
 Commit the updated spec file with your API changes.
--- a/server/scripts/generate_openapi_schema.py
+++ b/server/scripts/generate_openapi_schema.py
@@ -84,7 +84,7 @@ def diff_schemas(local_schema_path: Path) -> bool:
         print("Error: OpenAPI schema is out of date", file=sys.stderr)
         print("", file=sys.stderr)
         print("To update the schema, run from server/ directory:", file=sys.stderr)
-        print(f"  python scripts/generate_openapi_schema.py --output {local_schema_path}", file=sys.stderr)
+        print(f"  uvx --with tox-uv tox run -e schema", file=sys.stderr)
         return False
 
     return True


server/tox.ini
```diff
@@ -55,3 +56,17 @@ set_env =
...
+description = Check server OpenAPI spec is up-to-date
+allowlist_externals = bash
+set_env =
+    FLASK_APP = devel.openapi_app
+commands =
+    bash -c 'if ! diff -u <(flask spec | python -m json.tool --sort-keys) <(python -m json.tool --sort-keys ./openapi.json) > /dev/null; then \
+        echo >&2 ""; \
+        echo >&2 "ERROR: OpenAPI spec is out of date."; \
+        echo >&2 "To update it, run:"; \
+        echo >&2 "  FLASK_APP=devel.openapi_app flask spec --output ./openapi.json"; \
```
rene-oromtz: I think we should refer to use uv directly instead, what do you think?
```suggestion
        echo >&2 "  FLASK_APP=devel.openapi_app uv run flask spec --output ./openapi.json --quiet"; \
```

CONTRIBUTING.md
```diff
@@ -95,6 +95,26 @@ uvx --with tox-uv tox
...
+```shell
+uvx --with tox-uv tox -e openapi-check
+```
+
+If the check fails, regenerate the spec from the `server/` directory:
+
+```shell
+cd server/
+source .venv/bin/activate
+FLASK_APP=devel.openapi_app flask spec --output ./openapi.json --quiet
```
rene-oromtz: Same comment about `uv run` since its just one command, maybe `uv run`  will be more practical without sourcing the environment:

```suggestion
FLASK_APP=devel.openapi_app uv run flask spec --output ./openapi.json --quiet
```

pedro-avalos:
I agree with @rene-oromtz I think the `API.md` file should go. The `openapi.yaml` file is a good standard to define the API endpoints. However, I think the `API.md` document has more metadata (such as error responses) that should be added to our endpoint definitions.

I think this PR can move along and merge without removing the `API.md` file, and a follow-up PR should both delete it and update any endpoints with any missing metadata. That should simplify the work that @tang-mm needs to do for this documentation update.

rene-oromtz:
Ah I was about to open a Jira issue to track this when I realized @tang-mm already created one! ([CERTTF-780](https://warthogs.atlassian.net/browse/CERTTF-780)). 

Please ignore my comment then related to `API.md`, just remains the minor nitpick about using uv

[CERTTF-780]: https://warthogs.atlassian.net/browse/CERTTF-780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

CONTRIBUTING.md
tang-mm: Thanks for the suggestion!

server/scripts/generate_openapi_schema.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Generate or validate OpenAPI schema for Testflinger server"
+    )
+    parser.add_argument(
+        "--output",
+        type=Path,
+        default=Path("./openapi.json"),
```
pedro-avalos: The docstring for this script says that running without arguments will print to stdout. This should therefore be `default=None`

tang-mm: Fix added in [3023c18](https://github.com/canonical/testflinger/pull/852/commits/3023c18b0041d57cbf89e17373d4f02db145a1d0)

CONTRIBUTING.md
```diff
@@ -95,6 +95,26 @@ uvx --with tox-uv tox
...
+To check if the specification is up-to-date, run:
+
+```shell
+uvx --with tox-uv tox -e check-schema
+```
+
+If the check fails, regenerate the spec from the `server/` directory:
+
+```shell
+uvx --with tox-uv tox -e schema
```
rene-oromtz: Hmmm seems that "schema" is reserved when using tox-uv but seems we can do
```suggestion
uvx --with tox-uv tox run -e schema
```

rene-oromtz: Error when not passing `run`
```
uvx --with tox-uv tox -e schema
usage: tox [-h] [--colored {yes,no}] [--stderr-color {BLACK,BLUE,CYAN,GREEN,LIGHTBLACK_EX,LIGHTBLUE_EX,LIGHTCYAN_EX,LIGHTGREEN_EX,LIGHTMAGENTA_EX,LIGHTRED_EX,LIGHTWHITE_EX,LIGHTYELLOW_EX,MAGENTA,RED,RESET,WHITE,YELLOW}] [-v | -q]
           [--exit-and-dump-after seconds] [-c file] [--workdir dir] [--root dir] [--runner {virtualenv,uv-venv-runner,uv-venv-lock-runner}] [--version] [--no-provision [REQ_JSON]] [--no-recreate-provision] [-r] [-x OVERRIDE]
           {run,r,run-parallel,p,depends,de,list,l,devenv,d,schema,config,c,quickstart,q,exec,e,legacy,le} ...
tox: error: unrecognized arguments: -e

```

CONTRIBUTING.md
```diff
@@ -95,6 +95,26 @@ uvx --with tox-uv tox
...
+## Server API Changes
+
+If you modified the server API (endpoints, schemas, or parameters), you must
+update and commit the OpenAPI specification file (`server/schemas/openapi.json`) in the
+same pull request. The CI check will fail if the spec is out of sync.
+
+To check if the specification is up-to-date, run:
+
+```shell
+uvx --with tox-uv tox -e check-schema
```
rene-oromtz: This does not break as check-schema is not reserved but for consistency maybe we can use the run as well
```suggestion
uvx --with tox-uv tox run -e check-schema
```

server/scripts/generate_openapi_schema.py
```diff
@@ -0,0 +1,134 @@
...
+
+    # Compare using compact normalized form
+    generated_normalized = normalize_json(generated)
+    local_normalized = normalize_json(local)
+
+    if generated_normalized != local_normalized:
+        print("Error: OpenAPI schema is out of date", file=sys.stderr)
+        print("", file=sys.stderr)
+        print("To update the schema, run from server/ directory:", file=sys.stderr)
+        print(f"  python scripts/generate_openapi_schema.py --output {local_schema_path}", file=sys.stderr)
```
rene-oromtz: Since this is outputted and the first indicator for the users, should we also point them to use `uvx --with tox-uv tox run -e schema` to prevent any dependency issue?

```suggestion
        print(f"  uvx --with tox-uv tox run -e schema", file=sys.stderr)
```

I don't mind much to keep it as-is as the usage is also documented in the CONTRIBUTING.Md but thought on asking for your opinion on this @tang-mm

rene-oromtz (changes requested):
Sorry for the long wait! Just added some minor changes but overall looks great to me, thanks @tang-mm!

server/scripts/generate_openapi_schema.py
tang-mm: Good catch! Thanks for the fix

pedro-avalos (approved):

PR #793, opened 2025-09-02
fix(cli): improve confusing queue position message (#461)

gajeshbhat:
## Summary

Fixes the confusing queue position message that was causing user confusion when jobs were first in the queue.

## Problem

When a job is first in the queue (position 0), testflinger displays:
```
This job is waiting on a node to become available.
Jobs ahead in queue: 0
```

This message is confusing because "0 jobs ahead" suggests nothing is being processed and the job should start immediately, when actually it means the job is first in line and will start when an agent becomes available.

## Solution

Updated the queue position messages to be more intuitive:

**Before:**
- `Jobs ahead in queue: 0` (confusing)
- `Jobs ahead in queue: 2` (only shows jobs ahead)

**After:**
- `Queue position: next in line` (clear for position 0)
- `Queue position: 3 (jobs ahead: 2)` (shows both absolute position and jobs ahead)

## Changes

- **Modified** `cli/testflinger_cli/__init__.py`: Updated `do_poll()` method to display clearer queue position messages
- **Added** comprehensive tests in `cli/testflinger_cli/tests/test_cli.py`:
  - `test_poll_queue_position_first_in_line`: Tests "next in line" message
  - `test_poll_queue_position_with_jobs_ahead`: Tests position display with jobs ahead  
  - `test_poll_queue_position_changes`: Tests position updates as queue moves

## Testing

- All existing tests pass (74/74)
- New tests pass (3/3 queue position tests)
- Code formatting and linting pass
- No regressions introduced

## User Impact

Users will now see clear, unambiguous queue position information that eliminates confusion about their job's status in the queue.

Fixes #461

PR #853, opened 2025-12-18
Certtf 756 display number of jobs

ajzobro:
## Description

Converted the list of queues displayed on the agent detail page into a table (matching the format of the queue_detail page exactly).

image

## Resolved issues

Resolves [CERTF-756](https://warthogs.atlassian.net/browse/CERTTF-756).

## Documentation

No documentation changes required for web interface change.

## Web service API changes

N/A

## Tests

Updated `server/devel/create_sample_data.py` to allow for a random number of queues to be attributed to each agent for better testing.


cli/testflinger_cli/__init__.py
```diff
@@ -533,51 +542,6 @@ def _get_agents_status(self):
         except UnknownStatusError as exc:
             sys.exit(exc)
 
-    def _get_jobs_status(self):
-        """Retrieve the status of jobs in a specified queue."""
-        try:
-            jobs_data = self.client.get_jobs_on_queue(self.args.queue_name)
```
rene-oromtz: I will probably debate to move all of this logic to the client module at least for now...

I also went through this pain on my CLI PR as this is becoming harder and harder to maintain and we also use redundant operations practically everywhere e.g using _get_job_status instead of directly calling the client module. 

I would suggest we gather with @pedro-avalos early next year and come up with a better approach and we restructure this CLI in a more dev friendly way. (using the Canonical CLI spec as a starting point)

Nothing wrong with moving it but if we are going to rework this anyway, I would say we just wait a bit more for a more structured approach and keep the modification minimal.

cli/testflinger_cli/__init__.py
```diff
@@ -1108,14 +1070,14 @@ def do_poll(
                         prev_queue_pos = queue_pos
                         if queue_pos == 0:
                             print(
-                                "This job will be picked up after the "
-                                "current job is complete (it is next in line)"
+                                "This job will be picked up after the current "
+                                "(1) job ahead of it is complete"
```
rene-oromtz: This may also not be so accurate (probably neither current nor proposed option). 

There are not necessarily (1) job ahead (or a current job waiting to be completed) as being in a queue 0 will also means that there are no jobs ahead but the agent is just about to pick it up. This message was also recently change in #793 but if clarification is still needed, maybe we would need a different rewording. 

If we do need to modify this, will probably say:

> This job is next in line in {queue} queue. 

Or something along those lines

docs/how-to/view-agent-status.rst
```diff
@@ -0,0 +1,24 @@
+View agent status
+=================
+
+You can view the status of an agent using the agent-status command. This command
```
rene-oromtz:
```suggestion
You can view the status of an agent using the ``agent-status`` command. This command
```

server/src/testflinger/api/v1.py
```diff
@@ -784,8 +784,17 @@ def check_valid_uuid(job_id):
...
 
 @v1.get("/job/<job_id>/position")
-def job_position_get(job_id):
-    """Return the position of the specified jobid in the queue."""
+# TODO: This is billed as providing the position in a queue describing at what
+#  point the job will be picked up, however there are certainly cases where
+#  this is an oversimplification as each agent listens to multiple queues.
+#  Therefore, there could be countless other jobs keeping the agents for the
+#  queue which this job is in busy rather than servicing this job.
+def job_position_get(job_id: str) -> tuple[str, int]:
```
rene-oromtz: This optionally also return just a string
```suggestion
def job_position_get(job_id: str) -> str | tuple[str, int]:
```

and I just noticed it also returns a dict! 
```suggestion
def job_position_get(job_id: str) -> dict | str | tuple[str, int]:
```

We don´t have type checkers but maybe for accuracy use option 2?

cli/testflinger_cli/__init__.py
```diff
@@ -1108,14 +1070,14 @@ def do_poll(
...
                             )
                         else:
                             print(
                                 f"This job will be picked up after the "
-                                f"current job and {queue_pos} job(s) ahead "
-                                f"of it in the queue are complete"
+                                f"{queue_pos + 1} jobs ahead of it in the "
```
rene-oromtz: Also not sure if this is also accurate. I would say that current job is no longer part of the queue and also, current job may not be part of the queue were the job was submitted.

rene-oromtz (changes requested):
Thanks for taking care of this!
I just reviewed an added some comment, personally, I think we should skip the modifications in the CLI as this would be a breaking change. 

I think the visual aid in the UI is useful but if we were to get the number or jobs in the queue, we already have `queue_status` which could be scripted with the queues from `agent_status`. 

Also, thanks for adding the docs! I think we also should create a separate issue to document more properly the CLI...

ajzobro:
> this would be a breaking change.

I looked at it this way: if looping over a "list" and now looping over a "dictionary" the new keys will be the same as the former values would have been.

I could add another argument to include the number of waiting jobs, but that is inconsistent with providing all the data.

The yaml format is much more human readable out of the box, and there was already yaml support for other commands, I merely expanded upon it.

Another option is to support the number of jobs in the yaml format, and skip the json format.

I find it strange that we feed in yaml files but then spit out json.

cli/testflinger_cli/__init__.py
ajzobro: I'm going to revert all the changes related to this aspect -- I found the output to be confusing upon my very first use, as my job had *not* been picked from the queue and even having 0 jobs ahead of me does not guarantee that my job will be picked first by any agent on the queue I scheduled with, as every agent on that queue may have other work that is higher priority (either having been queued for longer time or actually being queued with a true priority).  Therefore this information is presently inaccurate and not truly helpful as well as my changes to the verbiage not being any better.

server/src/testflinger/api/v1.py
```diff
@@ -784,8 +784,17 @@ def check_valid_uuid(job_id):
 
 
 @v1.get("/job/<job_id>/position")
-def job_position_get(job_id):
-    """Return the position of the specified jobid in the queue."""
+# TODO: This is billed as providing the position in a queue describing at what
```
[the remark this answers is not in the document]
ajzobro: Moved to own issue, #858.

rene-oromtz (dismissed):
This looks good to me!

rene-oromtz:
@ajzobro can you please take a look at the failing [CI](https://github.com/canonical/testflinger/actions/runs/20788540159/job/59734201396?pr=853). There is a long line that needs to be addressed before approving/merging

I rerun the other CI run since it was failing due to the bug described in #864 but it should pass in your next commit

rene-oromtz (approved):

PR #855, opened 2025-12-23
Fix: zapper_iot must define test_data issue

kiya956:
We default use test_data username/password, while test_data is not defined, fallback to use provision_data username/password not just "ubunut"/"ubuntu"

## Description
We default use test_data username/password, while test_data is not defined, fallback to use provision_data username/password not just "ubunut"/"ubuntu"

## Resolved issues

Solving wrong userame/password while test_data is not defined.

## Tests
1. provision imx8mp-pdk with test_data is defined and verified username/password is correct
2. provision imx8mp-pdk without test_data and verified that we fallback to use username/password in provision_data

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
@@ -38,11 +38,22 @@ class DeviceConnector(ZapperConnector):
         """Validate the job config and data and prepare the arguments
         for the Zapper `provision` API.
         """
+        # We prefer using username/password in provision_plan
+        # while username/password are not defined in test_data
+        provision_plan = self.job_data.get("provision_data", {}).get(
+            "provision_plan", {}
+        )
+
+        config = provision_plan.get("config", {})
+
+        default_uname = config.get("username", "ubuntu")
+        default_password = config.get("password", "ubuntu")
+
         username = self.job_data.get("test_data", {}).get(
-            "test_username", "ubuntu"
+            "test_username", default_uname
         )
         password = self.job_data.get("test_data", {}).get(
-            "test_password", "ubuntu"
+            "test_password", default_password
         )
         ubuntu_sso_email = self.job_data["provision_data"].get(
             "ubuntu_sso_email"
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/tests/test_zapper_iot.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/tests/test_zapper_iot.py
@@ -58,7 +58,7 @@ class ZapperIoTTests(unittest.TestCase):
         self.assertEqual(args, ())
         self.assertDictEqual(kwargs, expected)
 
-    def test_validate_configuration_provision_plan(self):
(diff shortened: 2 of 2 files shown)


artiepoole (approved):
Thank you @kiya956 , this seems to solve my problem exactly.

p-gentili (changes requested):
This works well with provisioning but it will fail during the test phase, mostly because the username there defaults to `ubuntu`.

What's your use case @artiepoole ? Only provisioning?

artiepoole:
> What's your use case @artiepoole ? Only provisioning?

Sorry, it has been a few months. I have this message I sent to An which lead to this PR:

> Hey An, 
>
> [redacted]
> can we change the default of "ubuntu" to provision_data's username? As it stands, the preset and the test_username do not match by default and I have to add 
>
> test_data:
>   test_username: 'ce-certification-qa@lists.canonical.com'
>   test_cmds: |
> in order to stop this defaulting to ubuntu and mismatching when setting the username because of
>
>       username = data.get("test_data", {}).get("test_username", "ubuntu")
> and the error prints here:
> [redacted]
>
> I feel like, typically, the provision data's username/password is the same as the test username/password most of the time?

Which makes me think that the issue was with provisioning a device (a check was failing and so provisioning failed) attempting to do UC24 checkbox testing.

~~Note: I think I have installed this version on my devices and it has been working well, but they may have been updated since.~~ Actually I do not know what version of testflinger is running on our agents. It seems like a workaround may have been added to the oemqa-ci repo to automatcally fix this when using envicorn

p-gentili:
> Actually I do not know what version of testflinger is running on our agents. It seems like a workaround may have been added to the oemqa-ci repo to automatcally fix this when using envicorn

Ok, I guess we can close this PR for now. Ubuntu Core with console conf indeed requires the user to provide an email as username. Internally, we switched to system-user assertions. You guys might want to do the same.

kiya956:
@p-gentili Hi Paolo, 
The purpose of this is for provision without define test_data. In PE, CE  have a high probability that they only want to provision machines and debug it. And I don't think this could cause test phase fail if provision phase success.

And sorry, I don't get what this mean
```
 Internally, we switched to system-user assertions.
```

p-gentili:
Hey @kiya956, ok let's take care of that. Although, I'll need you to add a warning in case the `username` is different from `test_username` as defined below.
```
test_username = self.job_data.get("test_data", {}).get("test_username", "ubuntu")
```  

A warning like:
```
Provisioning is using a username different from what the test phase expects, which may prevent it from accessing the DUT later on.
```

kiya956:
> Hey @kiya956, ok let's take care of that. Although, I'll need you to add a warning in case the `username` is different from `test_username` as defined below.
> 
> ```
> test_username = self.job_data.get("test_data", {}).get("test_username", "ubuntu")
> ```
> 
> A warning like:
> 
> ```
> Provisioning is using a username different from what the test phase expects, which may prevent it from accessing the DUT later on.
> ```

Hi @p-gentili I found I can not git push, would you like to grant me permission? Thanks

p-gentili:
Closing in favor of https://github.com/canonical/testflinger/pull/1126

PR #880, opened 2026-01-16
fix: update the default release from xenial to jammy in maas.py 

### LINKED ISSUE #318: maas2 connector: fallback to out-of-standard-support xenial (16.04)
The fallback distro for the `maas2` connector is still Ubuntu 16.04 LTS (which [according to ubuntu.com/16-04](https://ubuntu.com/16-04) is supported through ESM only, with standard support ending in April 2021):

https://github.com/canonical/testflinger/blob/29247d8726d337f7192e8cde75f2017f45ccbf25/device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py#L78-L80

Note also that `self.job_data.get("provision_data")` might evaluate to `None` if no `provision_data` is provided at all, possibly this should be:

```python3
 provision_data = self.job_data.get("provision_data", {}) 
```

As for "xenial", it should probably be replaced by "jammy" or even "noble" at this point?

bladernr:
## Description

Updates the default fallback release from xenial to  jammy since xenial is no longer supported and very likely to not exist on a current maas system.

## Resolved issues
Resolves #318 
Resoves CERTTF-366

## Documentation

NA

## Web service API changes

NA

## Tests

NA

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
@@ -83,7 +83,7 @@ class Maas2:
             self.clear_tpm()
         provision_data = self.job_data.get("provision_data")
         # Default to a safe LTS if no distro is specified
-        distro = provision_data.get("distro", "xenial")
+        distro = provision_data.get("distro", "jammy")
         kernel = provision_data.get("kernel")
         user_data = provision_data.get("user_data")
         storage_data = provision_data.get("disks")
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+artifacts
 build
 dist
 env
@@ -17,8 +18,12 @@ docs/.sphinx/.doctrees/
 *.egg*
 *.bak
 *.charm
+*.swp
 
 _build
 __pycache__
 
 *.snap
+
+# AI agent instructions (user-specific)
+.github/copilot-instructions.md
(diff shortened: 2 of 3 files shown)


device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
```diff
@@ -113,7 +113,7 @@ def _install_efitools_snap(self):
             "-o",
             "UserKnownHostsFile=/dev/null",
             "ubuntu@{}".format(self.config["device_ip"]),
-            "sudo snap alias efi-tools-ijohnson.efibootmgr efibootmgr",
+            "sudo snap alias efi-tools-ijohnson.efibootmgr efibootmgr"
```
ajzobro: Would reformat: src/testflinger_device_connectors/devices/maas2/maas2.py
1 file would be reformatted,

Please run `uv run ruff check --fix`

bladernr: Yeah, that is apparently due to our decision to change formatters (yet again) and the fiile had never been touched :D  Reformatted and pushed... sigh, and now stupid codecov...

ajzobro (approved):
Looks good.  Though, I'm not a fan of the review changes getting re-based/squashed instead of merged as I can't review a delta from my last review.

I do think that we ought to squash-merge into the main branch, but allow for each change after a review is started to be it's own commit.

PR #881, opened 2026-01-17
docs: add example on artifacts usage

rene-oromtz:
## Description
This is a minor doc change to include an example on artifacts usage. 

## Resolved issues
Untracked

## Documentation
Add example on current artifact section

## Tests
Built locally:
Screenshot from 2026-01-16 18-21-01

### DIFF
--- a/docs/reference/test-phases.rst
+++ b/docs/reference/test-phases.rst
@@ -414,7 +414,8 @@ For example, under ``test_cmds``, you can create an artifacts directory and copy
 
 .. note::
   
-  From the above example, the first command is executed on the remote device, while the next two commands are executed on the Testflinger Agent.
+  From the above example, the command enclosed in single quotes is executed on the remote device via SSH, 
+  while the next two commands are executed on the Testflinger Agent.
 
 Artifacts can later be retrieved using the CLI:
 


docs/reference/test-phases.rst
```diff
@@ -389,4 +389,35 @@ See :doc:`../how-to/retrieve-logs` for detailed information on how to retrieve a
...
+    <provision-data>
+  test_data:
+    test_cmds: |
+      ssh ubuntu@$DEVICE_IP 'uname -a > /tmp/system_info.txt'
+      mkdir -p artifacts
+      scp ubuntu@$DEVICE_IP:/tmp/system_info.txt artifacts/
+
+.. note::
+  
+  From the above example, the first command is executed on the remote device, while the next two commands are executed on the Testflinger Agent.
```
ajzobro:
```suggestion
  From the above example, the `uname -a > /tmp/system_info.txt` command is executed on the remote device, while the next two commands are executed on the Testflinger Agent.
```

If this correction is true, then **_all_** commands are being executed on the "Testflinger Agent" and only by using ssh is the `uname` command being executed on the device.

ajzobro (changes requested):
I think that the interpretation of what a command is would include the ssh command, and thus it seems that all test commands run on the agent. It just so happens that the first agent command is requesting a command be executed on the device.

ajzobro (approved):

PR #882, opened 2026-01-19
change: don't log docker error when removing container

p-gentili:
## Description

```
***************************************************************************
* Starting testflinger setup phase on hp-pro-sff-400-g9-desktop-pc-c30464 *
***************************************************************************
Cleaning up container if it exists...
Error response from daemon: No such container: hp-pro-sff-400-g9-desktop-pc-c30464
```

Normally, there's no leftover from previous runs, but we try anyway to cleanup the container in case it still exists. The message usually displayed is misleading for a normal user: it makes them think there has been an issue while this is actually the expected situation.

I know I might be redirecting too much, at the same time I didn't want to complicate these scripts too much.

## Resolved issues

N/A

## Documentation

N/A

## Web service API changes

No

## Tests

I can't actually test this, can I?


antlassagne:
I don't have the power to approve, but I approve

rene-oromtz (dismissed):
Thanks for addressing this! For simplicity, I think this approach is the best

rene-oromtz (approved):

PR #892, opened 2026-01-21
Fix: Update code and tests to use HTTPStatus enumerations throughout instead of magic numeric constants

ajzobro:
## Description

Updated all code and tests to comply with the use of HTTPStatus rather than magic numbers for HTTP status codes

## Resolved issues

N/A

## Documentation

N/A

## Web service API changes

N/A

## Tests

Unit tests were ran for each component: cli, agent, server.


rene-oromtz (approved):
Thanks for taking care of this!

PR #893, opened 2026-01-21
Fix: Update string concatenations to use f-strings throughout

ajzobro:
## Description

Search for and updated all instances of regular strings that were concatenated with variables and promoted them to f-strings.

## Resolved issues

N/A

## Documentation

N/A

## Web service API changes

N/A

## Tests

Ran unit tests across all components: cli, agent, server.


rene-oromtz (changes requested):
Thanks for taking care of this! I just added a minor comment about a missing HTTPStatus that seems was not included in the previous PR

rene-oromtz (approved):

PR #898, opened 2026-01-28
zapper_oem to support runtime autoinstall generation method

kiya956:
## Description

Some of our image that include autoinstall template
and can be generated to real autoinstall script during runtime
We need to add user to template but copy a user data file

## Resolved issues

Can not provision GB300 project


Artur-at-work (approved):
This keeps default behavior and adds option to enable provisioning the bolt

Artur-at-work (changes requested):
As above, need to be added in schema (i.e `user_template_update = fields.Boolean(required=False)` )

kiya956:
> As above, need to be added in schema (i.e `user_template_update = fields.Boolean(required=False)` )

done

Artur-at-work (approved):
Lgtm, thanks

PR #899, opened 2026-01-28
chore(agent): Replace operator_libs_linux with charmlibs

pedro-avalos:
## Description

- This PR replaces `operator_libs_linux` (Charm library) with equivalent `charmlibs` (Python dependency)
- This simplifies our repository contents (since we don't need to vendor in these dependencies)
- This PR also doesn't add the `charmlibs-systemd` library because it seems we were not using it?

## Resolved issues

Untracked

## Documentation

n/a

## Web service API changes

n/a

## Tests

- Existing unit and integration tests


rene-oromtz (approved):

rene-oromtz (approved):
Thanks for the changes! This looks way cleaner, and yes, for some reason we were importing systemd but never really used it...

PR #901, opened 2026-02-03
chore: format pyproject for all components

rene-oromtz:
## Description
This is a small PR that attempt to bring more structure to our pyprojects.toml

Instead of specifying the files to be lint/formatted in the cmd line for the tox environment, this uses a more standard approach to use the `include` and `extend-exclude` section in pyproject.toml

## Resolved issues
NA

## Tests
No test added as this just moves the path from the cmd line to the pyproject.

### DIFF
--- a/agent/pyproject.toml
+++ b/agent/pyproject.toml
@@ -47,6 +47,12 @@ test = [
 
 [tool.ruff]
 line-length = 79
+extend-exclude = ["charms/testflinger-agent-host-charm/lib"]
+include = [
+    "src/**/*.py",
+    "tests/**/*.py",
+    "charms/**/*.py",
+]
 
 [tool.ruff.lint]
 select = [
--- a/agent/pyproject.toml
+++ b/agent/pyproject.toml
@@ -47,7 +47,7 @@ test = [
 
 [tool.ruff]
 line-length = 79
-extend-exclude = ["charms/testflinger-agent-host-charm/lib"]
+extend-exclude = ["charms/*/lib"]
 include = [
     "src/**/*.py",
     "tests/**/*.py",
--- a/agent/tox.ini
+++ b/agent/tox.ini
@@ -1,7 +1,6 @@
 [tox]
 env_list = 
     lock
-    format
     lint
     unit
 no_package = true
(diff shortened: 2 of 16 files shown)


agent/pyproject.toml
```diff
@@ -47,6 +47,12 @@ test = [
 
 [tool.ruff]
 line-length = 79
+extend-exclude = ["charms/testflinger-agent-host-charm/lib"]
+include = [
```
pedro-avalos: Since these can be directories, perhaps we can simplify to

```toml
include = ["src", "tests", "charms/**/src"]
```

agent/pyproject.toml
```diff
@@ -47,6 +47,12 @@ test = [
 
 [tool.ruff]
 line-length = 79
+extend-exclude = ["charms/testflinger-agent-host-charm/lib"]
```
pedro-avalos: Might want to generalize it?

```suggestion
extend-exclude = ["charms/*/lib"]
```

pedro-avalos (changes requested):
Some nitpicks and a comment regarding the tox files change.

agent/tox.ini
```diff
@@ -21,12 +21,18 @@ commands =
...
+    ruff format --check
 
 [testenv:lint]
 description = Run linting tests
-commands = 
-    ruff check src tests charms/testflinger-agent-host-charm/src charms/testflinger-agent-host-charm/tests
+commands =
+    ruff check
+
+[testenv:fix]
```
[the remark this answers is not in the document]
pedro-avalos: Might need to update contributing guides too

agent/tox.ini
```diff
@@ -1,7 +1,6 @@
 [tox]
 env_list = 
     lock
-    format
```
pedro-avalos: I think we should run this by default anyways. That's what we do with charms

pedro-avalos (changes requested):
I think we should run `format` by default, but otherwise this looks great

agent/tox.ini
rene-oromtz: Sounds good, I re-added the `format` env in tox default

pedro-avalos (approved):

PR #902, opened 2026-02-04
fix(oem_autoinstall): Raise ProvisioningError instead of RecoveryError to avoid agent going offline

Artur-at-work:
## Description
Problem:
When DUT is not ssh-able, `oem_autoinsall` will call `reboot_script` to execute snmp commands against that DUT. When power cycler is offline, snmp commands timeout and return failure, so reboot_script returns failure, then `oem_autoinsall` raises `RecoveryError` with code 46, which causes TF agent to turn itself offline.

Next time submit to this agent will fail due to agent being offline. This requires manual intervention to set agent back online.
I think this behavior started after we introduced the offline state for agents.

Fix:
This change will raise another exception type, but still fail the provisioning. So user still gets the error and expected to recover the dut

## Resolved issues
Closes: https://warthogs.atlassian.net/browse/OEX86-927

## Documentation
na

## Web service API changes
na

## Tests
1. Run agent locally and reproduce the problem by setting wrong snmp ip
2. Observer TF agent offline
3. Submit job and confirm failed
4. Set agent back online
5. (apply the fix). Submit job and confirm agent still online after snmp failed
6. Submit job again
Log output is in Resolved Issue.

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/oem_autoinstall/oem_autoinstall.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/oem_autoinstall/oem_autoinstall.py
@@ -30,7 +30,6 @@ import yaml
 
 from testflinger_device_connectors.devices import (
     ProvisioningError,
-    RecoveryError,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


kiya956 (approved):
LGTM

PR #915, opened 2026-02-06
chore: test main loop

ajzobro:
## Description
- **Problem**: The agent's main loop (`start_agent()`) lacked integration test coverage, making it difficult to verify end-to-end behavior during job processing, server connectivity issues, and offline state handling.
- **Approach**: Added integration tests for the main loop covering three critical scenarios: retry behavior after 401 errors, continuous polling with no available jobs, and offline state detection. Uses mocking of client/agent components to isolate the loop logic.

## Resolved issues

Test coverage improvement.

## Documentation

N/A

## Web service API changes

N/A

## Tests

- **Unit tests**: `agent/tests/test_main_loop.py`
  - `test_main_loop_retries_after_401`: Verifies loop continues after failed job check, calls `process_jobs` twice across poll cycles
  - `test_main_loop_handles_offline`: Verifies loop skips job processing when agent is marked offline
- **Run tests**: From `agent/` directory, run `uvx tox -e unit`


agent/tests/test_main_loop.py
```diff
@@ -0,0 +1,129 @@
...
+
+        with patch("testflinger_agent.load_config", return_value=config):
+            with patch("testflinger_agent.configure_logging"):
+                with patch(
+                    "testflinger_agent.TestflingerClient"
+                ) as mock_client_class:
+                    with patch(
+                        "testflinger_agent.TestflingerAgent"
+                    ) as mock_agent_class:
+                        with patch(
```
rene-oromtz: Would you consider using `@patch` decorators instead for preventing the multiple nesting?

agent/tests/test_main_loop.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def test_main_loop_handles_offline(self, config):
+        """Test main loop exits when agent goes offline."""
+        sleep_counter = [0]
+
+        def sleep_side_effect(interval):
+            sleep_counter[0] += 1
+            if sleep_counter[0] >= 1:
+                # End test after prescribed number of loops.
+                raise KeyboardInterrupt()
```
rene-oromtz: Same comment on using a mock side effect instead

rene-oromtz (changes requested):
Thanks for the added tests. I included some comments for overall using fixtures or patch decorators and also a clarification note! Please let me know what you think!

rene-oromtz (changes requested):
Just added some comments about some mocks and one nitpick, besides that look good to me. 

Also, in #921 I added a `test_init.py` test file, since this main loop is part of the `__init__` file we may want to copy/paste these tests there so we have a 1:1 test file per code file, what do you think?

rene-oromtz (approved):
LGTM, thanks for the changes

PR #918, opened 2026-02-06
ci: run agent tests on device-connectors modifications

rene-oromtz:
## Description
This minor PR allows agent tests to run whenever there is any modification on device-connectors. Device Connectors are used on agent tests so if any modification is made, we need to make sure it doesn't impact the agent e.g. modify agent lock file. 

This also fixes the uv.lock file for agent

## Resolved issues
NA

## Tests
NA


xobes (approved):
Good catch

pedro-avalos (approved):

PR #932, opened 2026-02-25
fix: longer timeout after power cycle

p-gentili:
## Description

The default timeout was defined based on the assumption the control host was already online - just waiting for the API server. In this particular case though, we need to wait for the whole boot process + service startup. 5 minutes should be enough.

## Resolved issues

N/A

## Documentation

N/A

## Web service API changes

N/A

## Tests

N/A


amalinowski75 (approved):
Makes sense

PR #951, opened 2026-03-03
fix(agent): remove writing to old endpoint to address BSON limit

rene-oromtz:
## Description
This PR removes writing to the old endpoint as its no longer needed, besides its creating problem with agents attempting to write to the old endpoint as its hitting the 16 MB BSON limit (single document for all chunks). See following jobs for massive serial logs:
`curl https://testflinger.canonical.com/v1/result/4c07d47c-cebc-4b30-a6ce-bf7fcd38eaaf/log/serial`

New endpoint does not have this problem because each chunk has its own document so it won't hit this limit. 

Error seen:
```
[26-03-03 15:17:38]    INFO: (client.py:381)| Fallback to new log endpoint
[26-03-03 15:17:49]   ERROR: (client.py:380)| HTTPSConnectionPool(host='testflinger.canonical.com', port=443): Max retries exceeded with url: /v1/result/4c07d47c-cebc-4b30-a6ce-bf7fcd38eaaf/serial_output (Caused by ResponseError('too many 500 error responses'))
[26-03-03 15:17:49]    INFO: (client.py:381)| Fallback to new log endpoint
[26-03-03 15:18:00]   ERROR: (client.py:380)| HTTPSConnectionPool(host='testflinger.canonical.com', port=443): Max retries exceeded with url: /v1/result/4c07d47c-cebc-4b30-a6ce-bf7fcd38eaaf/serial_output (Caused by ResponseError('too many 500 error responses'))
[26-03-03 15:18:00]    INFO: (client.py:381)| Fallback to new log endpoint
```

Jobs are then stuck trying to send all chunks for serial logging which takes 3 retries for each chunk. 

## Resolved issues
NA

PR #952, opened 2026-03-04
fix(agent): increase chunk size

rene-oromtz:
## Description
This increases the chunk size that I forgot to add in #951
Instead of 1KB uses 1MB as files can be quite large:
`-rw-r--r--  1 ubuntu ubuntu 43380221 Mar  4 00:50 test-serial.log`

Used 1MB as according to [this](https://stackoverflow.com/questions/2880722/can-http-post-be-limitless) 1MB is default for NGINX. In our staging/prod instance, the max body size limit is 500MB. 

## Resolved issues
NA


ajzobro (approved):
LGTM

PR #958, opened 2026-03-06
Add set_ubuntu_boot for image provision

kiya956:
that mkae sure second boot would not boot to usb

## Description
Add set_ubuntu_boot for image provision
that mkae sure second boot would not boot to usb

## Resolved issues
during provision we need to reboot for EFI to detect usb.
But it could cause system boot into usb directly.
I case this happen, we set ubuntu boot before reboot
## Tests
script can actually set the boot order


Artur-at-work (approved):
LGTM

PR #959, opened 2026-03-06
change: migrate from RPyC to REST APIs

p-gentili:
## Description

As per title, this PR migrate access to the control host from RPyC to REST: it should be transparent to the user.

## Resolved issues

https://github.com/canonical/testflinger/security/dependabot/20
https://github.com/canonical/testflinger/security/dependabot/30

## Documentation

N/A

## Web service API changes

N/A

## Tests

*Pending*

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
@@ -22,12 +22,10 @@ validating the configuration and preparing the API arguments.
 
 import json
 import logging
-import subprocess
 from abc import ABC, abstractmethod
 from typing import Any, Dict, Tuple
 
 import requests
-import rpyc
 
 from testflinger_device_connectors.devices import (
     DefaultDevice,
@@ -44,7 +42,6 @@ class ZapperConnector(ABC, DefaultDevice):
 
     PROVISION_METHOD = ""  # to be defined in the implementation
     ZAPPER_REQUEST_TIMEOUT = 60 * 90
-    ZAPPER_SERVICE_PORT = 60000
     ZAPPER_REST_PORT = 8000
 
     def _api_post(self, endpoint: str, **kwargs) -> requests.Response:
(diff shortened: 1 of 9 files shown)


device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
```diff
@@ -209,12 +226,35 @@ def _run(self, zapper_ip, *args, **kwargs):
...
+        job = resp.json()
+        job_id = job["job_id"]
+
+        # Stream SSE logs
+        url = (
+            f"http://{zapper_ip}:{self.ZAPPER_REST_PORT}"
+            f"/api/v1/provision/{job_id}/logs"
         )
+        with requests.get(
+            url, stream=True, timeout=self.ZAPPER_REQUEST_TIMEOUT
```
ajzobro: Q: Will this timeout be for the whole stream or in between data chunks?
A: https://docs.python-requests.org/en/latest/user/advanced/#timeouts

There can be a connection time out as well as a timeout between bytes of the streamed read.

Do we want both?

ajzobro (commented):
I have also been pushing for better test coverage. In particular aiming for 100% on new code. There are some complexities to the streamed data in particular that I think warrants some testing to ensure that the code will behave as expected and not leave things crashing out where not expected.

Happy to see this!

device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
```diff
@@ -65,63 +63,96 @@ def _api_post(self, endpoint: str, **kwargs) -> requests.Response:
...
 
-    @staticmethod
-    def _check_rpyc_server_on_host(host: str) -> None:
+    def _api_get(self, endpoint: str, **kwargs) -> requests.Response:
+        """Send a GET request to the Zapper REST API.
+
+        :param endpoint: API endpoint path.
+        :param kwargs: Additional keyword arguments passed to requests.get.
+        :returns: The response object.
+        :raises requests.RequestException: On any request failure.
```
ajzobro:
```
        :raises requests.RequestException: On any request failure or timeout.
```
This aspect of the comment could be updated to address the specific part that caught my attention before.  It was the concept of "failure" as opposed to a "timeout" (arguably a different failure, but also arguably different from a failure).

Doing nothing is also fine.

ajzobro (approved):

PR #962, opened 2026-03-11
fix(submit): add a wait for completion step

rene-oromtz:
## Description
Recent modification of the submit action made it possible to poll test phases individually for readability purposes. However, the old mechanism uses a bare `poll` which actually wait for the job to be completed. 
This adds an additional step to allow for job completion only if job is NOT a reserve job. This is needed for jobs that rely on job completion to download artifacts.

### DIFF
--- a/.github/actions/submit/action.yaml
+++ b/.github/actions/submit/action.yaml
@@ -240,6 +240,23 @@ runs:
         [ -n "$DEVICE_IP_FLAG" ] && echo "::notice::Device IP: $DEVICE_IP"
         exit $EXIT_STATUS
 
+    - name: Wait for job completion
+      if: inputs.poll == 'true' && inputs.dry-run != 'true' && steps.check-phases.outputs.reserve == 'false'
+      shell: bash
+      env:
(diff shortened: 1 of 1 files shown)


ajzobro (commented):
I think this looks good.

I want to confirm that this section is intended to run only when the RESERVE section is not expected to run.

I am quite curious as to why why even have a reserve section in a GH action;

I am also curious about how we will arrive at the "Cancel the job if the action is cancelled" section -- won't we likely be caught up in a while loop above?

And finally, with respect to the workflow zizmor issues; I agree, can you make a new issue to track it?

rene-oromtz:
1. Reserve section in a GH action: Needed because there are some teams that deploy via workflows and the only thing they need is the device IP (provided by the action) and knowing the machine is reserved for their own purposes. No need to wait until reservation expires because they are doing it by themselves. 
2. Cancel the job if the action is cancelled: This only happens if the action is cancelled via Github, not via tf cli
3. Sure thing! Created #963

ajzobro (approved):
These changes LGTM

PR #1005, opened 2026-04-07
new: allow using 'url' instead of `urls` when only one image is required

p-gentili:
## Description

When provisioning IoT devices it's frequently needed to provide multiple assets (bootloader, rootfs, etc). There are cases where this is not needed though, and it feels awkward to still use an array ( `urls` ), which is also different from every other device connector available in this repo.

Allow using `url` instead of `urls` in case only one asset is required.

## Resolved issues

N/A

## Documentation

Yes, but not here.

## Web service API changes

N/A

## Tests

Unit testing only.

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
@@ -90,7 +90,11 @@ class DeviceConnector(ZapperConnector):
             except (ValueError, KeyError) as e:
                 raise ProvisioningError from e
 
-        urls = self.job_data["provision_data"].get("urls", [])
+        if url := self.job_data["provision_data"].get("url"):
+            urls = [url]
+        else:
+            urls = self.job_data["provision_data"].get("urls", [])
+
         try:
             validate_urls(urls)
         except ValueError as e:
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
@@ -90,14 +90,10 @@ class DeviceConnector(ZapperConnector):
             except (ValueError, KeyError) as e:
                 raise ProvisioningError from e
 
-        url = self.job_data["provision_data"].get("url")
-        urls = self.job_data["provision_data"].get("urls", [])
-        if url and urls:
-            raise ProvisioningError(
-                "Provide only one of provision_data.url or provision_data.urls"
-            )
-        if url:
+        if url := self.job_data["provision_data"].get("url"):
             urls = [url]
+        else:
+            urls = self.job_data["provision_data"].get("urls", [])
 
         try:
             validate_urls(urls)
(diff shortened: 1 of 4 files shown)


device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
```diff
@@ -90,7 +90,15 @@ def _validate_configuration(
...
 
+        url = self.job_data["provision_data"].get("url")
         urls = self.job_data["provision_data"].get("urls", [])
+        if url and urls:
+            raise ProvisioningError(
+                "Provide only one of provision_data.url or provision_data.urls"
+            )
+        if url:
+            urls = [url]
+
```
rene-oromtz: I don't mind if we keep it as-is as an additional safety net, but since the job will now be rejected on server side. Could you consider something as this:
`urls = self.job_data["provision_data"].get("urls") or [self.job_data["provision_data"].get("url")]`

Server side needs to be refreshed first to enable the use of `url` and `urls` so there shouldn't be a risk of any job specifying both keys

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/tests/test_zapper_iot.py
```diff
@@ -34,6 +34,45 @@ def test_validate_configuration(self):
...
+        device.job_data = {
+            "provision_data": {
+                "preset": "TestPreset",
+                "url": "http://test.tar.gz",
+                "urls": ["http://other.tar.gz"],
+            }
+        }
+
+        with self.assertRaises(ProvisioningError):
+            device._validate_configuration()
```
rene-oromtz: This test can potentially be rewritten or removed if following this [comment](https://github.com/canonical/testflinger/pull/1005/changes#r3059551470)

rene-oromtz (dismissed):
LGTM! I added just a small comment that may be useful as the `url` and `urls` are not meant to be in any job (enforced server side) but I don't mind much to keep it as-it. I'll let you decide which one you prefer

p-gentili:
Hey @rene-oromtz, while working on your suggestion, I realize we'll need to make `url(s)` optional in the near future (a week), to support providing the image as attachment. Which means yet another server upgrade. Do you prefer to have a single PR for both changes?

rene-oromtz:
@p-gentili I don't mind if we upgrade the server twice as this changes are non disruptive. It will really depend on how large is the change to  support providing the image as an attachment. If it ends being too large, I'll prefer doing it separate to make it easier to review.

device-connectors/src/testflinger_device_connectors/devices/zapper_iot/__init__.py
p-gentili: Ok, I've removed the validation from this side as recommended. I didn't follow your snippet because, in the remote case one doesn't provide any of url/urls (it can happen from source), it would end up with `urls = [None]`. Also, formatting gets weird.

rene-oromtz (approved):
LGTM! Thanks for the changes

PR #1009, opened 2026-04-09
Add release and identifier labels to phase_duration metric.

gntzio:
## Description

To enable more meaningful data separation for the provisioning duration metric, this PR adds 2 labels to the `phase_duration` metric: `identifier` and `release`.

- `identifier` is derived from the `identifier` field in `testflinger-agent.conf`.
- `release` indicates the OS with which the DUT was provisioned. `release` becomes available after the provisioning phase and is applied to all subsequent phases, including provisioning. Implementation details:
  - After provisioning is finished, the TF Agent will ssh the DUT to read `/etc/os-release` information.
  - If successful, the Agent will normalizes the version to match the C3 definition (the logic is documented in [test_os_release.py](https://github.com/canonical/testflinger/blob/04af98123375cc2eca7bce64ffb3761a6133e110/agent/tests/test_os_release.py))
  - The normalization ensures consistent behavior in Grafana dashboards.

_Note: Additional rationale for the selected approach is provided in the linked ticket._

## Resolved issues

[CERTTF-832](https://warthogs.atlassian.net/browse/CERTTF-832)

## Documentation

Unit tests + linked ticket.

## Web service API changes

N/A

## Tests

Unit tests.

[CERTTF-832]: https://warthogs.atlassian.net/browse/CERTTF-832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/src/testflinger_agent/agent.py
+++ b/agent/src/testflinger_agent/agent.py
@@ -33,6 +33,7 @@ from testflinger_agent.handlers import (
 )
 from testflinger_agent.job import TestflingerJob
 from testflinger_agent.metrics import PrometheusHandler
+from testflinger_agent.os_release import query_dut_release
 
 try:
     # attempt importing a tarfile filter, to check if filtering is supported
@@ -299,6 +300,8 @@ class TestflingerAgent:
             rundir = None
             job = None
             event_emitter = None
+            release = ""
+            identifier = self.client.config.get("identifier", "")
             try:
                 job = TestflingerJob(job_data, self.client)
                 event_emitter = EventEmitter(
@@ -384,9 +387,27 @@ class TestflingerAgent:
                     exit_code, exit_event, exit_reason = job.run_test_phase(
                         phase, rundir
                     )
-                    # Measure phase duration and report to metrics handler
+                    phase_end = time.time()
+
+                    # Query DUT release after successful provision
+                    if phase == TestPhase.PROVISION and not exit_code:
+                        device_info = job.get_device_info(rundir)
+                        if device_info and (
+                            device_ip := device_info.get(
+                                "device_info", {}
+                            ).get("device_ip")
+                        ):
+                            username = (job_data.get("test_data") or {}).get(
+                                "test_username", "ubuntu"
+                            )
+                            release = query_dut_release(device_ip, username)
+
+                    # Report phase duration to metrics handler
                     self.metrics_handler.report_phase_duration(
-                        phase, int(time.time() - phase_start)
+                        phase,
+                        int(phase_end - phase_start),
+                        identifier=identifier,
+                        release=release,
                     )
                     self.client.post_influx(phase, exit_code)
                     event_emitter.emit_event(exit_event, exit_reason)
--- a/agent/src/testflinger_agent/os_release.py
+++ b/agent/src/testflinger_agent/os_release.py
@@ -1,0 +1,130 @@
+# Copyright (C) 2026 Canonical
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>
+
+"""Query and parse os-release information from a provisioned DUT."""
+
+import logging
+import re
+import subprocess
+
+logger = logging.getLogger(__name__)
+
+UNKNOWN_RELEASE = "unknown"
+
+SSH_TIMEOUT = 30
+SSH_OPTIONS = [
+    "-o",
+    "StrictHostKeyChecking=no",
+    "-o",
+    "UserKnownHostsFile=/dev/null",
+    "-o",
+    "ConnectTimeout=10",
+    "-o",
+    "BatchMode=yes",
+    "-o",
+    "LogLevel=ERROR",
+]
+
+
+def query_dut_release(device_ip: str, username: str = "ubuntu") -> str:
+    """SSH into a DUT and extract the release label from /etc/os-release.
+
+    Authentication uses SSH key-based auth, no password is required.
+
+    :param device_ip: IP address of the device under test
+    :param username: SSH username (default: "ubuntu")
+    :return: Release label string, or UNKNOWN_RELEASE on failure
+    """
+    cmd = [
+        "ssh",
+        *SSH_OPTIONS,
+        f"{username}@{device_ip}",
+        "cat /etc/os-release",
+    ]
+
+    try:
+        result = subprocess.run(  # noqa: S603
+            cmd,
+            capture_output=True,
+            text=True,
+            timeout=SSH_TIMEOUT,
+            check=False,
+        )
+        if result.returncode != 0:
+            logger.warning(
+                "Failed to query os-release from %s (rc=%d): %s",
+                device_ip,
+                result.returncode,
+                result.stderr.strip(),
+            )
+            return UNKNOWN_RELEASE
+
+        os_info = _parse_os_release(result.stdout)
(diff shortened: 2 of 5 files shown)


agent/src/testflinger_agent/agent.py
```diff
@@ -384,9 +387,25 @@ def process_jobs(self):
...
-                    # Measure phase duration and report to metrics handler
+                    phase_end = time.time()
+
+                    # Query DUT release after successful provision
+                    if phase == TestPhase.PROVISION and not exit_code:
+                        device_info = job.get_device_info(rundir)
+                        if device_info and (
+                            device_ip := device_info.get("device_ip")
+                        ):
+                            username = job_data.get("test_data", {}).get(
```
rene-oromtz: A valid yaml can include an empty value for `test_data` key so this needs to be:
`(job_data.get("test_data") or {})`
I think this is really low risk but better safe than sorry

agent/src/testflinger_agent/agent.py
```diff
@@ -384,9 +387,25 @@ def process_jobs(self):
...
                         phase, rundir
                     )
-                    # Measure phase duration and report to metrics handler
+                    phase_end = time.time()
+
+                    # Query DUT release after successful provision
+                    if phase == TestPhase.PROVISION and not exit_code:
+                        device_info = job.get_device_info(rundir)
+                        if device_info and (
+                            device_ip := device_info.get("device_ip")
```
rene-oromtz: Device ip is nested inside device_info so this needs to be:

```suggestion
                            device_ip := device_info.get("device_info", {}).get("device_ip")
```

e.g:
`{"device_info": {"device_ip": "ip", "agent_name": "audino"}}`

agent/src/testflinger_agent/os_release.py
```diff
@@ -0,0 +1,128 @@
...
+    "-o",
+    "ConnectTimeout=10",
+    "-o",
+    "BatchMode=yes",
+    "-o",
+    "LogLevel=ERROR",
+]
+
+
+def query_dut_release(device_ip: str, username: str = "ubuntu") -> str:
```
rene-oromtz, line 40: imo, probably worth to add a note that explicitly says this is made via ssh key based auth that is handled at the agent host charm level.

rene-oromtz (changes requested):
Thank you for taking care of this! Overall this looks good to me! 
I took the liberty to test it on staging and found a bug regarding the `device_ip` retrieval. Once that is addressed it should be good as I can see the `identifier` label was properly added:
`phase_duration_seconds_sum{agent_id="audino", identifier="202409-35503", ...", test_phase="provision"}`

agent/src/testflinger_agent/os_release.py
ajzobro: Agreed, I came here looking for that when I saw that the agent host reaches IN.

have we considered using a test command to have the DUT hit an API on the Agent?

agent/src/testflinger_agent/os_release.py
```diff
@@ -0,0 +1,128 @@
...
+        key, _, value = line.partition("=")
+        value = value.strip('"')
+        result[key] = value
+    return result
+
+
+def _derive_release_label(os_info: dict) -> str:
+    """Convert parsed os-release data into a release label.
+
+    Follows C3 logic:
```
ajzobro: TF is open source do we want a ref to C3 code/logic?

agent/src/testflinger_agent/os_release.py
```diff
@@ -0,0 +1,128 @@
+# Copyright (C) 2026 Canonical
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License.
```
ajzobro, line 5:
```suggestion
# the Free Software Foundation, version 3 of the License.
```

agent/tests/test_os_release.py
```diff
@@ -0,0 +1,131 @@
+# Copyright (C) 2026 Canonical
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License.
```
ajzobro, line 5:
```suggestion
# the Free Software Foundation, version 3 of the License.
```

agent/src/testflinger_agent/os_release.py
ajzobro, line 5:
```suggestion
# the Free Software Foundation, version 3 of the License.
```

agent/src/testflinger_agent/agent.py
gntzio: Wow, Copilot tricked me: https://github.com/canonical/testflinger/pull/1009#discussion_r3056516363
Reverted to my initial version.

agent/src/testflinger_agent/os_release.py
gntzio: Added `SSH key-based auth` to the docsting.
Exposing an API on the Agent side and having the DUT call it would introduce additional technical challenges:
- The DUT would need to know the Agent's IP:PORT combination.
- Firewall rules for the lab pocket network and lab cloud environment would need to be adapted.

In general, as I understand it, the communication flow is designed so that the DUT does not communicate back to the TF Agent.

agent/src/testflinger_agent/os_release.py
gntzio: Makes sense. I have adapted the message, but the logic stays "as is".

agent/src/testflinger_agent/os_release.py
gntzio: It is copy/paste from other files with adjusted year.
The original version comes from the [COPYING](https://github.com/canonical/testflinger/blob/main/cli/COPYING#L639C5-L639C66).
However, `agent` folder does not have it.

rene-oromtz: Hmmmm I think the comment from Andy is because the policy specify to only use v3 of the GNU GPL license but as of now, there is no later version isn't?

We also need to review the license in all files, the `COPYING` contains all of the v3 GPL but we can probably just use:

```
    <one line to give the program's name and a brief idea of what it does.>
    Copyright (C) <year>  <name of author>
This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 3 of the License, or
    (at your option) any later version.

    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU General Public License for more details.

    You should have received a copy of the GNU General Public License
    along with this program.  If not, see <https://www.gnu.org/licenses/>
```

This is the recommendation from https://www.gnu.org/licenses/gpl-3.0.en.html

> It is safest to attach them to the start of each source file to most effectively state the exclusion of warranty; and each file should have at least the “copyright” line and a pointer to where the full notice is found

I think we can leave it as-is and move this discussion to its own Jira card so we can standardize all of the copyright notices across all files but want to know @ajzobro opinion on this.

rene-oromtz (approved):
Sweet! Working as intended:

Prometheus:
`phase_duration_seconds_sum{agent_id="audino", identifier="202409-35503", ..., release="24.04.4", test_phase="provision"}`

Agent logs:
```
[26-04-10 14:24:54]    INFO: (job.py:129)| Running setup_command: tf-setup
[26-04-10 14:24:59]    INFO: (job.py:129)| Running provision_command: tf-provision
[26-04-10 14:38:42]    INFO: (os_release.py:75)| DUT release detected: 24.04.4
```

There is just a pending discussion with @ajzobro so let's see what he thinks, besides that LGTM

ajzobro:
I won't hold this up.

gntzio:
Thanks to both of you for the review.

@ajzobro thanks for highlighting the issue with licensing - it is important to keep it consistent in the long term.

It is not a problem to address your concerns and adapt copyright headers in newly added files. However, this would make them inconsistent with the other source files in the `agent` directory. Also, given that COPYING file is not part of that directory, it could bring even more confusion. 

I would suggest to address the copyright topic for the entire `agent` component in a separate PR.

ajzobro (approved):

PR #1014, opened 2026-04-13
fix(submit): allow submit to be called by nested composite actions

rene-oromtz:
## Description
The current `github_ref` that is used so the `testflinger-setup` has the same reference as the submit action and it work in almost all but one edge case. 

Whenever calling the submit action from another composite action in "Repo B" which is then called from a third repo called "Repo C" and using a specified hash or tag, it will not honor the revision from which the submit action is called e.g.
repoC: uses `actionB@my-branch`
repoB: `actionB` uses `submit@submit/v1.4.4`

The `github_ref` will then be set to `my-branch` which does not exist in this repository causing a failure. 
To enable developers to test local changes on composite action, this PR changes to `github.action_path` to create a local copy of the `testflinger-setup`. This uses the same reference as the submit action without needing a `github_ref`

## Resolved issues
NA

## Tests
Tested on this run for backward compatibility calling this submit action from another workflow: https://github.com/canonical/tf-submit-action-test/actions/runs/24295950198/job/70940894955

Dev has also confirmed this approach works on this MM thread: https://chat.canonical.com/canonical/pl/w3g3pi4d5insfb5ndny7srzema

### DIFF
--- a/.github/actions/queue-status/action.yaml
+++ b/.github/actions/queue-status/action.yaml
@@ -35,7 +35,7 @@ runs:
   steps:
     # Use testflinger-setup by copying action from adjacent directory.
     # This is required so we can use the same reference as the one
-    # checkout by external workflow or action.
+    # checked out by external workflow or action.
     - name: Stage testflinger-setup from adjacent action directory
       shell: bash
       env:
--- a/.github/actions/queue-status/action.yaml
+++ b/.github/actions/queue-status/action.yaml
@@ -33,19 +33,17 @@ outputs:
 runs:
   using: composite
   steps:
-    # This action is necessary as a workaround for https://github.com/actions/runner/issues/895.
-    # Since contexts cannot be used in the `uses:` key, we can't guarantee that this action will
-    # use the same branch of testflinger-setup
-    - name: Checkout testflinger actions
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+    # Use testflinger-setup by copying action from adjacent directory.
+    # This is required so we can use the same reference as the one
+    # checkout by external workflow or action.
+    - name: Stage testflinger-setup from adjacent action directory
+      shell: bash
       env:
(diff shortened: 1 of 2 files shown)


.github/actions/queue-status/action.yaml
```diff
@@ -33,19 +33,17 @@ outputs:
...
   using: composite
   steps:
-    # This action is necessary as a workaround for https://github.com/actions/runner/issues/895.
-    # Since contexts cannot be used in the `uses:` key, we can't guarantee that this action will
-    # use the same branch of testflinger-setup
-    - name: Checkout testflinger actions
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+    # Use testflinger-setup by copying action from adjacent directory.
+    # This is required so we can use the same reference as the one
+    # checkout by external workflow or action.
```
ajzobro:
```suggestion
    # checked out by external workflow or action.
```

.github/actions/submit/action.yaml
```diff
@@ -39,19 +39,17 @@ runs:
...
   steps:
 
-    # This action is necessary as a workaround for https://github.com/actions/runner/issues/895.
-    # Since contexts cannot be used in the `uses:` key, we can't guarantee that this action will
-    # use the same branch of testflinger-setup
-    - name: Checkout testflinger actions
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+    # Use testflinger-setup by copying action from adjacent directory.
+    # This is required so we can use the same reference as the one
+    # checkout by external workflow or action.
```
ajzobro:
```suggestion
    # checked out by external workflow or action.
```

ajzobro (dismissed):

ajzobro:
Looks good to me, not sure if my understanding of the comment is right though.

rene-oromtz:
@ajzobro thanks for the review! I just pushed your suggestions regarding the comments

ajzobro (approved):
LGTM

PR #1015, opened 2026-04-13
feat(server): enable authentication to webhook endpoint

rene-oromtz:
## Description
This PR add support for providing authentication to webhook endpoint via Charm definition. 

1. Add a typed config to make sure charm configures the webhook endpoint properly. As a bonus, the typed config is also applicable for all the other fields. 
2. Add optional authentication via API Key (this is managed at deployment level) 
3. As payload is not user controlled, we are only validating that we are attempting to send to the same server configured webhook. API Key is also scoped on Test Observer side. 

This follows the pattern discussed and approved in [MM](https://chat.canonical.com/canonical/pl/k39iy9c6i7r5586nm1xffcsfpe) to let TF server be the source of auth with Test Observer

## Resolved issues
Resolves [CERTTF-985](https://warthogs.atlassian.net/browse/CERTTF-985)

## Web service API changes
@v1.post("/job/<job_id>/events") should remain backward compatibility on current production usage but now it is enforced that the specified webhook in job should match the one configured by server

## Tests
Added additional unit tests on charm and server. 

Tested on staging sending an authenticated request, events are visible from staging TO:
Screenshot from 2026-05-06 09-29-33

[CERTTF-985]: https://warthogs.atlassian.net/browse/CERTTF-985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/src/testflinger/api/v1.py
```diff
@@ -766,26 +768,83 @@ def agents_status_post(job_id, json_data):
...
+        parsed_server_url.scheme != parsed_job_url.scheme
+        or parsed_server_url.netloc != parsed_job_url.netloc
+    ):
+        abort(HTTPStatus.FORBIDDEN, message="Unauthorized webhook URL")
+
+    # Enforce that the job webhook path is a subpath of the base URL
+    server_path = posixpath.normpath(parsed_server_url.path)
+    job_path = posixpath.normpath(parsed_job_url.path)
+    if not job_path.startswith(f"{server_path}/"):
+        abort(HTTPStatus.FORBIDDEN, message="Unauthorized webhook URL")
```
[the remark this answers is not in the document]
rene-oromtz: No longer applicable, new approach does not trust user configured input as we are now sending request only to authorized URL with scoped permissions

server/src/testflinger/api/v1.py
```diff
@@ -749,7 +750,7 @@ def agents_provision_logs_post(agent_name, json_data):
 @v1.input(schemas.StatusUpdate, location="json")
 def agents_status_post(job_id, json_data):
     """Post status updates from the agent to the server to be forwarded
-    to TestObserver.
+    to the server-configured webhook url.
```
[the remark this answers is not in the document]
rene-oromtz: Test Observer is still the webhook url, v1 just changed for generalization. Besides, server/API.md is already deprecated and changes are now updated in OpenAPI schema

server/charm/src/config.py
```diff
@@ -0,0 +1,48 @@
...
+        return value
+
+    @pydantic.field_validator("webhook_url")
+    @classmethod
+    def validate_webhook_url(cls, value):
+        """Validate that webhook_url includes a protocol and no paths."""
+        parsed_webhook = urlparse(value)
+        if parsed_webhook.scheme not in {"http", "https"}:
+            raise ValueError(
+                "webhook_url must include protocol (http:// or https://)"
```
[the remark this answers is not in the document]
rene-oromtz: This is only configured at charm config level by an admin, tightening the validation is not needed here as we only care about validating the string has a proper scheme and format.

server/src/testflinger/api/v1.py
```diff
@@ -766,26 +767,82 @@ def agents_status_post(job_id, json_data):
...
+            # Auth failures can happen if server configuration
+            # is no longer valid.
+            if response.status_code in (
+                HTTPStatus.UNAUTHORIZED,
+                HTTPStatus.FORBIDDEN,
+            ):
+                abort(
+                    HTTPStatus.INTERNAL_SERVER_ERROR,
+                    message="Webhook authentication failed",
+                )
```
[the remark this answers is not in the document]
rene-oromtz: missing tests added in [5af7468](https://github.com/canonical/testflinger/pull/1015/commits/5af746847af9cb3620b72e684887cfbea9c64f17)

server/charm/src/config.py
```diff
@@ -0,0 +1,48 @@
...
+    https_proxy: str = ""
+    no_proxy: str = "localhost,127.0.0.1,::1"
+    webhook_url: str = "http://test-observer-api.local/"
+    webhook_auth: str = ""
+
+    @pydantic.field_validator("external_hostname")
+    @classmethod
+    def validate_external_hostname(cls, value):
+        """Validate that external_hostname does not include protocol."""
+        if value.startswith(("http://", "https://")):
```
ajzobro, line 31: the docstring talks about "protocol" in generic terms, but the check only checks for `https?://`
The webhook url includes the protocol.
We could opt to just split on `//` and take the right hand side rather than raise an exception.
We could balk at any protocol (e.g. ftp://) by only looking for `'://' in value`

server/charm/src/config.py
```diff
@@ -0,0 +1,48 @@
...
+
+from urllib.parse import urlparse
+
+import pydantic
+
+
+class TestflingerServerConfig(pydantic.BaseModel):
+    """Testflinger Server Charm configuration."""
+
+    external_hostname: str = "testflinger.local"
```
ajzobro, line 13: This note from the unit tests belongs (also) in the config:
`    # Protocol is defined by ingress controller`

server/charm/src/config.py
```diff
@@ -0,0 +1,48 @@
...
+        """Validate that webhook_url includes a protocol and no paths."""
+        parsed_webhook = urlparse(value)
+        if parsed_webhook.scheme not in {"http", "https"}:
+            raise ValueError(
+                "webhook_url must include protocol (http:// or https://)"
+            )
+        if not parsed_webhook.netloc:
+            raise ValueError("webhook_url must include a host")
+        if parsed_webhook.path not in {"", "/"}:
+            raise ValueError("webhook_url must not include a path")
```
ajzobro: Why is it such that the webhook must not include the path?  As an open source tool, we should give some consideration to the use of webhooks other than our own.

server/schemas/openapi.json
```diff
@@ -1453,7 +1453,7 @@
     },
     "/v1/job/{job_id}/events": {
       "post": {
-        "description": "to TestObserver.\n\nThe json sent to this endpoint may contain data such as the following:\n{\n\"agent_id\": \"<string>\",\n\"job_queue\": \"<string>\",\n\"job_status_webhook\": \"<URL as string>\",\n\"events\": [\n{\n\"event_name\": \"<string enum of events>\",\n\"timestamp\": \"<datetime>\",\n\"detail\": \"<string>\"\n},\n...\n]\n}",
+        "description": "to the server-configured webhook url.\n\nThe json sent to this endpoint may contain data such as the following:\n{\n\"agent_id\": \"<string>\",\n\"job_queue\": \"<string>\",\n\"job_status_webhook\": \"<URL as string>\",\n\"events\": [\n{\n\"event_name\": \"<string enum of events>\",\n\"timestamp\": \"<datetime>\",\n\"detail\": \"<string>\"\n},\n...\n]\n}\n\n:param job_id: UUID as a string for the job\n:param json_data: JSON data containing the status updates and webhook URL",
```
ajzobro, line 1456:
```suggestion
        "description": "to the server-configured webhook url.\n\nThe json sent to this endpoint may contain data such as the following:\n{\n\"agent_id\": \"<string>\",\n\"job_queue\": \"<string>\",\n\"job_status_webhook\": \"<URL as string>\",\n\"events\": [\n{\n\"event_name\": \"<string enum of events>\",\n\"timestamp\": \"<datetime>\",\n\"detail\": \"<string>\"\n},\n...\n]\n}\n\n:param job_id: UUID as a string for the job\n:param json_data: JSON data containing the status updates and webhook URI",
```
URI not URL, are these parts put together? the base server from the testflinger server, and the "path" from the Agent?

server/src/testflinger/api/v1.py
```diff
@@ -767,26 +768,82 @@ def agents_status_post(job_id, json_data):
...
+    :param job_id: UUID as a string for the job
+    :param json_data: JSON data containing the status updates and webhook URL
     """
-    _ = job_id
+    if not check_valid_uuid(job_id):
+        abort(HTTPStatus.BAD_REQUEST, message="Invalid job_id specified")
+    if not database.job_exists(job_id):
+        abort(HTTPStatus.NOT_FOUND, message="Job not found")
+
     request_json = json_data
```
ajzobro: I think we can avoid this unnecessary variable rename, don't you?

server/src/testflinger/api/v1.py
```diff
@@ -767,26 +768,82 @@ def agents_status_post(job_id, json_data):
...
+        # Abort if no webhook configured server-side
+        abort(
+            HTTPStatus.INTERNAL_SERVER_ERROR,
+            message="Webhook URL not configured",
+        )
+
+    # Parse both url webhooks, consider webhook_url as the only authorized
+    # endpoint to send requests to.
+    parsed_job_url = urlparse(job_webhook)
+    parsed_authorized_url = urlparse(webhook_url)
```
ajzobro: consider renaming the simple "webhook_url" from all the places that it's relevant to "authorized_webhook_url" or "authorized_webhook_server"

server/src/testflinger/api/v1.py
```diff
@@ -767,26 +768,82 @@ def agents_status_post(job_id, json_data):
...
+        or parsed_job_url.netloc != parsed_authorized_url.netloc
+    ):
+        abort(
+            HTTPStatus.FORBIDDEN,
+            message="Invalid job_status_webhook URL specified",
+        )
+
+    # Attempt to get optional authentication header from server configuration
+    auth_headers = {}
+    if webhook_auth := os.environ.get("WEBHOOK_AUTH"):
```
ajzobro, line 808: Busting out the `:=` operator I see!  This might be the first use in this codebase, do we want to use it?

server/src/testflinger/api/v1.py
```diff
@@ -767,26 +768,82 @@ def agents_status_post(job_id, json_data):
...
+
     try:
-        s = requests.Session()
-        s.mount(
-            "",
-            HTTPAdapter(
+        with requests.Session() as webhook_session:
+            retry_adapter = HTTPAdapter(
                 max_retries=Retry(
                     total=3,
```
ajzobro: Do we need the response? Do we need to block our own API server just to talk to another API server?
Could we add this to a queue and allow some other non-api-endpoint process to see this through and make sure that it gets to the webhook server?

server/src/testflinger/api/v1.py
```diff
@@ -767,26 +768,82 @@ def agents_status_post(job_id, json_data):
...
                     allowed_methods=frozenset(["PUT"]),
                     backoff_factor=1,
                 )
-            ),
-        )
-        response = s.put(webhook_url, json=request_json, timeout=3)
-        return response.text, response.status_code
+            )
+            webhook_session.mount("http://", retry_adapter)
+            webhook_session.mount("https://", retry_adapter)
```
ajzobro: We should not need to use both schemes under any circumstances because we have already confirmed that there is only one approved for use by the config file; use that here.

server/src/testflinger/api/v1.py
```diff
@@ -767,26 +768,82 @@ def agents_status_post(job_id, json_data):
...
+            )
+
+            # Auth failures can happen if webhook authentication
+            # is no longer valid or missing entirely.
+            if response.status_code in (
+                HTTPStatus.UNAUTHORIZED,
+                HTTPStatus.FORBIDDEN,
+            ):
+                abort(
+                    HTTPStatus.INTERNAL_SERVER_ERROR,
```
ajzobro, line 837: Need to consider: what does the agent calling the webhook do when it gets 500?  We want it to continue on with life and keep going -- webhook or not, don't we?

server/src/testflinger/api/v1.py
```diff
@@ -767,26 +768,82 @@ def agents_status_post(job_id, json_data):
...
+                    message="Webhook authentication failed",
+                )
+            return response.text, response.status_code
     except requests.exceptions.Timeout:
-        return "Webhook Timeout", 504
+        abort(HTTPStatus.GATEWAY_TIMEOUT, message="Webhook Timeout")
+    except requests.exceptions.RequestException:
+        abort(HTTPStatus.BAD_GATEWAY, message="Webhook unreachable")
```
ajzobro, line 844: While I recognize that we might want users of our API to be knowledgable about how their request fared, I also think that it's completely reasonable that our server and our agents continue to function in spite of how another third-party service is working at any given point in time.

server/src/testflinger/database.py
```diff
@@ -627,3 +627,14 @@ def add_job_results(job_id: str, json_data: dict):
...
+
+
+def job_exists(job_id: str) -> bool:
+    """Validate job exists on database.
+
+    :param job_id: The job ID to check.
+    :return: True if the job exists, False otherwise.
+    """
+    return (
+        mongo.db.jobs.find_one({"job_id": job_id}, {"_id": True}) is not None
```
ajzobro, line 639: I could have sworn we already had a "database find one job" function that we could have used in exactly the same way -- but nope, somehow we don't already have that.  Must have been on the web views -- those could possibly use this function now...

server/tests/test_v1.py
```diff
@@ -998,8 +999,187 @@ def test_agents_status_put(mongo_app, requests_mock):
...
+        "agent_id": "agent1",
+        "job_queue": "myjobqueue",
+        "job_status_webhook": webhook,
+        "events": [],
+    }
+    output = app.post(f"/v1/job/{job_id}/events", json=status_update_data)
+
+    # Testflinger server should return 500 error if the webhook server
+    # responds with an auth error, as this indicates a misconfiguration
+    # that needs to be fixed at charm level by an admin.
```
ajzobro, line 1180: Consider a valid setup where there is no webhook defined on the server OR on any agents.
Consider a valid setup where some agents are configured without a webhook (e.g. some agents don't do work that reports to the webhook)

ajzobro: and by "agent" I think I mean "job"

server/charm/src/config.py
rene-oromtz: I modified the docstrings to specify HTTP scheme. As this is for the URL, I think its more readable and expected a user will configure http or https when they shouldn't

server/charm/src/config.py
rene-oromtz: I added this note in the validator

server/schemas/openapi.json
rene-oromtz: the json data coming from the agents contains the full URL to where to send the events so its correct URL

server/src/testflinger/api/v1.py
rene-oromtz: renamed the relevant variables as authorized_webhook_url

server/src/testflinger/api/v1.py
rene-oromtz: If I remember correctly, not the first! And personally I don't see a problem and why not, its mostly use in the v1.py but I been seeing it a lot in other charm development

server/src/testflinger/api/v1.py
rene-oromtz: Not really... Its mostly UI information for Test Observer but its not a big deal if we "lose" it afaik. I think given that, I reduced the retries to just 1 so agent is "blocked" at most 6 secs as its the most simple solution and in the worst scenario, we just lose that event in the UI

server/src/testflinger/api/v1.py
rene-oromtz: Currently It continues after exhausting the retries. Agent by itself isresilient to this type of failures. With the reduced retries, this should now continue faster

server/src/testflinger/api/v1.py
rene-oromtz: Yes, both continue to function, I reduced the retries but still think the logging is relevant in case we eventually need to identify the type of faliures

server/src/testflinger/database.py
rene-oromtz: Yeah... I think eventually we will need to check what can be moved to the database module... There is so much that could be reused

server/tests/test_v1.py
rene-oromtz: Configuration ts still valid, the agents does not send reports to the webhook unless they are explicitly told to. Its just the event endpoint the one that will not "work"

server/charm/src/config.py
rene-oromtz: Its really not needed, I removed this check

server/src/testflinger/api/v1.py
```diff
@@ -776,24 +776,22 @@ def agents_status_post(job_id, json_data):
...
     # Webhook specified in the job definition
-    job_webhook = request_json.pop("job_status_webhook")
+    job_webhook = json_data.pop("job_status_webhook")
 
     # Webhook base URL configured on the server
-    webhook_url = os.environ.get("WEBHOOK_URL")
-    if not webhook_url:
+    authorized_webhook_url = os.environ.get("WEBHOOK_URL")
+    if not authorized_webhook_url:
         # Abort if no webhook configured server-side
```
ajzobro, line 785: how does this get from config.py into environ variables? why is it in both places?
Could be this: config.py of charm -> environ vars of server application

server/src/testflinger/api/v1.py
ajzobro, line 772: It seems to me like the uri should be given separately, outside of the "json_data"

server/src/testflinger/api/v1.py
rene-oromtz: the charm loads the config as environment variables here:
https://github.com/canonical/testflinger/blob/bc0a048b525aaed25a11ad9bdf173de330d247ee/server/charm/src/charm.py#L478-L500

The charm code is separate from the server code so its in fact needed on both sides (charm code and server code). Whenever deployed as part of the charm, the typed config is enforced at charm level.

server/src/testflinger/api/v1.py
rene-oromtz: I'm not against this, in fact, on that same line I would like to replace `job_status_webhook` for `artifact_id` so we can construct the full path ourselves but this is also breaking for the current usage. For now, given that we are looking for the less resistance path would you be okay on leaving as-is and we can address both changes it in a follow-up Jira ticket?

rene-oromtz: As discussed in MM, I created a [Jira card ](https://warthogs.atlassian.net/browse/CERTTF-1132) so we can keep this is improvement in the backlog

ajzobro (approved):

ajzobro:
LGTM, I believe that for now we have improved upon the operation and can consider further improvements as we have time to do performance tuning.

PR #1018, opened 2026-04-14
fix(cli): secret command get client_id from auth module

### LINKED ISSUE #1017: bug: cli secret write does not take logged client_id
Description
----
To build secret write path, the cli  requires the client_id to build the path. Whenever used cached credentials, the CLI is not taking into account the client_id from the auth module and expect client_id to be provided manually

Logs
----
```
testflinger-cli login
Successfully authenticated as user 'rene_stg'
```
```
testflinger-cli secret write /new/path 'test'
Error writing secret: Authentication is required
```

Expected Results
---
user is already logged via `testflinger-cli login`, secret write command should be aware of the logged user

rene-oromtz:
## Description
Minor bug fix for the secrets subcommands. Currently, they are taking into account the `client_id` that is provided via arguments or environment variables. 

The auth module has `auth.client_id` attribute that should be used instead. 
This PR also removes a trailing slash in case its provided so the path can be built correctly.

## Resolved issues
Resolves #1017 
Resolves [CERTTF-1019](https://warthogs.atlassian.net/browse/CERTTF-1019)

## Tests
All existing tests are passing. Updated secret write and delete test so it can also accept the path with leading slash

Tested on staging as well:
```
uv run testflinger-cli --server https://testflinger-staging.canonical.com login
Successfully authenticated as user 'rene_stg'
```
```
uv run testflinger-cli --server https://testflinger-staging.canonical.com secret write new/path 'test'
Secret 'new/path' written successfully
```

[CERTTF-1019]: https://warthogs.atlassian.net/browse/CERTTF-1019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -580,7 +580,9 @@ class TestflingerCli:
             "write", help="Write a secret value"
         )
         write_parser.set_defaults(func=self.secret_write)
-        write_parser.add_argument("path", help="Path for the secret")
+        write_parser.add_argument(
+            "path", help="Path for the secret", type=helpers.regex_path
+        )
         write_parser.add_argument("value", help="Value of the secret")
         self._add_auth_args(write_parser)
 
@@ -590,7 +592,9 @@ class TestflingerCli:
         )
         delete_parser.set_defaults(func=self.secret_delete)
         delete_parser.add_argument(
-            "path", help="Path for the secret to delete"
+            "path",
+            help="Path for the secret to delete",
+            type=helpers.regex_path,
         )
         self._add_auth_args(delete_parser)
 
@@ -1843,9 +1847,7 @@ class TestflingerCli:
 
         secret_data = {"value": self.args.value}
 
-        # Remove leading slash if present to build the correct endpoint URL
-        path = self.args.path.lstrip("/")
-        endpoint = f"/v1/secrets/{self.auth.client_id}/{path}"
+        endpoint = f"/v1/secrets/{self.auth.client_id}/{self.args.path}"
         try:
             self.client.put(endpoint, secret_data, headers=auth_headers)
         except client.HTTPError as exc:
@@ -1862,9 +1864,7 @@ class TestflingerCli:
         if auth_headers is None or self.auth.client_id is None:
             sys.exit("Error deleting secret: Authentication is required")
 
-        # Remove leading slash if present to build the correct endpoint URL
-        path = self.args.path.lstrip("/")
-        endpoint = f"/v1/secrets/{self.auth.client_id}/{path}"
+        endpoint = f"/v1/secrets/{self.auth.client_id}/{self.args.path}"
         try:
             self.client.delete(endpoint, headers=auth_headers)
         except client.HTTPError as exc:
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -1838,11 +1838,14 @@ class TestflingerCli:
         except CredentialsError as exc:
             sys.exit(exc)
 
-        if auth_headers is None or self.client_id is None:
+        if auth_headers is None or self.auth.client_id is None:
             sys.exit("Error writing secret: Authentication is required")
 
         secret_data = {"value": self.args.value}
-        endpoint = f"/v1/secrets/{self.client_id}/{self.args.path}"
+
+        # Remove leading slash if present to build the correct endpoint URL
+        path = self.args.path.lstrip("/")
+        endpoint = f"/v1/secrets/{self.auth.client_id}/{path}"
         try:
             self.client.put(endpoint, secret_data, headers=auth_headers)
         except client.HTTPError as exc:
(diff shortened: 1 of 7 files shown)


cli/testflinger_cli/__init__.py
```diff
@@ -1392,11 +1392,14 @@ def secret_write(self):
...
 
-        if auth_headers is None or self.client_id is None:
+        if auth_headers is None or self.auth.client_id is None:
             sys.exit("Error writing secret: Authentication is required")
 
         secret_data = {"value": self.args.value}
-        endpoint = f"/v1/secrets/{self.client_id}/{self.args.path}"
+
+        # Remove leading slash if present to build the correct endpoint URL
+        path = self.args.path.lstrip("/")
```
ajzobro: So we need to ensure that there is no preceding slash...

What about the trailing slash?
 - value///////////

What if the user's value is really badly formed?
 - //////////value

cli/testflinger_cli/__init__.py
```diff
@@ -1410,10 +1413,12 @@ def secret_delete(self):
         except CredentialsError as exc:
             sys.exit(exc)
 
-        if auth_headers is None or self.client_id is None:
+        if auth_headers is None or self.auth.client_id is None:
             sys.exit("Error deleting secret: Authentication is required")
 
-        endpoint = f"/v1/secrets/{self.client_id}/{self.args.path}"
+        # Remove leading slash if present to build the correct endpoint URL
+        path = self.args.path.lstrip("/")
```
ajzobro: see above.

We should also aggressively test some really bad values.

cli/tests/test_secret.py
```diff
@@ -30,18 +30,15 @@
 
 @pytest.mark.parametrize(
     "test_path",
-    [
-        "path",
-        "my/secret/path",
-    ],
+    ["path", "my/secret/path", "/my/secret/path"],
```
ajzobro: see above, increase the robustness of the bad paths.

- ///value
- /val/ue/
- value///

ajzobro (changes requested):
There are some other considerations that should also be made.  I think we need to ensure that whatever solution we are providing as the "Testflinger", and not as the vault behind testflinger, should be well considered such that we (testflinger) are not simply exposing the vault's own limitations.

And by that I mean that we should feel free to impose our own rules on the names (paths?) of the secrets such that they will be able to be handled in a fully predictable way.

I am more in favor of rejecting a secret's creation than I am over massaging it.

TL;DR: Don't fix it, just reject it.


Secret 'new/path///////////////////////////////////////////////////////////////////////////////////////////////////////////////a' written successfully

cli/testflinger_cli/__init__.py
rene-oromtz: The preceding slash is required while creating the endpoint, so the cli code does need to take care of it not including any preceding slash. I do agree that the server must reject any other malformed endpoint, I can see on making this more strict or even maybe normalizing it on CLI side. I'm also oaky on plain reject on CLI + additional strict control server side (for API usage)

rene-oromtz: Nvm! It is only required to check for malformed paths on CLI before forming endpoint! The schema in the server is for the test_data. But the secret is "stored" in the path defined by user.
Thanks for noticing this! I'll take care of it

rene-oromtz:
@ajzobro thanks for pointing this out! Definitively needed a stricter check on CLI side. I took some general recommendation for vault path naming and add a validation at argparse level. I also added a bunch of scenarios to validate malformed and valid path so please let me know if there is any other scenario I should consider to test against!

cli/testflinger_cli/helpers.py
```diff
@@ -13,6 +13,11 @@
 from testflinger_cli.consts import SNAP_NAME, SNAP_PRIVATE_DIRS
 from testflinger_cli.errors import SnapPrivateFileError
 
+# Only accept paths that are separated by forward slashes
+# Expected format should not start or end with a slash
+# Examples of valid paths: "path", "my/secret/path"
+PATH_PATTERN = re.compile(r"^[a-zA-Z0-9_-]+(?:\/[a-zA-Z0-9_-]+)*$")
```
ajzobro, line 19: although we allow _ and -, we do not test those characters in our tests.

cli/testflinger_cli/helpers.py
```diff
@@ -233,3 +238,18 @@ def regex_arg(value):
...
+    """Validate that a file path argument is a valid regex pattern.
+
+    :param value: The file path string to validate
+    :param pattern: The regex pattern to validate against
+    :return: The original value if valid
+    :raises ArgumentTypeError: If the value does not match the regex pattern
+    """
+    if not pattern.match(value):
+        raise argparse.ArgumentTypeError(
+            f"Invalid value '{value}', not a valid path"
```
ajzobro: We should give some advice here as to what the acceptable pattern is.

Valid paths:
```
path/a/b/a/a3/_/-/--------__-__---_--_-__---_-__----_----_-/a________B_______________CV______-----____-__-_-__-/-_/_-/--/__/aaAA/a
9
0
-0
--00/__00
```

ajzobro (changes requested):
Only thing I really think we need to add is a better error message.

cli/testflinger_cli/helpers.py
rene-oromtz: Added a more explicit error message as suggested

cli/testflinger_cli/helpers.py
rene-oromtz: Just added a test for underscore, hyphen was already included

rene-oromtz:
@ajzobro added a more explicit error message for invalid paths. Please let me know if something else is missing!

ajzobro (dismissed):
LGTM, thanks for taking the suggestions!

ajzobro (approved):
Good addition, glad we cleaned that up.

PR #1019, opened 2026-04-15
feat(cli): add ephemeral to reserve command

rene-oromtz:
## Description
With ephemeral now supported server side. Adding support for this type of deployments via `reserve` command.

## Resolved issues
Resolves [CERTTF-965](https://warthogs.atlassian.net/browse/CERTTF-965)

## Documentation
Updated reserve docs with a tip for maas deployments and ephemeral flag. 
Also added doc for ephemeral support in maas device connector reference

## Web service API changes
NA

## Tests
Added unit tests and tested in staging:
```
uv run testflinger-cli --server https://testflinger-staging.canonical.com reserve --distro noble --key lp:rene-orozco --timeout 60 --queue audino --ephemeral

The following job will be submitted:
{
...
  "provision_data": {
    "distro": "noble",
    "ephemeral": true
  }
}
```

[CERTTF-965]: https://warthogs.atlassian.net/browse/CERTTF-965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -352,6 +352,11 @@ class TestflingerCli:
             help="Reservation timeout in seconds (default "
             f"{DEFAULT_RESERVE_TIMEOUT})",
         )
+        parser.add_argument(
+            "--ephemeral",
+            action="store_true",
+            help="MAAS ephemeral (in-memory) deployment.",
+        )
 
     def _add_status_args(self, subparsers):
         """Command line arguments for status."""
@@ -1398,7 +1403,14 @@ class TestflingerCli:
         # Handle distro if provided
         provision_data = {}
         if self.args.distro:
-            provision_data = {"provision_data": {"distro": self.args.distro}}
+            # If distro is provided, we can assume this is a MAAS deployment
+            # Setting ephemeral based on user input.
+            provision_data = {
+                "provision_data": {
+                    "distro": self.args.distro,
+                    "ephemeral": self.args.ephemeral,
+                }
+            }
         else:
             try:
                 images = self.client.get_images(queue)
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -1400,16 +1400,21 @@ class TestflingerCli:
             self.do_list_queues()
         )
 
+        # Ephemeral only supported for MAAS deployments via distro argument
+        if self.args.ephemeral and not self.args.distro:
+            sys.exit(
+                "Error: --ephemeral option requires --distro to be specified."
+            )
+
         # Handle distro if provided
         provision_data = {}
         if self.args.distro:
-            # If distro is provided, we can assume this is a MAAS deployment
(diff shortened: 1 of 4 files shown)


cli/testflinger_cli/__init__.py
```diff
@@ -1395,10 +1400,22 @@ def reserve(self):
             self.do_list_queues()
         )
 
+        # Ephemeral only supported for MAAS deployments via distro argument
+        if self.args.ephemeral and not self.args.distro:
+            sys.exit(
+                "Error: --ephemeral option requires --distro to be specified."
```
ajzobro, line 1406: Probably ok.  I don't have a better solution, but I did want to comment that `--distro` (today) is synonymous with `maas` and so then `--ephemeral` and `--distro` do align in applicability; but that might not always be true (e.g. if we allow for distro to be specified in the future for other provision methods, those provision methods may still not support `--ephemeral`)  The *right* solution, would be to consider the queue and check against the provision method(s) supported by that queue -- which is a whole other problem already known and unsolved.  Note: this same problem already exists with respect to `--distro` -- just because you specify that you want a `--distro` doesn't mean that it will actually work with all the agent workers on that queue -- and today, we won't know until the agent claims the job and fails.

docs/reference/device-connector-types.rst
```diff
@@ -107,6 +107,12 @@ The ``maas2`` device connector supports the following ``provision_data`` keys:
    * - ``disks``
      - Specify a custom disk configuration for the machine. For more information, see the
        :doc:`maas_storage`.
+   * - ``ephemeral``
+     - Set to ``true`` to deploy in-memory rather than to disk. Ephemeral deployment
+       is supported only with MAAS 3.5.0 and later; on older MAAS versions this
+       setting is ignored. For more information, see `MAAS documentation:
+       Ephemeral deployment
+       <https://canonical.com/maas/docs/about-deploying-machines#p-17464-ephemeral-os-deployments-maas-35>`_.
```
ajzobro, line 115: Looks good.  Do we know if this link is a permalink or might it change in the future?

`#p-17464-ephemeral-os-deployments-maas-35`

ajzobro (approved):
LGTM other than the comments, but I don't actually know that there is action to take with this one.

docs/reference/device-connector-types.rst
rene-oromtz: To be honest, I don't know... But MAAS docs seems stable enough, in our docs there are other links following the same structure and they been there for a long time

cli/testflinger_cli/__init__.py
rene-oromtz: Yes, preferably I would want to use a single key key for all the deployments and server do a rejection based on provision type but we are not there yet (hopefully we get there soon). For now at least it will be clear that this is just for MAAS so users knows they are trying a MAAS deployment

PR #1025, opened 2026-04-21
Add always to the job completion waiting

aglinserer:
## Description

When using the testflinger submit GH action, if the provision or test step fail, the whole action is ended. This can lead to following actions not being able to fetch the status since testflinger results are not yet available.

In our case this leads to false messages in steps after the submit (pinging people etc).
In this [CI run](https://github.com/canonical/cypress-image-definitions/actions/runs/24711551479/job/72277224149) you see that no polling is happening and another action in this workflow now reports "Test job still running or results unavailable".
The [same workflow](https://github.com/canonical/cypress-image-definitions/actions/runs/24716551232/job/72294379117) but running with an `always` added to the completion waiting reports a failure correctly.

## Resolved issues

No issue has been created.

## Documentation

Only addition is a `always()` to the wait for completion step.

## Web service API changes

None.

## Tests

Fork has been executed, see above.

### DIFF
--- a/.github/actions/submit/action.yaml
+++ b/.github/actions/submit/action.yaml
@@ -239,7 +239,7 @@ runs:
         exit $EXIT_STATUS
 
     - name: Wait for job completion
-      if: inputs.poll == 'true' && inputs.dry-run != 'true' && steps.check-phases.outputs.reserve == 'false'
+      if: always() && !cancelled() && inputs.poll == 'true' && inputs.dry-run != 'true' && steps.check-phases.outputs.reserve == 'false'
       shell: bash
       env:
         SERVER: https://${{ inputs.server }}


rene-oromtz:
@aglinserer thanks a lot for jumping in to fix the issue! Only thing that I'm worried about is that I'm not sure if given the wait for job completion will run even when the action is cancelled, it will conflict with the `Cancel the job if the action is cancelled` which will not run as a previous step will still be doing a while loop. I'm not sure about this but probably worth to double check. 

One other alternative is that realistically, we should exit if provision and reserve fails but test phase failures are expected, maybe we can `continue-on-error`
```
    - name: Track test phase
      id: track-test
      continue-on-error: true
```

We may need to add another step right after wait for job completion to report the test failure to the action but at least this should be clean and the wait for job completion will always run regardless of test failures. Please let me know what you think!

fhost (approved):
LGTM

aglinserer:
A cancelled run showing it works when cancelling during the provision phase: https://github.com/canonical/cypress-image-definitions/actions/runs/24730188636/job/72342171412
Will get back with a cancel during testing as well

rene-oromtz (approved):
Thanks for adding the test! LGTM, using cancelled() to prevent other temporary failures on other phases as discussed via MM

PR #1026, opened 2026-04-22
new: allow providing the image over attachment for IoT

p-gentili:
## Description

The server side now allows the user to provide the boot image as at attachment, useful when there's no accessible URL hosting the artifact.

## Resolved issues

Part of https://warthogs.atlassian.net/browse/ZAP-1436

## Documentation

On server side

## Web service API changes

N/A

## Tests

Updated.

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
@@ -181,18 +181,12 @@ class ZapperConnector(ABC, DefaultDevice):
 
     @staticmethod
     def _find_provision_attachment() -> Optional[Path]:
-        """Return the provision attachment if one was uploaded, else None.
-
-        The server schema enforces at most one provision attachment per
-        Zapper IoT job, so the first regular file in the directory is
-        unambiguously the boot binary.
-        """
+        """Return the provision attachment if one was uploaded, else None."""
         attachments_dir = Path.cwd() / ATTACHMENTS_DIR / "provision"
-        if not attachments_dir.is_dir():
+        try:
+            return next(p for p in attachments_dir.iterdir() if p.is_file())
+        except (FileNotFoundError, NotADirectoryError, StopIteration):
             return None
-        return next(
-            (p for p in attachments_dir.iterdir() if p.is_file()), None
-        )
 
     def _run(self, *args, **kwargs):
         """Run the Zapper provisioning via the REST API.
--- a/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
@@ -35,6 +35,9 @@ from testflinger_device_connectors.devices import (
 
 logger = logging.getLogger(__name__)
 
+# should mirror `testflinger_agent.config.ATTACHMENTS_DIR`
+ATTACHMENTS_DIR = "attachments"
+
 
 class ZapperConnector(ABC, DefaultDevice):
     """Abstract base class defining a common interface for Zapper-driven
@@ -45,7 +48,6 @@ class ZapperConnector(ABC, DefaultDevice):
     ZAPPER_CONNECTION_TIMEOUT = 30
     ZAPPER_READ_TIMEOUT = 60 * 90
     ZAPPER_REST_PORT = 8000
-    _boot_binary: Optional[Path] = None
 
     def _api_post(self, endpoint: str, **kwargs) -> requests.Response:
         """Send a POST request to the Zapper REST API.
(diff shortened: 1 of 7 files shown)


device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
```diff
@@ -175,11 +179,22 @@ def _validate_configuration(
...
         raise NotImplementedError
 
+    @staticmethod
+    def _find_provision_attachment() -> Optional[Path]:
+        """Return the provision attachment if one was uploaded, else None."""
+        attachments_dir = Path.cwd() / ATTACHMENTS_DIR / "provision"
+        if not attachments_dir.is_dir():
+            return None
+        files = sorted(p for p in attachments_dir.iterdir() if p.is_file())
+        return files[0] if files else None
```
ajzobro: I don't think that taking the first file alphabetically is a very robust solution.  If we intend to keep it this way, can you include commentary as to why this choice was made?

server/src/testflinger/api/schemas.py
```diff
@@ -173,20 +173,31 @@ class ZapperIoTPresetProvisionData(BaseZapperProvisionData):
...
     preset = fields.String(required=True)
     preset_kwargs = fields.Dict(required=False)
 
     @validates_schema
-    def validate_url_or_urls(self, data, **_):
-        """Validate that exactly one of `url` or `urls` is provided."""
-        has_url = "url" in data
-        has_urls = "urls" in data
-        if has_url and has_urls:
+    def validate_image_source(self, data, **_):
```
ajzobro: if this is a "boot" image, could we consider adding the classifier onto the variable name so as not to confuse it with other "image" contexts.

server/src/testflinger/api/schemas.py
```diff
@@ -173,20 +173,31 @@ class ZapperIoTPresetProvisionData(BaseZapperProvisionData):
...
+        """Validate that at most one URL source is provided.
```
ajzobro:
```suggestion
        """Validate that at most one boot image source is provided.
```

server/src/testflinger/api/schemas.py
```diff
@@ -173,20 +173,31 @@ class ZapperIoTPresetProvisionData(BaseZapperProvisionData):
...
+        if "url" in data and "urls" in data:
             raise ValidationError(
                 "Provide only one of 'url' or 'urls', not both."
             )
-        if not has_url and not has_urls:
-            raise ValidationError("Either 'url' or 'urls' must be provided.")
+        if (
+            "url" not in data
+            and "urls" not in data
+            and not data.get("attachments")
```
ajzobro: we seem to be missing an error message in the case where url(s) and an attachment have been provided.  It seems reasonable (based on the comments above) that if the attachment will be used to the exclusion of the urls that if both a boot image provisioning  attachment and a url/urls are provided that we should respond with an appropriate error rather than silently discard or otherwise not use the url/urls value(s).

server/src/testflinger/api/schemas.py
```diff
@@ -173,20 +173,31 @@ class ZapperIoTPresetProvisionData(BaseZapperProvisionData):
...
+        ):
+            raise ValidationError(
+                "Either 'url'/'urls' or a provision attachment must "
+                "be provided."
+            )
 
 
 class ZapperIoTCustomProvisionData(BaseZapperProvisionData):
```
ajzobro: Since both of these leverage the Base -- is it really the base that should be getting these new rules?  Or should we have a ZapperIoT*ProvisionData class?

class ZapperIoTCustomProvisionData(BaseZapperProvisionData):
class ZapperIoTPresetProvisionData(BaseZapperProvisionData):

I have the same comments for both classes.

server/tests/test_v1.py
```diff
@@ -387,6 +387,39 @@ def test_add_job_zapper_iot_custom_provision_data_single_url(mongo_app):
...
+    job_data = {"job_queue": "test", "provision_data": provision_data}
+    app, _ = mongo_app
+    output = app.post("/v1/job", json=job_data)
+    assert HTTPStatus.OK == output.status_code
+
+
+def test_add_job_zapper_iot_custom_provision_data_attachment(mongo_app):
+    """Test that a job with zapper_iot_custom using an attachment works."""
+    provision_data = {
+        "attachments": [{"agent": "provision/image.img"}],
```
ajzobro, line 405: see comments above about picking the first file alphabetically -- can we at least only look at *.img files if that's the standard?

server/tests/test_v1.py
```diff
@@ -387,6 +387,39 @@ def test_add_job_zapper_iot_custom_provision_data_single_url(mongo_app):
...
+        "provision_plan": {},
+    }
+    job_data = {"job_queue": "test", "provision_data": provision_data}
+    app, _ = mongo_app
+    output = app.post("/v1/job", json=job_data)
+    assert HTTPStatus.OK == output.status_code
+
+
```
ajzobro, line 413: Add more tests or combinations of mutually exclusive things (url/urls/attachments)

ajzobro (commented):
I know it's in draft, but it caught my eye :-)

device-connectors/src/testflinger_device_connectors/devices/zapper/__init__.py
p-gentili: Indeed very bad approach :) We expect 0 or 1 attachments, added a proper check to the schema.

server/src/testflinger/api/schemas.py
p-gentili: Done, thanks!

server/tests/test_v1.py
p-gentili: I don't want to enforce any format, or compression algorithm. Now I'm checking for number of attachments in the schema.

p-gentili:
> I know it's in draft, but it caught my eye :-)

Thanks for all the comments! Yeah, I usually provide a draft PR here to facilitate the job of the server side reviewer (we use this package as *the* client). I'll leave it as draft until the other side is merged, but it should already look a bit better.

ajzobro (approved):
OK. Agree with comments.

PR #1027, opened 2026-04-22
Fix: TF Agent host charm rebuilds COS scrape jobs based on supervisor config.

gntzio:
## Description

After deploying the Grafana Agent hosts in TEL, the `/etc/grafana-agent.yaml` config was not properly populated with scrape endpoints.

Following investigation, the suspicion is that since `charm.py` is instantiated on every Juju event, the in-memory `self.scrape_endpoints` variable is not consistently populated.

To address this, `COSAgentProvider.scrape_configs` are now sourced directly from the `supervisord` configuration.

## Resolved issues

[CERTTF-1038](https://warthogs.atlassian.net/browse/CERTTF-1038)

## Documentation

Charm unit tests.

## Web service API changes

N/A

## Tests

Charm unit tests.

[CERTTF-1038]: https://warthogs.atlassian.net/browse/CERTTF-1038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/charms/testflinger-agent-host-charm/src/supervisord.py
+++ b/agent/charms/testflinger-agent-host-charm/src/supervisord.py
@@ -64,6 +64,24 @@ def get_supervisor_agents_port_mapping() -> dict[str, int]:
     return agent_port_mapping
 
 
+def get_supervisor_scrape_jobs() -> list[dict]:
+    """Build scrape jobs from persisted supervisor configs.
+
+    Returns:
+        list[dict]: COS-compatible scrape jobs for each configured agent.
+    """
+    return [
+        {
+            "job_name": agent_name,
+            "metrics_path": "/metrics",
+            "static_configs": [{"targets": [f"localhost:{port}"]}],
+        }
+        for agent_name, port in sorted(
+            get_supervisor_agents_port_mapping().items()
+        )
+    ]
+
+
 def parse_supervisor_config_file(
     conf_path: str,
 ) -> Tuple[Optional[str], Optional[int]]:
--- a/agent/charms/testflinger-agent-host-charm/src/charm.py
+++ b/agent/charms/testflinger-agent-host-charm/src/charm.py
@@ -51,10 +51,14 @@ class TestflingerAgentHostCharm(ops.charm.CharmBase):
             self.on.update_testflinger_action,
             self.on_update_testflinger_action,
         )
-        self.scrape_jobs = []
         self._grafana_agent = COSAgentProvider(
             self,
             scrape_configs=self.get_scrape_jobs,
+            refresh_events=[
+                self.on.config_changed,
+                self.on.update_configs_action,
+                self.on.upgrade_charm,
+            ],
         )
         self.framework.observe(self.on.secret_changed, self._on_secret_changed)
         self.framework.observe(self.on.update_status, self._on_update_status)
@@ -142,7 +146,6 @@ class TestflingerAgentHostCharm(ops.charm.CharmBase):
         ) as service_template:
             template = Template(service_template.read())
 
-        self.scrape_jobs = []
         for agent_dir in agent_dirs:
             agent_config_path = agent_dir
             agent_name = agent_dir.name
@@ -162,15 +165,6 @@ class TestflingerAgentHostCharm(ops.charm.CharmBase):
                 agent_config_path=agent_config_path,
                 virtual_env_path=VIRTUAL_ENV_PATH,
                 metric_endpoint_port=metric_endpoint_port,
-            )
-            self.scrape_jobs.append(
-                {
-                    "job_name": agent_name,
-                    "metrics_path": "/metrics",
-                    "static_configs": [
-                        {"targets": [f"localhost:{metric_endpoint_port}"]}
-                    ],
-                }
             )
 
             with open(
@@ -324,7 +318,7 @@ class TestflingerAgentHostCharm(ops.charm.CharmBase):
         self.unit.status = ops.ActiveStatus()
 
     def get_scrape_jobs(self):
-        return self.scrape_jobs
+        return supervisord.get_supervisor_scrape_jobs()
 
 
 if __name__ == "__main__":
--- a/agent/charms/testflinger-agent-host-charm/tests/unit/test_charm.py
+++ b/agent/charms/testflinger-agent-host-charm/tests/unit/test_charm.py
@@ -50,6 +50,47 @@ def test_update_tf_cmd_scripts_on_config_changed(
     mock_update_scripts.assert_called_once()
 
 
+@patch("charm.COSAgentProvider")
+@patch("charm.supervisord.get_supervisor_scrape_jobs")
+def test_cos_provider_reads_scrape_jobs_from_supervisor_configs(
+    mock_get_scrape_jobs,
+    mock_cos_provider,
+    ctx,
+    state_in,
+):
+    """Test scrape job publication uses persisted supervisor configs."""
+    expected_scrape_jobs = [
+        {
+            "job_name": "agent1",
+            "metrics_path": "/metrics",
+            "static_configs": [{"targets": ["localhost:8000"]}],
+        }
+    ]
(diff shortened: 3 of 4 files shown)


agent/charms/testflinger-agent-host-charm/src/supervisord.py
ajzobro, line 37:
```suggestion
        for conf_file in [c for supervisor_conf_dir.iterdir() if c.suffix == ".conf"]:
```
Consider moving the if block up, I think it make is more obvious that the loop is only considering .config files.  Alternatively use glob.

ajzobro, line 38:
```suggestion
```
See above, would also change indentation on lines 39-69

rene-oromtz (approved):
LGTM, as far as I remember, Grafana Agent required the relation with Prometheus to actually configure the scrape jobs. 

Due to initial TEL deployment without Prometheus, that relation data may never got populated from our charm side... I think the key is actually the `refresh_events` addition which also makes sense because when we add new agents, we should also be refreshing the scrape_jobs with the new agents config.

agent/charms/testflinger-agent-host-charm/src/supervisord.py
ajzobro, line 42: What if the config doesn't have an agent name and port -- it seems that we will just stkoip them  Is that OK or should that be an error condition?

ajzobro, line 49: The OSError and IOError tracebacks that mght have been produced by parse_supervisor_config_file are caught within parse_supervisor_config_file.

Please consider adding tests for the failure conditions and new code.

ajzobro, line 54: The ValueError that parse_supervisor_config_file is already handled within parse_supervisor_config_file and causes it to return None, None

Same for the

agent/charms/testflinger-agent-host-charm/src/supervisord.py
```diff
@@ -64,6 +64,24 @@ def get_supervisor_agents_port_mapping() -> dict[str, int]:
...
+    """
+    return [
+        {
+            "job_name": agent_name,
+            "metrics_path": "/metrics",
+            "static_configs": [{"targets": [f"localhost:{port}"]}],
+        }
+        for agent_name, port in sorted(
+            get_supervisor_agents_port_mapping().items()
+        )
```
ajzobro, line 81:
```suggestion
        ) if agent_name
```

Do we want to include "None" in this list?

Need to test behavior if the value were to come back None anyway

ajzobro (changes requested):

agent/charms/testflinger-agent-host-charm/src/supervisord.py
rene-oromtz: As most of these comments are not related to the code change, do you think we should open a Jira card to track this improvements you already identified? This may be better suited to address in a followup PR if they are not directly impacting the code changes.

agent/charms/testflinger-agent-host-charm/src/supervisord.py
gntzio: I would leave the logic "as is" to avoid semantic complexity. 
`get_supervisor_agents_port_mapping` should not by design return `(None, None)`.
In other case, other consumers should perform the same check (e.g.in `charm.py`).

gntzio:
@ajzobro, thanks for pointing out potential improvements.
As far as I can see, they are (mostly) not related to the changes in this PR itself. 
I would appreciate it, if these could be addressed in a separate "improvements" PR.

agent/charms/testflinger-agent-host-charm/src/supervisord.py
rene-oromtz: Can it be None? It defaults to empty dict {} and only inserts to the agent_port_mapping if agent and port was found.

agent/charms/testflinger-agent-host-charm/src/supervisord.py
rene-oromtz: I went ahead and open [CERTTF-1041](https://warthogs.atlassian.net/browse/CERTTF-1041) please let me know what you think!

agent/charms/testflinger-agent-host-charm/src/supervisord.py
ajzobro: If "by design" you mean that every agent will always have a config file, that config file can be read and the values within the config file for agent name and port will always be defined: then why do we have any error handling in parse_supervisor_config_file at all?   Most of the return statetments return None, None -- if we're going to make the argument that by design we never end up in any of those failure paths, then why do they exist?

This new code that you wrote is definitely in scope, and does not handle all possible returns from parse_supervisor_config_file in a way that doesn't just pass along the problem.

ajzobro (changes requested):
Please address the issues that can easily be addressed within the scope of the code that was introduced here, including tests for the failure conditions.

CERTTF-1041 now eixsts as a strategy for handling (kicking along) the technical debt this time.

agent/charms/testflinger-agent-host-charm/src/supervisord.py
gntzio: `parse_supervisor_config_file()` can return `(None, None)`, but `get_supervisor_scrape_jobs()` does not consume that output directly. `get_supervisor_agents_port_mapping()` already filters invalid parser results and returns only valid agent -> port entries, as stated in its docstring.

agent/charms/testflinger-agent-host-charm/src/supervisord.py
ajzobro, line 48: we can probably break after this unless we want to continue processing files and collect all the egent configs and take the last defined port for any config file for this agent.

agent/charms/testflinger-agent-host-charm/src/supervisord.py
ajzobro: > Can it be None? It defaults to empty dict {} and only inserts to the agent_port_mapping if agent and port was found.

Indeed, `get_supervisor_agents_port_mapping` cleans that up and returns an empty dict; so this will just crash.

>>> [{'a':'b','c':x,'d':y} for x, y in sorted( [ {} ])]
  File "<stdin>", line 1, in <module>
ValueError: not enough values to unpack (expected 2, got 0)
>>>

ajzobro (commented):
If we write the tests for the error cases we will clearly see the issues in the code.

agent/charms/testflinger-agent-host-charm/src/supervisord.py
gntzio: The example is not equivalent to this code path: we iterate over `sorted(get_supervisor_agents_port_mapping().items())`. 
So the equivalent expression would be `[{'a': 'b', 'c': x, 'd': y} for x, y in sorted({}.items())]`, which evaluates to `[]`.

rene-oromtz: That is failing because of this `[ {} ])` . you should be using: `[{'a':'b','c':x,'d':y} for x, y in sorted([])]` or you can mock this directly:
```
>>> agent_port_mapping = {}
>>> scrape_config = [
...         {
...             "job_name": agent_name,
...             "metrics_path": "/metrics",
...             "static_configs": [{"targets": [f"localhost:{port}"]}],
...         }
...         for agent_name, port in sorted(
...             agent_port_mapping.items()
...         )
... ]
>>> 
>>> print(scrape_config)
[]
```
Ah Gena was faster in replying

ajzobro:
I guess I'm not getting any more tests, huh?

ajzobro (approved):
Agree with creation of 1041 for improved testing and fixing the relied-upon-but-out-of-scope functions.

rene-oromtz:
Thanks @ajzobro! Agree that there is lots to improve. At least now you already identify what needs to be done it should be a bit faster to implement in 1041

PR #1030, opened 2026-04-24
fix: resolve mypy type errors in cli/testflinger_cli/auth.py

### LINKED ISSUE #1028: fix: type errors in cli/testflinger_cli/auth.py detected by mypy
Running mypy on `cli/testflinger_cli/auth.py` reveals 4 type errors:
cli/testflinger_cli/auth.py:79: error: Missing return statement  [return]
cli/testflinger_cli/auth.py:91: error: Missing return statement  [return]
cli/testflinger_cli/auth.py:126: error: Argument "fallback" to "get" of "ConfigParser" has incompatible type "None"; expected "str"  [arg-type]
cli/testflinger_cli/auth.py:180: error: Argument 1 has incompatible type "str | None"; expected "str | bytes"  [arg-type]

**Proposed fixes:**

- Lines 79 and 91: add `return None` after `_handle_auth_error()` calls — both methods declare `str` as return type but have a path that returns nothing
- Line 126: change `fallback=None` to `fallback=""` — ConfigParser.get() expects `str`, not `None`  
- Line 180: change `self.jwt_token` to `self.jwt_token or ""` in `jwt.decode()` call — jwt_token can be `str | None` but jwt.decode expects `str | bytes`

Happy to submit a PR with these fixes.

Ahmedaltu:
## Summary

Fixes type errors in `cli/testflinger_cli/auth.py` detected by mypy.

## Changes

- `_authenticate_with_credentials` and `_authenticate_with_refresh_token`: added `return None` after `_handle_auth_error()` calls and updated return types from `str` to `str | None` to reflect actual behaviour
- `get_stored_refresh_token`: changed `fallback=None` to `fallback=""` in `ConfigParser.get()` call — ConfigParser expects `str`, not `None`
- `decode_jwt_token`: changed `self.jwt_token` to `self.jwt_token or ""` in `jwt.decode()` call — `jwt_token` is `str | None` but `jwt.decode()` expects `str | bytes`

## Verification

Before:
cli/testflinger_cli/auth.py:79: error: Missing return statement  [return]
cli/testflinger_cli/auth.py:91: error: Missing return statement  [return]
cli/testflinger_cli/auth.py:126: error: Argument "fallback" to "get" of "ConfigParser" has incompatible type "None"; expected "str"  [arg-type]
cli/testflinger_cli/auth.py:180: error: Argument 1 has incompatible type "str | None"; expected "str | bytes"  [arg-type]

After:
(no errors for auth.py)

All 216 unit tests pass. tox checks pass (lock, format, lint, unit).

Closes #1028

### DIFF
--- a/cli/testflinger_cli/auth.py
+++ b/cli/testflinger_cli/auth.py
@@ -37,8 +37,8 @@ from testflinger_cli.errors import (
 class TestflingerCliAuth:
     """Class to handle authentication and authorisation for Testflinger CLI."""
 
-    def __init__(self, client_id: str, secret_key: str, tf_client):
-        self.client_id = client_id
+    def __init__(self, client_id: str | None, secret_key: str, tf_client):
+        self.client_id: Optional[str] = client_id
         self.secret_key = secret_key
         self.client = tf_client
 
@@ -76,7 +76,7 @@ class TestflingerCliAuth:
 
         return None
 
-    def _authenticate_with_credentials(self) -> str:
+    def _authenticate_with_credentials(self) -> str | None:
         """Authenticate using client_id and secret_key."""
         try:
             response = self.client.authenticate(
@@ -87,8 +87,11 @@ class TestflingerCliAuth:
             return response["access_token"]
         except client.HTTPError as exc:
             self._handle_auth_error(exc)
+            return None
 
-    def _authenticate_with_refresh_token(self, refresh_token: str) -> str:
+    def _authenticate_with_refresh_token(
+        self, refresh_token: str
+    ) -> str | None:
         """Authenticate using stored refresh token."""
         try:
             response = self.client.refresh_authentication(refresh_token)
@@ -98,6 +101,7 @@ class TestflingerCliAuth:
                 self.clear_refresh_token()
                 raise InvalidTokenError(exc.msg) from exc
             self._handle_auth_error(exc)
+            return None
 
     def _handle_auth_error(self, exc: client.HTTPError) -> None:
         """Handle authentication HTTP errors."""
@@ -172,7 +176,7 @@ class TestflingerCliAuth:
 
         :return: Dict with the decoded JWT
         """
-        if not self.is_authenticated():
+        if not self.is_authenticated() or not self.jwt_token:
             return None
 
         try:


cli/testflinger_cli/auth.py
```diff
@@ -123,7 +127,7 @@ def get_stored_refresh_token(self) -> str | None:
             return None
 
         # Set client_id from config file to keep token owner
-        self.client_id = config_file.get("AUTH", "client_id", fallback=None)
+        self.client_id = config_file.get("AUTH", "client_id", fallback="")
```
ajzobro: Why should client_id be set to an empty string instead of None here?  I think that None is appropriate.

cli/testflinger_cli/auth.py
```diff
@@ -177,7 +181,7 @@ def decode_jwt_token(self) -> Optional[dict]:
 
         try:
             decoded_jwt = jwt.decode(
-                self.jwt_token, options={"verify_signature": False}
+                self.jwt_token or "", options={"verify_signature": False}
```
ajzobro: Like above, rather than an empty string I feel like None is appropriate here as it ought to land us in the exception handler.

ajzobro (changes requested):
Thanks for your offer to contribute to this project!

Please showcase how your code changes are non-breaking by including tests which ensure that the new code is still handling these conditions the same way. Especially cases where None has been replaced with "".  See my comments below.

Ahmedaltu:
Reverted both runtime changes. Added # type: ignore[arg-type] on lines 130 and 184 to satisfy mypy while keeping None as the intended value in both cases. No behaviour change.

cli/testflinger_cli/auth.py
Ahmedaltu: Reverted to fallback=None. Added # type: ignore[arg-type] to satisfy mypy while keeping None as the intended value. No behaviour change.

cli/testflinger_cli/auth.py
Ahmedaltu: Reverted to passing self.jwt_token directly. Added # type: ignore[arg-type] to satisfy mypy. None will still land in the exception handler as intended. No behaviour change.

cli/testflinger_cli/auth.py
ajzobro:
```suggestion
    def __init__(self, client_id: str | None, secret_key: str, tf_client):
```

cli/testflinger_cli/auth.py
```diff
@@ -123,7 +127,7 @@ def get_stored_refresh_token(self) -> str | None:
             return None
 
         # Set client_id from config file to keep token owner
-        self.client_id = config_file.get("AUTH", "client_id", fallback=None)
+        self.client_id = config_file.get("AUTH", "client_id", fallback=None)   # type: ignore[arg-type]
```
ajzobro: Rather than add this comment here, we need to allow for `client_id` to be allowed to be `None`, therefore if your goal is to make mypy happy then client_id will need to be properly typed rather than ignore the discrepancy.

Use of `mypy` is no more something we intend to start supporting than, e.g. `pycharm` and all of it's various linter comment styles.  The project uses `ruff` and commonly `# noqa:` comments can be used to silence select lines as needed.

Please don't add new comments just for `mypy`.

cli/testflinger_cli/auth.py
```diff
@@ -177,7 +181,7 @@ def decode_jwt_token(self) -> Optional[dict]:
 
         try:
             decoded_jwt = jwt.decode(
-                self.jwt_token, options={"verify_signature": False}
+                self.jwt_token, options={"verify_signature": False} # type: ignore[arg-type]
```
ajzobro: The same can be said here as well; if the jwt_token can be None then we could potentially see that coming and avoid the jwt decode exception, but because it is handled, we don't need to change this here, either.  If we decide to add type checking, e.g. using `ty` with `uv` at a later time we will probably still want to find this instance.

No need to add this comment given our current tooling.

cli/testflinger_cli/auth.py
ajzobro:
```suggestion
        if not self.is_authenticated() or not self.jwt_token:
```

cli/testflinger_cli/auth.py
ajzobro: see above

ajzobro (changes requested):
Thanks again for your work on this, I have a few more comments.

Ahmedaltu:
Removed all # type: ignore comments. Fixed the types properly — client_id is now str | None in the __init__ parameter and typed as Optional[str] on the attribute. Applied both suggestions. Thanks for the guidance.

ajzobro (approved):
Looks good, thank you for your efforts and time!

This passed all the unit tests (locally) and I have approved it to be merged.

Ahmedaltu:
The CI failure is a runner permissions issue for external contributors. Could you approve the workflow run? Thanks.

PR #1034, opened 2026-04-27
fix(submit): be sure we call the snap yq command

fhost:
There is two different yq commands. The one from snap, and the one from apt. The submit action expects that yq is the one from the snap. To be sure to call the correct one, use a full path.


ajzobro (approved):
LGTM now just hope that the snap is installed :-)

PR #1037, opened 2026-04-28
feat(server/cli): include an option for secrets TTL

rene-oromtz:
## Description
This PR allows to define a custom expiration (time to live) for the secrets. 

With this users have a more granular control on how to manage their sensitive data stored in Testflinger. 
Users now can:
1. Specify any expiration in seconds, after which the secret is automatically removed
2. Specify no expiration
3. Specify if secret should be removed after first use (potentially useful in CI) 

This introduces a `exists` method that is used to validate secrets existence to accept a job, once the job is passed to the agent, it will then be read. If its ephemeral, it will be removed after retrieving the secret value. 

> [!NOTE]
> With TTL, users will need to account the expiration could potentially expire the secrets after submitting a job but yet to be accepted by an agent. 

## Resolved issues
Resolves [CERTTF-865](https://warthogs.atlassian.net/browse/CERTTF-865)

## Documentation
Documentation added for the new TTL usage

## Tests
Tested on staging environment:

1. Write a non-expiring secret:
```
uv run testflinger-cli --server https://testflinger-staging.canonical.com secret write my_secrets/non_expiry 'secret_value' --expire-after -1
Secret 'my_secrets/non_expiry' written successfully
This secret will not expire automatically
```

2. Make sure the submission success and secret existence is validated in the job:
```
uv run testflinger-cli --server https://testflinger-staging.canonical.com submit ~/Desktop/Testflinger/staging.yaml 
Job submitted successfully!
job_id: 597dc1e8-9c00-4c68-89a6-15daeb4e4175
```

3. Write a single-use secret:
```
uv run testflinger-cli --server https://testflinger-staging.canonical.com secret write my_secrets/single_use 'secret_value' --single-use
Secret 'my_secrets/single_use' written successfully
This secret will be deleted after first use
```

From job logs:
```
+ echo 'The secret is **28dd40**'
The secret is **28dd40**
```

4. Make sure the submission succeeds and secret existence is validated in the job:
```
uv run testflinger-cli --server https://testflinger-staging.canonical.com submit ~/Desktop/Testflinger/staging.yaml 
Job submitted successfully!
job_id: 76cb71d7-6e0c-4911-8745-e21c6a46a238
```

From job logs (the masking return is the same as on purpose added the same value to this new secret):
```
+ echo 'The secret is **28dd40**'
The secret is **28dd40**
```

5. Attempting to use single use secret will return an error on next submission:
```
uv run testflinger-cli --server https://testflinger-staging.canonical.com submit ~/Desktop/Testflinger/staging.yaml 
Unexpected error status from testflinger server: [422] Inaccessible secret paths: my_secrets/single_use
```

6. Create a secret with 10 minute expiration, expiration time is also returned:
```
uv run testflinger-cli --server https://testflinger-staging.canonical.com secret write my_secrets/expiry 'secret_value' --expire-after 600
Secret 'my_secrets/expiry' written successfully
This secret will expire at 2026-04-29 17:30:44 UTC
``` 

7. Job is able to be submitted:
```
uv run testflinger-cli --server https://testflinger-staging.canonical.com submit ~/Desktop/Testflinger/staging.yaml 
Job submitted successfully!
job_id: 3509746e-444d-4741-811d-8ba1073f96da
```

From job logs (the masking return is the same as on purpose added the same value to this new secret):
```
+ echo 'The secret is **28dd40**'
The secret is **28dd40**
```

8. After 2026-04-29 17:30:44 UTC (10 minutes), secret is unable to be reused:
```
date -u; uv run testflinger-cli --server https://testflinger-staging.canonical.com submit ~/Desktop/Testflinger/staging.yaml 
Wed Apr 29 05:33:46 PM UTC 2026
Unexpected error status from testflinger server: [422] Inaccessible secret paths: my_secrets/expiry
```

[CERTTF-865]: https://warthogs.atlassian.net/browse/CERTTF-865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/src/testflinger/api/v1.py
```diff
@@ -133,12 +134,11 @@ def validate_secrets(data: dict):
...
-    for secret_path in secrets.values():
-        try:
-            current_app.secrets_store.read(client_id, secret_path)
-        except (AccessError, StoreError, UnexpectedError):
-            inaccessible_paths.append(secret_path)
+    inaccessible_paths = [
+        secret_path
+        for secret_path in secrets.values()
+        if not current_app.secrets_store.exists(client_id, secret_path)
+    ]
```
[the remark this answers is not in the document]
rene-oromtz: This was added as part of the same suggestion in https://github.com/canonical/testflinger/pull/1037#discussion_r3157188653
Exists now only matches non expiry secrets

server/src/testflinger/api/v1.py
```diff
@@ -1223,7 +1223,18 @@ def secrets_put(client_id, path, json_data):
...
+            key=path,
+            value=json_data["value"],
+            **{"ephemeral": True}
+            if json_data.get("ephemeral", False)
+            else {
+                "expire_after": json_data.get(
+                    "expire_after", DEFAULT_SECRET_EXPIRATION
+                )
+            },
+        )
```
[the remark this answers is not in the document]
rene-oromtz: Unit tests added in [d0aac3d](https://github.com/canonical/testflinger/pull/1037/commits/d0aac3dc0dc06c697fad5c6661caf3f43d00be6b)

docs/reference/secrets.rst
```diff
@@ -36,3 +36,24 @@ Environment variable names must follow standard shell naming conventions:
...
+- Specify any expiration (in seconds) after which the secret is automatically removed.
+- Specify no expiration. If needed, users can still remove it manually.
+- Specify single use secrets. Useful if you don't want your data to be retained by 
+  Testflinger and remove it immediately after its first use. 
+
+.. note:: 
+
+    The expiration of the secret is considered from the time it was last created or updated.
+
+For more information on how to define expiration for a secret,
```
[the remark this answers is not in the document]
rene-oromtz: Vault Store implementation was not pushed before review, but its available in [2fef19d](https://github.com/canonical/testflinger/pull/1037/commits/2fef19de0cddd74ded984cd97b3d3cc76657e146)

server/src/testflinger/api/v1.py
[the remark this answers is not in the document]
rene-oromtz: Added `@validates_schema` to match other mutually exclusive schemas

server/src/testflinger/secrets/mongo.py
```diff
@@ -46,6 +48,27 @@ def __init__(
...
+
+    def _ensure_ttl_index(self, namespace: str):
+        """Create the TTL index for a namespace collection, if needed."""
+        if namespace in self._ttl_index_initialized_namespaces:
+            return
+
+        try:
+            self.database.secrets[namespace].create_index(
+                "expire_at", expireAfterSeconds=0
+            )
```
rene-oromtz, line 61: FYI, expireAfterSecond=0 indicates the expiration is determined by specific clock time instead of seconds:
https://www.mongodb.com/docs/manual/tutorial/expire-data/#expire-documents-at-a-specific-clock-time

docs/how-to/use-secrets.rst
```diff
@@ -33,6 +33,24 @@ that is automatically determined by the authenticated Testflinger ``client-id``.
...
+if you don't specify any expiration.
+
+.. code-block:: shell
+
+    testflinger-cli secret write path/to/secret 'my_secret_value' --expire-after 604800 # 7 days
+
+.. tip::
+    For a non-expiring secret you can specify ``-1`` in the ``--expire-after`` argument
+
+If you prefer the secret to be single-use and expire after its first use, you can also use the
```
ajzobro:
```suggestion
If you prefer the secret to be single-use and expire after its first use, you can alternatively use the
```

docs/reference/secrets.rst
```diff
@@ -36,3 +36,24 @@ Environment variable names must follow standard shell naming conventions:
 
 - Must start with a letter or underscore
 - Can only contain letters, numbers and underscores
+
+Retention
+---------
+
+Secrets in Testflinger are by default retained for a year after which they
```
ajzobro:
```suggestion
Secrets in Testflinger are, by default, retained for a year after which they
```

docs/reference/secrets.rst
```diff
@@ -36,3 +36,24 @@ Environment variable names must follow standard shell naming conventions:
...
+are automatically removed. This is to prevent stale data to pile up and to let
```
ajzobro:
```suggestion
are automatically removed. This is to prevent stale data to pile up and to ensure 
```

docs/reference/secrets.rst
```diff
@@ -36,3 +36,24 @@ Environment variable names must follow standard shell naming conventions:
...
+Retention
+---------
+
+Secrets in Testflinger are by default retained for a year after which they
+are automatically removed. This is to prevent stale data to pile up and to let 
+users control the data retention of their private data automatically. 
+
+Users can:
+
+- Specify any expiration (in seconds) after which the secret is automatically removed.
```
ajzobro:
```suggestion
- Specify any expiration (in seconds), after which the secret is automatically removed.
```

docs/reference/secrets.rst
```diff
@@ -36,3 +36,24 @@ Environment variable names must follow standard shell naming conventions:
...
+- Specify no expiration. If needed, users can still remove it manually.
+- Specify single-use secrets. Useful if you don't want your data to be retained by 
+  Testflinger and remove it immediately after its first use.
```
ajzobro:
```suggestion
  Testflinger after its first use. 
```

docs/reference/secrets.rst
```diff
@@ -36,3 +36,24 @@ Environment variable names must follow standard shell naming conventions:
...
+
+Secrets in Testflinger are by default retained for a year after which they
+are automatically removed. This is to prevent stale data to pile up and to let 
+users control the data retention of their private data automatically. 
+
+Users can:
+
+- Specify any expiration (in seconds) after which the secret is automatically removed.
+- Specify no expiration. If needed, users can still remove it manually.
+- Specify single-use secrets. Useful if you don't want your data to be retained by
```
ajzobro:
```suggestion
- Specify single-use secrets. Useful if you don't need your data to be retained by 
```

docs/reference/secrets.rst
```diff
@@ -36,3 +36,24 @@ Environment variable names must follow standard shell naming conventions:
...
+Users can:
+
+- Specify any expiration (in seconds) after which the secret is automatically removed.
+- Specify no expiration. If needed, users can still remove it manually.
+- Specify single-use secrets. Useful if you don't want your data to be retained by 
+  Testflinger and remove it immediately after its first use. 
+
+.. note:: 
+
+    The expiration of the secret is considered from the time it was last created or updated.
```
ajzobro:
```suggestion
    The expiration of the secret is measured from the time it was last created or updated.
```

server/src/testflinger/secrets/mongo.py
```diff
@@ -46,6 +48,27 @@ def __init__(
         self.cipher = cipher
         self.data_key_name = data_key_name
         self.algorithm = algorithm
+        self._ttl_index_initialized_namespaces = set()
```
ajzobro, line 51: It seems like this will start as an empty set for each instance of the server (e.g. each of the k8s charms) -- shouldn't we be asking mongo if the index exists?

server/src/testflinger/secrets/mongo.py
```diff
@@ -46,6 +48,27 @@ def __init__(
...
+    def _ensure_ttl_index(self, namespace: str):
+        """Create the TTL index for a namespace collection, if needed."""
+        if namespace in self._ttl_index_initialized_namespaces:
+            return
+
+        try:
+            self.database.secrets[namespace].create_index(
+                "expire_at", expireAfterSeconds=0
+            )
+        except OperationFailure as error:
```
ajzobro, line 62: What exception is thrown (any?) if mongo db already had an index on `expires_at`?

server/src/testflinger/secrets/mongo.py
```diff
@@ -63,6 +86,14 @@ def read(self, namespace: str, key: str) -> str:
             raise AccessError(f"Unable to access '{key}' under '{namespace}'")
 
         encrypted_value = result["value"]
+
+        # Deny access to expired secrets while waiting for MongoDB TTL cleanup
+        expire_at = result.get("expire_at")
+        if expire_at is not None and expire_at.replace(
```
ajzobro:
```suggestion
        if expire_at and expire_at.replace(
```

server/src/testflinger/secrets/mongo.py
```diff
@@ -73,15 +104,38 @@ def read(self, namespace: str, key: str) -> str:
...
+        value: str,
+        expire_after: int | None = DEFAULT_SECRET_EXPIRATION,
+        ephemeral: bool = False,
+    ) -> datetime | None:
+        """Write the `value` for `key` under `namespace`.
+
+        :param namespace: the namespace under which to store the secret
+        :param key: the key for the secret
+        :param value: the value of the secret to store
+        :param expire_after: Expiration time in seconds for the secret.
```
ajzobro, line 136: Add notes and any missing logic to this function to ensure the mutual exclusivity of the `expire_after` and `ephemeral` arguments. This function has no protective decorators (e.g. schema checks).  It would be a programming error if we were to call this function with mismatched arguments, but we should protect against that so any such conditions will be caught.

server/src/testflinger/secrets/mongo.py
```diff
@@ -73,15 +104,38 @@ def read(self, namespace: str, key: str) -> str:
...
+        :param ephemeral: whether the secret should be deleted after being read
+        :returns: The expiry datetime if a TTL was set, otherwise None.
+        :raises AccessError: if the secret cannot be accessed
+        :raises StoreError: if there is an issue with the MongoDB store
+        """
         try:
             encrypted_value = self.cipher.encrypt(
                 value=value.encode("utf-8"),
```
ajzobro, line 144: We should probably protect against empty strings, too.

server/src/testflinger/secrets/mongo.py
```diff
@@ -120,3 +192,27 @@ def delete(self, namespace: str, key: str):
...
+            return (
+                self.database.secrets[namespace].find_one(
+                    {
+                        "key": key,
+                        "$or": [
+                            {"expire_at": {"$exists": False}},
+                            {"expire_at": {"$gt": datetime.now(timezone.utc)}},
+                        ],
+                    },
+                    {"_id": 1},
```
ajzobro, line 216: isn't this line just specifying that the `_id` field must be present in the response?  According to (https://www.mongodb.com/docs/manual/reference/method/db.collection.find/#inclusion-or-exclusion) that is the default:

> projection (document): Specifies which fields to include or exclude in the returned document.
> Use 1 or true to include a field; 0 or false to exclude it.
> The _id field is included by default unless explicitly excluded.

In other words: could we leave that out? or does this make the response smaller by filtering out all the other data?  we don't really want to fetch the data anyway -- and findone doesn't return a cursor.

ajzobro (commented):
Will be back to review tests, work in progress.

server/src/testflinger/secrets/mongo.py
rene-oromtz: This was added to not waste a DB call if the in memory process already holds a reference for the TTL in the existing set. If we do call for create index, it won't matter much as the operation is idempotent, if it has the same definition, it skips without error so its not worth to add another test to check on index existence.

server/src/testflinger/secrets/mongo.py
rene-oromtz: None, if the index is the same it won't attempt to create a new index and won't fail if already exists. 
This is the same as we are currently doing in the database module where we are defining our other indexes.

server/src/testflinger/secrets/mongo.py
rene-oromtz: I'll add a note that they are mutually exclusive and that this is guaranteed by schema. 
The function currently "protects" itself as if ephemeral is defined, it will always be prioritized no matter is expire_after was provided. In our case, the schema is the right place for further validation logic.

server/src/testflinger/secrets/mongo.py
rene-oromtz: I'll add a validation check at schema level for requiring at least 1 character. This is the right place to ensure secret value is not empty

server/src/testflinger/secrets/mongo.py
rene-oromtz: Yes, this makes the response smaller as it only returns the id if it exists (excluding all data). As we are validating the existence, there is no need to include all the other fields but it doesn't affect if we do... Guess we can leave it as-is taking this was copilot suggestion and there is not much gain or lose either way. I think my original approach was doing a find one without any filtering

docs/reference/secrets.rst
```diff
@@ -36,3 +36,24 @@ Environment variable names must follow standard shell naming conventions:
...
+Users can:
+
+- Specify any expiration (in seconds), after which the secret is automatically removed.
+- Specify no expiration. If needed, users can still remove it manually.
+- Specify single-use secrets. Useful if you don't need your data to be retained by 
+  Testflinger after its first use.
+
+.. note:: 
+
+    The expiration of the secret is neasured from the time it was last created or updated.
```
ajzobro:
```suggestion
    The expiration of the secret is measured from the time it was last created or updated.
```

ajzobro (changes requested):
one letter off ;-)

docs/reference/secrets.rst
rene-oromtz: :(

ajzobro (approved):
LGTM

PR #1004, opened 2026-04-06
feat(server): migrate charm to use uv plugin 

rene-oromtz:
## Description
This PR changes the way charm is packed by migrating server charm to use `uv` plugin instead of `charm` plugin. This results in a faster and more reliable `charmcraft pack` as this uses pre-built wheels instead of building all dependencies for charm on every single run. This is heavily inspired in another of our [certification charm](https://github.com/canonical/hexr/pull/902)

As a bonus, this also [adds the coverage](https://app.codecov.io/gh/canonical/testflinger/commit/7b4096e1564708f7dd218d71455a115ff1962479/indirect-changes) that was missing from the server charm code.

### Implementation

- Create `pyproject.toml` for Charm
- Generate uv.lock from `pyproject.toml`
- Remove `requirements.txt`
- Use `uv` plugin in charmcraft.yaml
- Add `tox.ini` based on what we use in our other projects
- Move test to use `server/charm` as working directory
- Updated `README.md` and charm `CONTRIBUTING.md`

## Resolved issues
Resolves [CERTTF-894](https://warthogs.atlassian.net/browse/CERTTF-894)

## Documentation
NA

## Web service API changes
No change to the server, only the way server is packed

## Tests
No additional unit and integration test added, this uses the current server charm module. 

[CERTTF-894]: https://warthogs.atlassian.net/browse/CERTTF-894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1039, opened 2026-04-29
refactor(agent): migrate charm to uv plugin

rene-oromtz:
## Description
This PR changes the way charm is packed by migrating agent charm to use `uv` plugin instead of `charm` plugin. This results in a faster and more reliable `charmcraft pack` as this uses pre-built wheels instead of building all dependencies for charm on every single run. This is the same process as the one made in #1004 

As a bonus, this also adds the coverage that was missing from the agent charm code.

### Implementation

- Create `pyproject.toml` for Charm
- Generate uv.lock from `pyproject.toml`
- Remove `requirements.txt` and `requirements-dev.txt`
- Use `uv` plugin in charmcraft.yaml
- Add `tox.ini` based on what we use in our other projects and server changes
- Move test to use `agent/charms/testflinger-agent-host-charm` as working directory
- Updated `README.md` and charm `CONTRIBUTING.md`
- Updated `agent-tox.yaml` and `tics.yaml` with this new reference

> [!NOTE]
> It seems that on the previous structure, files were not formatted correctly specifically with the libraries sorting.
> The new modification makes the libraries to be sorted differently after running `format`

## Resolved issues
Resolves [CERTTF-1047](https://warthogs.atlassian.net/browse/CERTTF-1047)

## Documentation
NA

## Web service API changes
No change to the agent, only the way agent is packed

## Tests
No additional unit and integration test added, this uses the current agent charm tests. 

[CERTTF-1047]: https://warthogs.atlassian.net/browse/CERTTF-1047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


agent/charms/testflinger-agent-host-charm/uv.lock
```diff
@@ -0,0 +1,995 @@
+version = 1
+revision = 3
+requires-python = ">=3.10"
```
ajzobro, line 3: Should we provide .python-version to lock this to 3.10 specifically?

agent/charms/testflinger-agent-host-charm/charmcraft.yaml
```diff
@@ -109,6 +109,14 @@ parts:
 
       craftctl default
 
+  # Need a separate part for copying the templates
+  # uv plugin only copies `src` and `lib` directories into the final charm
+  templates:
+    plugin: dump
+    source: .
+    stage:
+    - templates
```
ajzobro, line 118: There are some other files, e.g. `files/ssh-config` that will not be included (anymore?/now?) are they still needed or can we get rid of the files that aren't going to be included in the new charm?

Another files that appears to have out-lived its use is `run_tests`

agent/charms/testflinger-agent-host-charm/pyproject.toml
```diff
@@ -0,0 +1,75 @@
...
+    "W",   # pycodestyle warnings
+]
+ignore = [
+    "D100", # docstring: Missing docstring in public module
+    "D102", # docstring: Missing docstring in public method
+    "D103", # docstring: Missing docstring in public function
+    "D104", # docstring: Missing docstring in public package
+    "D107", # docstring: Missing docstring in __init__
+    "D205", # docstring: missing-blank-line-after-summary
+]
```
ajzobro: Why would we ignore the missing docstrings instead of just add them?

ajzobro (commented):
A few questions...

agent/charms/testflinger-agent-host-charm/uv.lock
rene-oromtz: I can add it! Jammy defaults to 3.10 but will give us more control once we jump to noble

agent/charms/testflinger-agent-host-charm/charmcraft.yaml
rene-oromtz: files/ssh-config I think its not needed, not sure if it previously was but now the ssh config is defined at charm config. 
`run_tests` is something I forgot to remove! Thanks for the heads up

agent/charms/testflinger-agent-host-charm/pyproject.toml
rene-oromtz: We should... I added the same exceptions we had but we probably should. Let me see how much changes are required, if this is something quick I can add it right away.

agent/charms/testflinger-agent-host-charm/charmcraft.yaml
rene-oromtz: both files are now removed after confirming those are no longer needed

agent/charms/testflinger-agent-host-charm/pyproject.toml
rene-oromtz: I added all the missing docstrings... This add a bunch of changes so please let me know if you prefer I tackle those on a follow up PR. I can just revert the last commit and push that one onto a different branch/PR if it makes this easier to review

ajzobro (approved):
LGTM

agent/charms/testflinger-agent-host-charm/pyproject.toml
ajzobro: Could just add them in the next PR -- and could probably just use co-pilot for that task.

rene-oromtz: I added them manually lol at the end it was "Copilot" via VS Code autocomplete :laughing: I should probably trust more Copilot for doing this type of tasks automatically...

PR #1040, opened 2026-04-30
Fix Reserve Job documentation page

pieqq:
Fix .rst typo that prevented the list to display properly.


rene-oromtz (approved):
Thanks for fixing this!

PR #1044, opened 2026-05-01
chore: allow pytest parallelization

rene-oromtz:
## Description
This PR is meant to improve the dev experience by allowing parallelization via [pytest-xdist](https://github.com/pytest-dev/pytest-xdist). This plugin spawns a number of workers equal to the number of cpus available (which can be way faster on local environments). 

This is already done in the agent code so trying to extend it to the other components too. Currently, all tests in all test files are fairly independent but worth nothing that if we ever want to preserve state between states, the tests will be affected as they are randomly assigned to a worked. 

As a bonus, I also added `term-missing:skip-covered` which will show the lines that are missing coverage. The skip covered will hide it from the terminal output but they are still available in the coverage report. 

## Resolved issues
NA

## Documentation
NA

## Web service API changes
NA

## Tests
NA


cli/.gitignore
```diff
@@ -1,3 +1,5 @@
 attachments.tar.gz
 test/test.json
 coverage.xml
+
+.coverage
```
rene-oromtz:
```suggestion
.coverage

```

ajzobro: Interesting.  I hadn't realized until now that we had so many .gitignore files and on my local edits for this effort, I had instead modified the toplevel .gitignore with:

`**/.coverage*`

Because I was seeing:

`cli/.coverage.<hostname>.pid3487534.XCbmyNxx.HI5pM4gTRTih`

ajzobro (dismissed):
Only question is about the .gitignore path

cli/.gitignore
rene-oromtz: Yeah, I added here because for some reason in my run I had those... But seems it only happens if the run don't complete cleanly as the coverage files are normally combined. I'll add it at the top level instead. 
Seems this should be enough: `.coverage*` as per [gitignore](https://git-scm.com/docs/gitignore) docs:

> If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level

rene-oromtz: Just pushed in [f3934d2](https://github.com/canonical/testflinger/pull/1044/commits/f3934d2108887ff16ff1b466acc1b724ec4e01fc)[f3934d2](https://github.com/canonical/testflinger/pull/1044/commits/f3934d2108887ff16ff1b466acc1b724ec4e01fc)! This should be a bit more clean

ajzobro (approved):

PR #1045, opened 2026-05-04
docs: add an example on ephemeral deployments

rene-oromtz:
## Description
This modifies the existing test job example for MAAS devices to include reference for ephemeral deployments. 
It also modifies the wording as MAAS is not only for servers so the wording is now more general.

## Resolved issues
NA. Considering user feedback


ajzobro (approved):
Looks great!

PR #1053, opened 2026-05-08
chore: Use pytest jubilant for charm integration tests

pedro-avalos:
## Description

- This PR adds [`pytest-jubilant`](https://github.com/canonical/pytest-jubilant) to charm integration tests
- This helps in case someone wants to test locally by skipping setup or teardown steps
- It also extracts the `juju` fixture out to the fixture provided by pytest-jubilant (which is used by other charms)
- Lastly, the Juju log can now be dumped and stored as an artifact in the CI
- I have also removed `ops[testing]` from integration testing dependencies, as I don't think it's actually needed for integration tests

### DIFF
--- a/.github/workflows/agent-tox.yml
+++ b/.github/workflows/agent-tox.yml
@@ -109,4 +109,11 @@ jobs:
           if-no-files-found: error
           retention-days: 2
       - name: Run charm integration tests
-        run: uvx --with tox-uv tox -e integration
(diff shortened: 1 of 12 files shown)


rene-oromtz (approved):
Thank you for taking the time to add this! LGTM
Integration tests are still running but seems a GH issue assigning runners and setting up uv 🤔

ajzobro (approved):
LGTM

PR #1055, opened 2026-05-08
feat(dev): introduce just to monorepo

rene-oromtz:
## Description
This PR takes inspiration on how other Canonical projects are managing monorepos e.g. [charmlibs](https://github.com/canonical/charmlibs) by using [just](https://github.com/casey/just)

Just is essentially a command runner which in our case can be leveraged to improve developer experience while working locally. The proposed approach is not breaking and does not add complexity by reusing the `tox.ini` file and the environments/commands defined there. 

This introduces "recipes" that are easy to remember and that can be executed from any directory inside the repository: 

1. `justfile`: main file where recipes are located and where just usage is defined, it is defined per Testflinger component or with the suffix "all" to execute on all components. 
2. `docs.just`: recipes that are documentation only. 

## Resolved issues
NA

## Documentation
Updated the `CONTRIBUTING.md` making just as the preferred approach given the benefits but also kept the old reference to run using `uvx` directly. 

## Tests
Demo'ed and tested locally:

1. List general info
```
$ just
Canonical Testflinger Monorepo

List available components in this monorepo with just list-components

List all commands with just help, or:
- Run all checks for all components: just check-all
- Run ruff format for all components: just format-all
- Run ruff for all components: just lint-all

Run individual checks for a component:
- just check <component> (Run all checks: lint, format, unit tests)
- just lint <component> (Check linting issues)
- just format <component> (Solves all fixable lint and formatting issues)
- just unit <component> (Run unit tests)
- just check-schema (Check server schema is up to date)
- just schema (Generate server schema)

Build the docs: just docs
```

2. Running a recipe inside a component but reference another:
```
~/Documents/Canonical/testflinger/agent/charms/testflinger-agent-host-charm/src$ just format server
...
format: commands[0]> djlint --reformat src

Reformatting 12/12 files ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 00:00    


0 files were updated.

format: commands[1]> ruff format
48 files left unchanged
format: commands[2]> ruff check --fix
All checks passed!
  format: OK (0.31=setup[0.02]+cmd[0.27,0.01,0.01] seconds)
  congratulations :) (0.32 seconds)
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,20 @@ You can install `uv` with snap:
 sudo snap install --classic astral-uv
 ```
 
+Additionally, this project uses [just] as a task runner, you can install it with:
+
+```shell
+uv tool install rust-just
+```
+
+Then run `just` from anywhere in the repository for usage.
+
 ### Development Environment
+
+[just] provides an alternative way to run formatting, linting and unit tests without needing to be located in each subproject.
+All `just` recipes in this repository use `uv` underneath.
+
+If you prefer using `uv` directly, you'll need to set up your virtual environments manually.
 
 Within a subproject, you can set up the virtual environment with the following
 command:
@@ -40,11 +53,25 @@ To learn more about `uv`, refer to the [`uv` documentation][uv].
 
 #### Add a Dependency
 
-To add a new dependency to a subproject, please use `uv`, as
-it will automatically add it to both the `pyproject.toml` and `uv.lock` files:
+To add a new dependency to a component, please use `just`, as
+it will automatically add it to both the `pyproject.toml` and `uv.lock` files
+for the specified component.
 
 ```shell
-uv add ...
+just add <component> <flags> <package>
+```
+
+e.g.
+
+```shell
+just add server 'urllib>=2.6.3'
+```
+
+Alternatively, you can also use `uv` within each component subproject:
+
+```shell
+cd <component>
+uv add <package>
 ```
 
 If the dependency is only a development dependency, please add it to the `dev`
@@ -55,12 +82,17 @@ To learn more about the `uv add` command, refer to the
 
 ### Remove a Dependency
 
-To remove a dependency from a subproject, please use `uv`, as
+To remove a dependency from a subproject, please use `just`, as
 it will automatically remove it from both the `pyproject.toml` and `uv.lock`
 files:
 
 ```shell
-uv remove ...
+just remove <component> <package>
+```
+
+Alternatively, you can also use `uv` within each component subproject:
+```shell
+uv remove <package>
 ```
 
 If the dependency is only a development dependency, please remove it from the
@@ -75,6 +107,12 @@ If there is a discrepancy between a subproject's `pyproject.toml` and lock file,
 you can generate the lock file (`uv.lock`) with:
 
 ```shell
+just lock <component>
+```
+
+Or alternatively, within each subproject directory:
+
+```shell
 uv lock
 ```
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -84,12 +125,37 @@ To learn more about the `uv lock` command, refer to the
...
-subproject where you made changes.
+Before pushing anything, it's a good idea to run tests for the specified component:
+
+```shell
+just check <component>
+```
+
+This will run all available checks for the specified component. You can also run them individually:
+
+
```
ajzobro:
```suggestion
```

docs.just
```diff
@@ -0,0 +1,62 @@
+# Docs recipes, continue using Makefile as source of truth
+set working-directory := 'docs'
+
+[doc('Watch, build, and serve the docs.')]
```
ajzobro: Is "Watch" the right verb here?

docs.just
```diff
@@ -0,0 +1,62 @@
...
+run:
+    make run
+
+[doc('Show help for the docs recipes.')]
+help:
+    @just --list docs --unsorted
```
ajzobro, line 10: I need to learn more about just, obviously, but will the docs.just file be incorporated into the broader just command?

Will this doc-specific help be exposed at the top level?

docs.just
```diff
@@ -0,0 +1,62 @@
...
+
+[doc('Only remake demo tapes with vh.')]
```
ajzobro:
```suggestion
[doc('Only remake demo tapes with vhs.')]
```

justfile
```diff
@@ -0,0 +1,153 @@
+mod docs  # load docs module to expose docs subcommands
```
ajzobro, line 1: Ah, yes, the answer to an earlier question.

justfile
```diff
@@ -0,0 +1,153 @@
...
+    @echo ''
+    @echo '{{BOLD}}Run individual checks for a component:{{NORMAL}}'
+    @echo '- {{CYAN}}just check <component>{{NORMAL}} (Run all checks: lint, format, unit tests)'
+    @echo '- {{CYAN}}just lint <component>{{NORMAL}} (Check linting issues)'
+    @echo '- {{CYAN}}just format <component>{{NORMAL}} (Solves all fixable lint and formatting issues)'
+    @echo '- {{CYAN}}just unit <component>{{NORMAL}} (Run unit tests)'
+    @echo '- {{CYAN}}just check-schema{{NORMAL}} (Check server schema is up to date)'
+    @echo '- {{CYAN}}just schema{{NORMAL}} (Generate server schema)'
+    @echo ''
+    @echo '{{BOLD}}Build the docs: {{CYAN}}just docs{{NORMAL}}'
```
ajzobro, line 26: which part of `docs.just` does this run?

are the docs subcommands all under `docs`? e.g. `just docs help` would run `docs.just:help`?

docs.just
rene-oromtz: Yes, I took that description from our [Makefile](https://github.com/canonical/testflinger/blob/489ea52bf5c985a7e1ad89aa61db05a93c29493b/docs/Makefile#L27) and even in the sphinx docs guide there are using the same verb: https://github.com/canonical/sphinx-docs-guide/blob/main/Makefile#L38

docs.just
rene-oromtz: It will be incorporated but the help is not showed in the top level, you need to `just docs help` to get the docs help specifics

justfile
rene-oromtz: Its the first recipe in the `docs.just` file, in this case `just docs run`. I mirrored the pattern in charmlibs repo and thought that probably this is the command that we will run more often but if you prefer, I can change it to the help. 

And yes, all the docs subcommand follow the same pattern you mentioned: `just docs <subcommand>`

docs.just
ajzobro: Could you add that simple clue to the base help so that running `just docs help` is the first thing people do for the docs?

rene-oromtz: Sure thing!

rene-oromtz:
@ajzobro thanks for the review! All requested changes were added, please let me know if something else is missing

docs.just
ajzobro: That still doesn't make it right.

justfile
ajzobro: `just help` is helpful.
`just` is also helpful.

Not the same, but *meh* it's not the product, it's a helpful tool.

docs.just
ajzobro:
```suggestion
[doc('Build, serve and preview the docs.')]
```

ajzobro (changes requested):
Tiny change, looks nice other than this annoyance.

docs.just
rene-oromtz: As discussed in MM, added instead `serve, build and auto-rebuild on change`

ajzobro (approved):
LGTM, thanks for this.  I will be proposing more changes in the future, e.g. "just build agent charm" or similar

PR #1111, opened 2026-05-21
feat(server-charm): add configuration options to set OIDC

rene-oromtz:
## Description
This PR adds the required configuration options so it can enable OIDC. The ground work was already made so this only adds the configuration at the charm level. 

It also adds `ProxyFix` conditionally based on if the application is running behind a proxy, this is different than the HTTP_PROXY and NO_PROXY as those are used for outbound and this is required for inbound requests. Without ProxyFix, the OIDC provider is unable to redirect back to the application. 

~Finally, this also "sneaks" a hotfix needed with PyMongo > 4.9 e.g. db["fs.<collection>"] otherwise the application is unable to start~ I moved pymongo fix to #1118 to also include some unit tests as they were missing entirely for GridFS

> [!NOTE]
> Integration tests are failing because of this MongoDB issue, the tests uses the image that is available on `main` until this fix is landed, expect failures as the application will never start (it crashes with the above traceback)

## Resolved issues
Resolves [CERTTF-714](https://warthogs.atlassian.net/browse/CERTTF-714)

## Tests
Added unit tests for charm configuration and also tested on staging:
Screenshot from 2026-05-21 14-23-28

[CERTTF-714]: https://warthogs.atlassian.net/browse/CERTTF-714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/charm/charmcraft.yaml
+++ b/server/charm/charmcraft.yaml
@@ -112,12 +112,12 @@ config:
       type: string
       default: "localhost,127.0.0.1,::1"
       description: Resources that we should be able to access bypassing proxy
-    behind_proxy:
+    enable_proxyfix:
       type: boolean
       default: false
       description: |
-        Whether the application is running behind a proxy. 
-        Required if OIDC is enabled and the application is running behind a proxy,
+        Enable Flask's ProxyFix middleware.
+        Required when the application is behind a reverse proxy and OIDC is enabled,
         so that redirects from the OIDC provider can be properly handled.
     webhook_url:
       type: string
--- a/server/charm/charmcraft.yaml
+++ b/server/charm/charmcraft.yaml
@@ -120,6 +120,22 @@ config:
       type: string
       default: ""
       description: Optional bearer token for authenticating with the webhook endpoint
+    web_secret_key:
+      type: string
+      default: ""
+      description: The secret key used for signing web sessions and cookies
+    oidc_client_id:
+      type: string
+      default: ""
+      description: The OIDC client ID used for authenticating with the OIDC provider
+    oidc_client_secret:
+      type: string
+      default: ""
+      description: The OIDC client secret used for authenticating with the OIDC provider
+    oidc_provider_issuer:
+      type: string
+      default: ""
+      description: The OIDC provider issuer URL for authenticating users
 
 
 parts:
--- a/server/charm/charmcraft.yaml
+++ b/server/charm/charmcraft.yaml
@@ -112,6 +112,13 @@ config:
       type: string
       default: "localhost,127.0.0.1,::1"
       description: Resources that we should be able to access bypassing proxy
+    behind_proxy:
+      type: boolean
+      default: false
+      description: |
+        Whether the application is running behind a proxy. 
+        Required if OIDC is enabled and the application is running behind a proxy,
+        so that redirects from the OIDC provider can be properly handled.
     webhook_url:
       type: string
       default: "http://test-observer-api.local/"
--- a/server/charm/src/charm.py
+++ b/server/charm/src/charm.py
@@ -496,6 +496,10 @@ class TestflingerCharm(ops.CharmBase):
             "NO_PROXY": self.typed_config.no_proxy,
             "WEBHOOK_URL": self.typed_config.webhook_url,
             "WEBHOOK_AUTH": self.typed_config.webhook_auth,
+            "WEB_SECRET_KEY": self.typed_config.web_secret_key,
+            "OIDC_CLIENT_ID": self.typed_config.oidc_client_id,
+            "OIDC_CLIENT_SECRET": self.typed_config.oidc_client_secret,
+            "OIDC_PROVIDER_ISSUER": self.typed_config.oidc_provider_issuer,
         }
         return env
 
--- a/server/charm/src/charm.py
+++ b/server/charm/src/charm.py
@@ -494,7 +494,7 @@ class TestflingerCharm(ops.CharmBase):
             "HTTP_PROXY": self.typed_config.http_proxy,
             "HTTPS_PROXY": self.typed_config.https_proxy,
             "NO_PROXY": self.typed_config.no_proxy,
-            "BEHIND_PROXY": str(self.typed_config.behind_proxy),
+            "ENABLE_PROXYFIX": str(self.typed_config.enable_proxyfix),
             "WEBHOOK_URL": self.typed_config.webhook_url,
             "WEBHOOK_AUTH": self.typed_config.webhook_auth,
             "WEB_SECRET_KEY": self.typed_config.web_secret_key,
--- a/server/charm/src/charm.py
+++ b/server/charm/src/charm.py
@@ -494,6 +494,7 @@ class TestflingerCharm(ops.CharmBase):
             "HTTP_PROXY": self.typed_config.http_proxy,
             "HTTPS_PROXY": self.typed_config.https_proxy,
             "NO_PROXY": self.typed_config.no_proxy,
+            "BEHIND_PROXY": str(self.typed_config.behind_proxy),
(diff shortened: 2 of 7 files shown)


server/tests/test_app.py
```diff
@@ -31,3 +32,15 @@ def test_setup_mongo_fails_without_config():
...
+def test_proxyfix_enabled(monkeypatch):
+    """Ensure ProxyFix is enabled when BEHIND_PROXY is true."""
+    monkeypatch.setenv("BEHIND_PROXY", "true")
+    app = create_flask_app(type("", (), {"TESTING": True})())
+    assert isinstance(app.wsgi_app, ProxyFix)
+
+
+def test_proxyfix_disabled(testapp):
+    """Ensure ProxyFix is disabled when BEHIND_PROXY is not set."""
+    assert not isinstance(testapp.wsgi_app, ProxyFix)
```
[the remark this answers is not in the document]
rene-oromtz: `test_proxyfix_enabled` sets the envvar at the function level, the `test_proxyfix_disabled` is isolated from this and uses the testapp correctly which has the default to false

server/charm/src/config.py
```diff
@@ -18,8 +18,13 @@ class TestflingerServerConfig(pydantic.BaseModel):
     http_proxy: str = ""
     https_proxy: str = ""
     no_proxy: str = "localhost,127.0.0.1,::1"
+    behind_proxy: bool = False
     webhook_url: str = "http://test-observer-api.local/"
     webhook_auth: str = ""
+    web_secret_key: str = ""
+    oidc_client_id: str = ""
```
ajzobro, line 25: If the webhook gets to have test-observer referenced, shouldn't we be able to put "testfinger" here as a default/example value?
```suggestion
    oidc_client_id: str = "testflinger"
```

server/charm/tests/unit/test_config.py
```diff
@@ -51,3 +51,93 @@ def test_invalid_webhook_url():
...
+    """Test that invalid OIDC provider issuer raises validation error."""
+    # Reject oidc_provider_issuer that do not include protocol
+    with pytest.raises(ValueError):
+        TestflingerServerConfig(
+            oidc_client_id="client-id",
+            oidc_client_secret="client-secret",  # noqa: S106
+            oidc_provider_issuer="oidc-provider.local",
+            web_secret_key="web-secret-key",  # noqa: S106
+        )
+
```
ajzobro, line 116:
```suggestion

    # Test valid OIDC provider with path
```

server/charm/tests/unit/test_config.py
```diff
@@ -51,3 +51,93 @@ def test_invalid_webhook_url():
...
+        oidc_client_id="client-id",
+        oidc_client_secret="client-secret",  # noqa: S106
+        oidc_provider_issuer="https://keycloak.example.com/realms/myrealm",
+        web_secret_key="web-secret-key",  # noqa: S106
+    )
+    assert (
+        config.oidc_provider_issuer
+        == "https://keycloak.example.com/realms/myrealm"
+    )
+
```
ajzobro, line 128:
```suggestion

    # Test valid OIDC provider with no path, ending in slash
```

server/charm/charmcraft.yaml
```diff
@@ -120,6 +127,22 @@ config:
       type: string
       default: ""
       description: Optional bearer token for authenticating with the webhook endpoint
+    web_secret_key:
+      type: string
+      default: ""
+      description: The secret key used for signing web sessions and cookies
+    oidc_client_id:
+      type: string
+      default: ""
```
ajzobro, line 136:
```suggestion
      default: "testflinger"
```

server/src/testflinger/application.py
```diff
@@ -103,6 +105,12 @@ def inject_oidc_status():
     def inject_vanilla_framework_version():
         return {"vanilla_framework_version": VANILLA_FRAMEWORK_VERSION}
 
+    # Tell Flask it's behind a proxy so it can properly handle redirects
+    if os.environ.get("BEHIND_PROXY", "false").lower() == "true":
+        tf_app.wsgi_app = ProxyFix(
```
ajzobro, line 114: Given the notes in the ProxyFix documentation (following) I feel it may be prudent to leave a comment here in case someone encounters a differently configured proxy in their environment in the future.

> This middleware should only be used if the application is actually behind such a proxy, and should be configured with the number of proxies that are chained in front of it. Not all proxies set all the headers.

server/charm/src/charm.py
```diff
@@ -494,8 +494,13 @@ def app_environment(self) -> dict:
             "HTTP_PROXY": self.typed_config.http_proxy,
             "HTTPS_PROXY": self.typed_config.https_proxy,
             "NO_PROXY": self.typed_config.no_proxy,
```
ajzobro, line 496: I'm not a fan of having two opposing variables which are independently set.

If `NO_PROXY` then we cannot `BEHIND_PROXY` -- is there a way that we can have one variable for the concept.

If I am mistaken about their specific uses, then renaming is in order, but presently they appear to be opposite sides of a boolean choice.

server/charm/src/config.py
rene-oromtz: In here there is not much need on a default, the `oidc_client_id` only makes sense when used with the client_secret and issuer. Its basically all or none

server/charm/src/charm.py
rene-oromtz: There are different scopes, you can use both `NO_PROXY` and `BEHIND_PROXY` together as the first one specifies the resources to bypass proxy (as string) while the second is a boolean to enable ProxyFix. Perhaps a rename is needed to make this clearer (it is confusing as the NO_PROXY would appear as another boolean)

Perhaps changing `BEHIND_PROXY` to `ENABLE_PROXYFIX`?

server/charm/charmcraft.yaml
rene-oromtz: Same as this [comment](https://github.com/canonical/testflinger/pull/1111#discussion_r3337535534) no need to set a default given we need all OIDC parameters or None

server/src/testflinger/application.py
rene-oromtz: Added the note in [3c3bac9](https://github.com/canonical/testflinger/pull/1111/commits/3c3bac9d0974fb822fd3ffd33e8f000fcf6cf5d8)

server/charm/tests/unit/test_config.py
rene-oromtz: Added in [3c3bac9](https://github.com/canonical/testflinger/pull/1111/commits/3c3bac9d0974fb822fd3ffd33e8f000fcf6cf5d8)

server/charm/tests/unit/test_config.py
rene-oromtz: Added in [3c3bac9](https://github.com/canonical/testflinger/pull/1111/commits/3c3bac9d0974fb822fd3ffd33e8f000fcf6cf5d8)

server/charm/src/charm.py
rene-oromtz: For now, I renamed the charm option to `ENABLE_PROXYFIX` in [ac76eb2](https://github.com/canonical/testflinger/pull/1111/commits/ac76eb259f86a0e31d526a7e92c67f5ac7b534f5) given I think this is clear on intent. 

Let me know what you think and if you have any other preference, I'm open to suggestions as well.

ajzobro: Sure, that's seems appropriate and relevant for this PR.

ajzobro (approved):
LGTM

PR #1123, opened 2026-05-29
feat(server-charm): support ingress interface and remove traefik_route

rene-oromtz:
## Description
This follows a similar approach recommended in this [PR](https://github.com/canonical/hexr/pull/1135) by replacing `traefik_route` with `ingress`

> With ISD273, when an ingress (v2) relation is present on a k8s substrate, ingress-configurator auto-creates a NodePort Service, reads node InternalIPs, and publishes (node_ip, nodeport) pairs to IS-managed haproxy via haproxy-route. TLS terminates at IS haproxy; the backend hop is plain HTTP.

The current pattern is preserved, charm can support `nginx-ingress-integrator` or `ingress-configurator` but not both. 

This also makes changes to integration tests so it can be used for end to end integration tests:
1. `server-tox`: needed to remove the dump logs as we need two models which seems not supported by `pytests-jubilant`
2. `concierge.yaml`: Created so it can also bootstrap lxd

## Resolved issues
Resolves [CERTTF-1137](https://warthogs.atlassian.net/browse/CERTTF-1137)

## Tests
For a full end to end integration test, this needs an additional `machine` model. This is because the `haproxy` charm only exists as a `machine` charm. Without the `haproxy` charm, the `ingress-configurator` never fires the `on_ingress_ready` event. 

For doing this, we need to bootstrap both microk8s and lxd and create a temporary model for each. While this makes the setup more complex, it really reflects the `ingress-configurator` expected behavior. 

> [!NOTE]
> Given we are no longer using the pytest-jubilant juju fixture, I have removed juju-dump-logs from integration tests. I opened an [enhancement](https://github.com/canonical/pytest-jubilant/pull/67) upstream to add support for different cloud/controllers if that gets approved, I'll reformat to use that juju_factory and return that juju-dump-logs

[CERTTF-1137]: https://warthogs.atlassian.net/browse/CERTTF-1137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/charm/tests/integration/conftest.py
```diff
@@ -16,12 +16,48 @@
...
+        collect_juju_logs(request, juju)
+
+
+@pytest.fixture(scope="module")
+def machine_juju(request: pytest.FixtureRequest):
+    """Create temporary machine model."""
+    with jubilant.temp_model(controller=MACHINE_CONTROLLER) as juju:
+        juju.wait_timeout = 600
+        yield juju
+        collect_juju_logs(request, juju)
```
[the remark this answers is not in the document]
rene-oromtz: Marking as fail instead

server/charm/tests/integration/conftest.py
```diff
@@ -16,12 +16,48 @@
...
+
+
+@pytest.fixture(scope="module")
+def k8s_juju(request: pytest.FixtureRequest):
+    """Create temporary K8s model for running tests."""
+    with jubilant.temp_model(controller=K8S_CONTROLLER) as juju:
+        juju.wait_timeout = 600
+        yield juju
+        collect_juju_logs(request, juju)
+
```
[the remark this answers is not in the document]
rene-oromtz: Marking as fail instead

.github/workflows/server-tox.yml
```diff
@@ -106,9 +106,13 @@ jobs:
...
-        run: sudo concierge prepare --verbose --juju-channel=3/stable --charmcraft-channel=latest/stable -p microk8s
-      - name: Enable MicroK8s ingress
-        run: sudo microk8s enable ingress
+        run: sudo concierge prepare --verbose
+      - name: Export Juju controller names
+        run: |
+          K8S_CONTROLLER=$(juju controllers --format=json | jq -r '.controllers | to_entries[] | select(.value.cloud == "microk8s") | .key')
+          LXD_CONTROLLER=$(juju controllers --format=json | jq -r '.controllers | to_entries[] | select(.value.cloud == "localhost") | .key')
+          echo "JUJU_K8S_CONTROLLER=$K8S_CONTROLLER" >> "$GITHUB_ENV"
+          echo "JUJU_MACHINE_CONTROLLER=$LXD_CONTROLLER" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
rene-oromtz: `jq` is guaranteed by `concierge` the other check could be useful

.github/workflows/server-tox.yml
```diff
@@ -119,11 +128,7 @@ jobs:
           if-no-files-found: error
           retention-days: 2
       - name: Run charm integration tests
-        run: uvx --with tox-uv tox -e integration -- --juju-dump-logs logs
-      - name: Upload logs
-        if: ${{ !cancelled() }}
```
ajzobro, line 124: Do we not want logs?

server/charm/src/charm.py
```diff
@@ -128,138 +129,78 @@ def __init__(self, *args):
             self.on.retry_key_rotation_action,
             self._on_retry_key_rotation_action,
         )
-        # TODO: Remove nginx route when migration to traefik route is completed
```
ajzobro, line 131: Should we remove, or make configurable? Consider what environments we want our published charm to make available to the community.

server/charm/src/charm.py
```diff
@@ -128,138 +129,78 @@ def __init__(self, *args):
...
-    def _on_traefik_ready(self, event: TraefikRouteRequirerReadyEvent) -> None:
-        """Handle TraefikRouteRequirer ready event."""
-        if not self._check_route_conflict():
+    def _on_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
+        """Handle IngressPerAppReadyEvent ready event."""
+        if self._has_ingress_conflict():
             return
-        if self.unit.is_leader():
-            self._configure_traefik_route()
+        logger.info("Ingress is ready with URL: %s", event.url)
```
ajzobro, line 173: is the port number part of the important or interesting information?

ajzobro (commented):
I have a couple questions.

.github/workflows/server-tox.yml
rene-oromtz: Logs will still be available from the CI run but won't be downloadable. This is one of the disadvantages of not using the juju fixture from `pytest-jubilant` that we can't use directly the `--juju-dump-logs` flag. But on second thought, maybe I can replicate a similar behavior in `collect_juju_logs`. Right now this is only printing but I can probably store those logs as well and submit them as an artifact.

server/charm/src/charm.py
rene-oromtz: That is an interesting question given NGINX is already deprecated and not even recommended by Kubernetes project. While we are moving towards `ingress-configurator`, Kubernetes is recommending gateway API (which also has a charm). Perhaps we should remove NGINX and replace with gateway API? Definitively something worth discussing to decide on next steps

server/charm/src/charm.py
rene-oromtz: afaik is not necessary, we only need the URL, ingress configurator will decide if using HTTP or HTTPS based on its relation with HAProxy charm

.github/workflows/server-tox.yml
rene-oromtz: I returned the log collection and submission as artifacts in [49f78d9](https://github.com/canonical/testflinger/pull/1123/commits/49f78d9b5a2e1ae7ae6c0388432b69a3eb8acd21)

server/charm/src/charm.py
ajzobro: I see no reason at this time to add something new, I was simply trying to capitalize on the existing code and config that was in place for NGINX

ajzobro (dismissed):
LGTM

ajzobro (approved):
Let's try that merge again :-)

PR #1126, opened 2026-06-02
Fix: IoT provisioning must define test_data issue

kiya956:
## Description
We default use test_data username/password, while test_data is not defined, fallback to use provision_data username/password not just "ubunut"/"ubuntu"

## Resolved issues
Solving wrong userame/password while test_data is not defined.

## Tests
1. provision imx8mp-pdk with test_data is defined and verified username/password is correct
2. provision imx8mp-pdk without test_data and verified that we fallback to use username/password in provision_data


kiya956:
@p-gentili ready for review

p-gentili (approved):
LGTM

p-gentili:
@kiya956 you need to sign your commits :D

kiya956:
> @kiya956 you need to sign your commits :D

foget that everytime. T T

PR #1144, opened 2026-06-11
Add workshop development files

ajzobro:
### Workshop Development Environment

Along with the other tools, you can use Canonical's `workshop` to develop and
test Testflinger services in a container environment. Consider the following:

#### Workshop Basics

See the [`workshop` documentation](https://ubuntu.com/workshop/docs/) for more details.

```shell
# See the available workshops for a given project, add `--global` for all known workshops.
workshop list

# Bring up a workshop, e.g. `dev` workshop:
workshop launch dev

# Execute commands in a workshop
workshop exec dev -- echo "hello world"

# Start a terminal session in a workshop
workshop shell dev
```

#### Testflinger dev workshop examples

The following examples demonstrate specific uses of `workshop` and `just` for
typical server development.

Note: the justfile recipes attempt to abstract some of the `workshop` functions
away from the developer for ease of use, but it is still important to understand
certain aspects about the testflinger services, the way that they are run in the
`workshop` container, and how those services can be exposed to the host system.

```shell
# At it's simplest: Start the workshop (if necessary), start the testflinger
# server and attempt to connect the service to the host's ports
just workshop serve
```

To allow for multiple `git` worktrees to be used for parallel development of
testflinger features, the workshops must be able to be `connected` and,
importantly, `disconnected` from the host so that the developer may readily
control which workshop is exposed at a given time.

```shell
# Disconnect the workshop container from the local host system.
just workshop disconnect

# Connect the workshop container's internal services to the local host syetem.
just workshop connect

# Disconnect and then teardown the internal testflinger services.
just workshop teardown

# To see what other commands are available try:
just workshop help

# To execute any just command in the workshop, simply use `workshop exec -- just <command>`:
# To follow testflinger server logs:
workshop exec dev -- just server::logs

# To populate the instance with some data:
workshop exec dev -- just server::populate

# For a large data set you could try:
workshop exec dev -- just server::populate \
        --agents 3000 \
        --jobs 500 \
        --queues 4500 \
        --advertised-queues 2

# This command will clean-up the workshop's docker environment. It is not
# implemented in the `server` justfile to avoid it ever being run by mistake
# on the host system.
just workshop docker-prune

```

To allow for testing in an environment where your web browser is able
to connect to a local development IdP (in the development case: dex) we
reommend setting the localhost to match `dex`, `testflinger-server` and also
`testflinger-metrics` via a /etc/hosts entry pointing to localhost.

```shell
# For convenience you can run the following on your hosts file using sudo:
sudo just server::set-up-hosts-file

# Or you can just do it directly:
echo "127.0.0.1 dex testflinger-server testflinger-metrics" | sudo tee -a /etc/hosts
```

## Resolved issues

N/A

## Documentation

CONTRIBUTING.md

## Web service API changes

N/A

## Tests

Yes, added `.workshop/test_workshop_actions.py` or simply `just workshop test`


.workshop/dev.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+      workshop run dev just --list --unsorted
+          List all just recipes.
+
+      Note: Ensure that `dex` and `testflinger-server` point to localhost in your
+            /etc/hosts file, e.g. use `just run set-up-hosts-file` on the host.
+
+    EOF
+  opencode: opencode "$@"
+  claude: claude "$@"
```
ajzobro:
```suggestion
```

.gitignore
```diff
@@ -19,6 +19,7 @@ docs/.sphinx/.doctrees/
 *.bak
 *.charm
 *.swp
+*.lock
```
pedro-avalos: This should probably be `.workshop.lock` or maybe `*.workshop.lock`, but I don't think we will have that many  workshops for the time being.

.workshop/dev.yaml
```diff
@@ -0,0 +1,68 @@
...
+  just: just "$@"
+  server-up: just server-up
+  server-restart: just server-restart
+  server-down: just server-down
+  server-logs: just server-logs
+  docker-prune: docker system prune -a --yes
+  add-sample-data: just add-sample-data
```
pedro-avalos: These have been added as server just recipes, so this needs to be updated. Honestly, though, I think we should remove them and expect users to use `workshop run just server::serve` etc. instead of maintaining a somewhat-duplicated command list here

.workshop/dev.yaml
```diff
@@ -0,0 +1,68 @@
...
+      workshop run dev sample-data-custom -s http://localhost:5000 -a 10 -j 10 -q 10 -d 1
+          Run create_sample_data.py with custom arguments.
+
+      workshop run dev just --list --unsorted
+          List all just recipes.
+
+      Note: Ensure that `dex` and `testflinger-server` point to localhost in your
+            /etc/hosts file, e.g. use `just run set-up-hosts-file` on the host.
+
+    EOF
```
pedro-avalos: I am not sure I like this hard-coded help... Between `workshop actions` and `workshop run just help` I think we should be okay, and we can document these two command usages in the contributing guide

justfile
```diff
@@ -141,6 +141,59 @@ schema:
...
+    set -euo pipefail
+    # When `dex` not found, add dex and testflinger-server as hostnames
+    LINE="127.0.0.1 dex testflinger-server"
+    if ! grep -Eq '(^|[[:space:]])dex($|[[:space:]])' /etc/hosts; then
+        if ! sudo -n true 2>/dev/null; then
+            echo "sudo credentials required; run: sudo -v"
+            exit 1
+        fi
+        echo "$LINE" | sudo -n tee -a /etc/hosts >/dev/null
+    fi
```
rene-oromtz: Given this is a one time change, do we want a recipe for it? I think its already in the contributing guide or the hacking, so perhaps that is more than enough? What do you think?

justfile
```diff
@@ -141,6 +141,59 @@ schema:
...
+    cd 'server'
+    docker compose down --remove-orphans
+
+[doc('Follow the local server container logs, reconnecting on restart.')]
+server-logs:
+    #!/usr/bin/env -S bash -e
+    while true; do
+        docker logs -f testflinger-server || true
+        sleep 1
+    done
```
rene-oromtz: This can have some value, but given the structure changed, this needs to be on the server recipes

justfile
```diff
@@ -141,6 +141,59 @@ schema:
...
+server-up:
+    #!/usr/bin/env -S bash -e
+    cd 'server'
+    docker compose up -d --build
+
+[doc('Rebuild and restart only the local testflinger server container.')]
+server-restart:
+    #!/usr/bin/env -S bash -e
+    cd 'server'
+    docker compose up -d --build --no-deps testflinger
```
rene-oromtz: As far as I remember, the server automatically refresh on any change (or at least should) so perhaps this is not entirely needed?

.workshop/dev.yaml
rene-oromtz: This needs to be updated to match the latest recipes, but if I understand correctly, the names for the "commands" are arbitrary, correct? If that's the case, I think only the just commands needs to be updated

rene-oromtz (changes requested):

justfile
rene-oromtz: Oh nvm, I think this is mostly required for the workshop, so this will also needs to be moved to the server recipes

pedro-avalos: Is this a setup that we can move into an [in-project SDK](https://ubuntu.com/workshop/docs/tutorial/part-3-sketch-sdks/#convert-to-in-project-sdk) [hook](https://ubuntu.com/workshop/docs/explanation/sdks/runtime-hooks/), then instead of a recipe? Maybe `setup-base`?

justfile
pedro-avalos: You might just be able to do `docker compose logs --follow` instead of this loop

server/justfile
```diff
@@ -88,3 +97,28 @@ admin:
...
+    set -euo pipefail
+    # When `dex` not found, add dex and testflinger-server as hostnames
+    LINE="127.0.0.1 dex testflinger-server"
+    if ! grep -Eq '(^|[[:space:]])dex($|[[:space:]])' /etc/hosts; then
+        if ! sudo -n true 2>/dev/null; then
+            echo "sudo credentials required; run: sudo -v"
+            exit 1
+        fi
+        echo "$LINE" | sudo -n tee -a /etc/hosts >/dev/null
+    fi
```
pedro-avalos: Given this runs `sudo`, I'd say it should have a confirmation prompt, maybe something like: `[confirm('Modify /etc/hosts for local development?')]`

server/justfile
```diff
@@ -88,3 +97,28 @@ admin:
...
+
+[doc('Create a large pre-canned sample data set on the local server.')]
+[group('deploy')]
+populate-quick:
+    @uv run devel/create_sample_data.py \
+        -s http://localhost:5000 \
+        -a 3000 \
+        -j 500 \
+        -q 4300 \
+        -d 2
```
pedro-avalos: I'd recommend relabeling the recipe, rephrasing to match `populate` description, and using `populate` recipe here:

```suggestion
[doc('Populate the server with a large sample data set.')]
[group('deploy')]
populate-large:
    @just populate
        --agents 3000 \
        --jobs 500 \
        --queues 4300 \
        --advertised-queues 2
```

Since `-s` defaults to `localhost:5000`, I'm not sure defining here is needed.

Moreover, I'd say either keep it all in one line if you're using the short flags OR keep it an argument per line with the long options for readability

.gitignore
pedro-avalos, line 1: Looks like you got rid of the workshop lockfile from the .gitignore, could you add it as `.workshop.lock` for now?

.workshop/dev.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+actions:
+  just: just "$@"
+  serve: just server::serve
+  restart: just server::restart
+  teardown: just server::teardown
+  logs: just server::logs
+  docker-prune: docker system prune -a --yes
+  populate-quick: just server::populate-quick
+  populate: just server::populate "$@"
```
pedro-avalos: I'm still not sold on having these `just` recipes in the actions if we provide `just "$@"` as an action.

I think we can leave this as:

```yaml
actions:
  just: just "$@"
  docker: docker "$@"
```

CONTRIBUTING.md
```diff
@@ -67,6 +67,29 @@ source .venv/bin/activate
...
+      workshop run dev populate -s http://localhost:5000 -a 10 -j 10 -q 10 -d 1
+      workshop run dev populate-quick
+
+      workshop run dev docker-prune
+          Aggressively remove unused Docker data in the workshop.
+          
+      Note: Ensure that `dex` and `testflinger-server` point to localhost in your
+            /etc/hosts file, e.g. use `just server::set-up-hosts-file` on the host.
+```
+
```
pedro-avalos: For consistency with the other shell code blocks, let's do the following:

- Add `shell` to the code block
- Use `#` to denote comments (i.e., don't break commands if users try to copy-paste these)
- See earlier comments about actions, I think we should provide bare minimum actions to avoid duplicating commands/recipes

    ```shell
    # See the available workshops
    workshop list

    # Bring up a workshop
    workshop launch dev

    # Execute commands in a workshop
    workshop exec dev -- echo "hello world"

    # Start a terminal session in a workshop
    workshop shell dev

    # See pre-defined actions
    workshop actions dev

    # Run a pre-defined action
    workshop run dev just help

    # Aggressively remove unused Docker data in the workshop
    workshop run dev docker system prune -a --yes

    # NOTE: Ensure `dex` and `testflinger-server` point to localhost in `/etc/hosts` on the host
    just server::set-up-hosts-file
    ```

server/justfile
ajzobro: Specifying a value that is correct for the setup (and will always be correct for *this* setup) is proper and not excessive.  It is more brittle to depend on a default value.

.gitignore
ajzobro: I didn't intend that and have fixed it on my end; I did go with *.lock because until there is a valid reason to add a given .lock file, I think it's a safe choice.

CONTRIBUTING.md
```diff
@@ -67,6 +67,29 @@ source .venv/bin/activate
 
 To learn more about `uv`, refer to the [`uv` documentation][uv].
 
+### Workshop Environment
+
+Along with the other tools, you can use Canoncial `workshop` to develop and
+validate testflinger services in a container environment. Consider the following
+
+#### Testflinger dev workshop actions:
+```
```
ajzobro, line 105:
```suggestion
```shell
```

CONTRIBUTING.md
ajzobro: These seem unnecessary as they do not relate to this project's necessary/normal flow; These are teaching workshop and we should probably just link to workshop documentation instead.
> workshop list
> workshop exec dev -- echo "hello world"
> workshop shell dev
> workshop actions dev

This one ceases to be a handy recipe that is preserved and intentionally inside the workshop; there is now a chance of a user running `docker system prune -a --yes` on their local machine (through faulty memory, understanding or even copy-paste), which could be catastrophic if they have a local docker instance. 
The proposed workshop action ensures this command will only ever be run inside the workshop.
> workshop run dev docker system prune -a --yes

I struggle to understand the benefit of:
`workshop run just <blah>` vs `workshop exec just <blah>` -- there is no difference that I can tell: creating  a workshop action that is the same name as the function it calls plus passing all the arguments to that function is rather silly, though possible.

Originally, workshop actions were added to capture the things that would be useful and done in the workshop -- even though these are repeats of what is in the `justfile` which in some ways repeats what's in the tox file, etc. The scope being the workshop, the most-common actions were provided.

------------------------------------------------------------------------

I will also state that this isn't shaping up to be as wonderful as originally hoped.  If you work in one git checkout (same folder) always, this can be useful in its current state -- **because** if those ports are available everything works.  However the `system` sdk in workshop will immediately try to bind ports 5000 and 5556 on `workshop start` and will only release on `workshop stop`.

This will cause problems like:
```
$ workshop start
error: cannot perform the following tasks:
- Start "dev" workshop (Error occurred when starting proxy device: Error: Failed listening on 127.0.0.1:5556: listen tcp 127.0.0.1:5556: bind: address already in use)
```

Having multiple worktrees where only one can be "used" at a time reduces the usefulness of workshop + worktrees immensely.  `workshop list --global` becomes necessary to identify what other `dev` instances might be running.

Having only one workshop (`dev`) implementing all SDKs (e.g. including `opencode`) makes it so that I can only use `opencode` in the one worktree and the one workshop that I am able to run.

At present, my global workshop list looks like this:

```
$ workshop list --global
PROJECT                                                                  WORKSHOP       STATUS   NOTES
~/projects/testflinger/worktrees/chore/add-workshop                      dev            Ready    -
~/projects/testflinger/worktrees/feature/api-lockdown-when-oidc-enabled  ai             Ready    -
~/projects/testflinger/worktrees/feat/update-api-docs-generation         ai             Ready    -
~/projects/testflinger/worktrees/docs/endpoint-roles-table               ai             Ready    -
~/projects/testflinger/worktrees/fix/harden-against-injection            ai             Ready    -
~/projects/testflinger/worktrees/feat/metrics-own-port                   dev            Stopped  -
~/projects/testflinger/worktrees/docs/auth-flows                         ai             Ready    -
```

Need to rethink the whole thing.

.workshop/dev.yaml
ajzobro: Here's why I didn't propose that solution:

The main point of this workshop is to provide a standard, consistent, and easy to use development area that is isolated from the developer's own workspace.

As an individual who will be using the `workshop` to launch and serve the server code for testing, I won't ever be using the `just` commands related to server launching and won't have them memorized in favor of the short-form `workshop run serve` over `workshop run just server::serve`.

Similarly, I don't love the `workshop run X <options>` because it's not different from `workshop exec X <options>`

server/justfile
ajzobro: I opted to make any aspect of sudo happen overtly and now the expectation will be that the user call `sudo just set-up-hosts-file`, this is proper as the caller of the helper function very well may have the permissions to change that file without sudo and so it is not our job to guess at their permissions.

This is documented as `sudo just set-up-hosts-file`

CONTRIBUTING.md
ajzobro: I updated this file, but did so by considering many other options and accepting much of the proposed suggestions.

.gitignore
ajzobro: If you insist, I will, but I stand by the thought that there is no valid reason to every commit a *.lock file and when there is, that development activity will re-assess this match pattern.

ajzobro: Ah, I see now, you're pointing out that we do have many `uv.lock` today and it's valid to suggest to users that `.lock` files be added in the future as well. That's my mistake and oversight, will fix.

.workshop/justfile
```diff
@@ -0,0 +1,59 @@
...
+    fi
+
+[doc('Clean up docker files in the workshop.')]
+[group('workshop')]
+docker-prune:
+    workshop exec dev -- docker system prune -a -f
+
+[doc('Start the server in the workshop and connect it to the host.')]
+[group('server')]
+serve: start && connect
```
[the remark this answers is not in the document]
ajzobro: valid for post; runs `start` first, then `serve`, then `connect`

.workshop/dev.yaml
ajzobro: FYI< I mostly did this, please re-review.

.workshop/test_workshop_actions.py
rene-oromtz, line 1: Given this file can also potentially be used in CI, can you please create a Jira issue so we can track this? I don't think its necessary now but just so we keep an eye on this.

CONTRIBUTING.md
```diff
@@ -67,6 +67,98 @@ source .venv/bin/activate
...
 To learn more about `uv`, refer to the [`uv` documentation][uv].
 
+### Workshop Development Environment
+
+Along with the other tools, you can use Canonical's `workshop` to develop and
+test Testflinger services in a container environment. Consider the following:
+
+#### Workshop Basics
+
+See the [`workshop` documentation][workshop documentation] for more details.
```
rene-oromtz, line 91: Perhaps in this workshop basics we can add how can a user install workshop and adding a note of the prerequisited (LXD 6/stable)?

Given all other dev tools (uv, just) are showing a quick an easy way to install, perhaps we can do the same with workshop. 

I'm thinking:
1. Install workshop via snap...
2. note: you require LXD 6...

What do you think?

rene-oromtz (commented):
Just added two small comments, overall LGTM!

.workshop/test_workshop_actions.py
ajzobro: Yeah, good idea. CERTTF-1352

CONTRIBUTING.md
ajzobro: I can add that.  I'm considering just ripping the instructions straight from workshop docs...

rene-oromtz (approved):
Thanks for the changes! LGTM

.workshop/dev.yaml
```diff
@@ -0,0 +1,29 @@
...
+    plugs:
+      tcp5000:
+        interface: tunnel
+        endpoint: localhost:5000  # on the host
+      tcp9090:
+        interface: tunnel
+        endpoint: localhost:9090  # on the host
+      tcp5556:
+        interface: tunnel
+        endpoint: localhost:5556  # on the host
```
pedro-avalos, line 29: I'm okay with this, but what's the reason for not naming them `testflinger-server`, `testflinger-metrics`, and `dex-server`?

server/justfile
```diff
@@ -89,3 +98,26 @@ admin:
...
+[group('deploy')]
+populate-large:
+    @just populate \
+        --agents 3000 \
+        --jobs 500 \
+        --queues 4500 \
+        --advertised-queues 2
+
+[doc('Add required hostnames to the hosts file to ensure that `dex` and `testflinger-server` and `testflinger-metrics` are available for local testing.')]
+[group('deploy')]
```
pedro-avalos, line 112: I think this should have a

`[confirm('Update /etc/hosts for development?')]`

pedro-avalos (approved):
Looks good. I left some nitpicks, but they are non-blocking

.workshop/dev.yaml
ajzobro: Names for the docker slots cannot align with the name for the system plugs or else workshop will attempt to connect them on launch/start and so if there is another workshop with them already connected (e.g. another git worktree) then this new workshop cannot be stood up -- preventing it from being used for anything, even if not needed to be connected to the host.

This setup intentionally renames the system plugs so that they are reflective of the system resources and not the the hosted services to allow for them to be manually connected to a given workshop instance without hindering that workshop (other workshops) from being able to be started and used.

The workshop environments don't only exist for the purpose of connecting the testflinger server to localhost ports for testing and so blocking the workshop start-up was undesirable.

Let me know how much of this you think should be incorporated into comments and where you would like to see those comments.

server/justfile
ajzobro: I don't.

1) `just`'s confirmation is annoying, simple `enter` doesn't work; `y` is expressly needed and never declared, one must guess their way there.
2) it does something funny to the console (wrapping in python exacerbated this and turned me off to this feature even more)

But mostly:

1) if a user tries to do this they will either: succeed or fail with a relevant error message (permission denied, only if changes were needed)

2) sudo will already prompt them if necesssary

Thoughts?

ajzobro (commented):
Responses provided, Please let me know what further changes you would like to see.

server/justfile
pedro-avalos: Yeah, but sudo will not always prompt. Say a user does something like:

```shell
sudo apt update
sudo apt upgrade
sudo apt install ... # install some dep for testflinger, or whatever
just ... # no prompt for sudo
```

But I think since the recipe name is pretty explicit and documented, the confirm may be okay (as opposed to charm setup, which can be destructive, for example, or some CI-specific recipes that should be private and confirm-gated).

pedro-avalos (approved):
~~Looks good. I left some nitpicks, but they are non-blocking~~

EDIT: GH seems to have saved my last comment; meant to make this a blank approve lol

server/justfile
ajzobro: Yes, of course, you are correct that previously run sudo commands will allow for executing the just  command without re-typing the sudo password.

However, please take another look at the `just` recipe: it no longer contains the `sudo` call itself, so they would have to do `sudo just set-up-hosts-file` to affect their `/etc/hosts` file.

PR #1146, opened 2026-06-12
new: flag to indicate whether the agent will have access to the DUT after provisioning

p-gentili:
## Description

There are a couple of cases where the provisioned image will NOT provide testflinger SSH access to the DUT.
- images w/o a SSH server, intentionally
- images w/ a system-user assertion built-in, intentionally made w/o the agent keys

## Resolved issues

N/A

## Documentation

On the server side

## Web service API changes

No

## Tests

The code change is limited, tested only via unit testing.


rene-oromtz (approved):
LGTM, just consider that this require server to be promoted to stable. Right now all in beta is properly working on staging but prod needs to get this [PR](https://github.com/canonical/certification-ops/pull/515) merged as its defining which Test Observer address to use (and the authentication)

Perhaps we could see if we can deploy this next week once we land all changes in Prod

PR #1147, opened 2026-06-12
feat(server): copy job YAML, Job YAML tab, syntax-highlighted job definition

mz2:
## Overview

Adds a **"Copy job YAML"** button and a clearer job-definition view to the dashboard job detail page. The stored `job_data` is reconstructed into a submittable job-definition YAML (the `Job` schema top-level fields, in schema order), shown in a default **Job YAML** tab and copyable to the clipboard in one click.

### New: Job YAML tab + Copy button
- **Server-side YAML rendering** (`views.py`): `build_job_yaml()` rebuilds a submittable definition from the stored `job_data`, with multiline strings (e.g. the test script) rendered as literal `|` blocks.
- A new **Job YAML** tab (the full submittable definition) is shown by default.
- **Copy button** (`clipboard.js`): a generic `[data-copy-target]` handler copies the clean, unhighlighted YAML from a hidden payload and shows brief "Copied!" feedback.

### Changed: how the existing Provision / Test tabs are presented

These tabs already existed but rendered their content awkwardly. This PR fixes the presentation:

- **Provision Data** previously rendered the raw Python `dict` repr — e.g. `{'distro': 'jammy'}` — because the template emitted `{{ job.job_data.provision_data }}` directly. It now renders as **YAML** (`distro: jammy`) via a new `as_yaml` Jinja filter. **Rationale:** the underlying job definition *is* YAML, so showing a Python repr was both inconsistent with how jobs are authored and harder to read for nested data; rendering YAML also matches the new Copy-job-YAML output exactly.
- **Test Data → "Test Script"**: this tab only ever shows `test_data.test_cmds`, which is a shell script — not a generic data structure. The tab was renamed from "Test Data" to **"Test Script"** to say what it actually is, and the script is now **bash syntax-highlighted**. **Rationale:** "Test Data" was misleading (it implied the whole `test_data` mapping); "Test Script" is accurate, and highlighting a multi-line shell script makes it far easier to scan.

### Syntax highlighting
- Highlighting is done **server-side with Pygments** (no client-side JS): YAML for the definition sections, bash for the test script. A small generated `pygments.css` is loaded only on this page. Output is XSS-safe — Pygments HTML-escapes the content.

### Local-dev conveniences
- `docker-compose.yml` bind-mounts `src/` and sets `TEMPLATES_AUTO_RELOAD` (env-guarded, default off, in `application.py`) for hot-reload without rebuilds; `create_sample_data.py` seeds representative `provision_data`.

## Screenshots

**Job YAML tab (new, shown by default)** — full submittable definition, syntax-highlighted, with the Copy button on the tab row:

testflinger-copy-job-yaml

**Provision Data tab** — now rendered as YAML (`distro: jammy`) instead of a Python repr:

testflinger-provision-tab

**Test Script tab** — renamed from "Test Data" (it always was including just the test_cmds), with bash syntax highlighting:

testflinger-test-tab

## Testing

- Unit tests (`tests/test_views.py`): `build_job_yaml`, the `as_yaml` and `highlight` filters (including HTML-escaping), the copy button/payload, and that sections render as YAML rather than a Python repr.
- Browser end-to-end coverage (Playwright) drives the real copy interaction and verifies the clipboard contents match the submittable YAML.

No API changes, database migrations, or new endpoints. `TEMPLATES_AUTO_RELOAD` is development-only and defaults off.


server/src/testflinger/static/assets/js/clipboard.js
ajzobro, line 1: However, the copy button does not copy the yaml into either of my paste buffers (Ctrl + V and middle-click), I also never saw the "Copied!" span, so I think it's just not linked up. I used Chrome for testing.

server/src/testflinger/templates/job_detail.html
```diff
@@ -30,75 +39,77 @@ <h2 class="p-muted-heading" id="job-definition">
...
+          <div class="p-tabs__item">
+            <button class="p-tabs__link"
+                    role="tab"
+                    aria-selected="false"
+                    aria-controls="test-data"
+                    id="test-data-tab">Test Script</button>
+          </div>
+        {% endif %}
+      </div>
+      <button class="p-button--base has-icon"
```
ajzobro, line 81: With respect to the the button, I think it's pushing the tab area over to the left and leaving a large unused space below it.

Image

ajzobro (changes requested):
The page looks great!

server/src/testflinger/templates/job_detail.html
rene-oromtz: Yeah, I think its because the margin-bottom in CSS: https://github.com/canonical/testflinger/blob/86bfdf0e5ea9ce681b190a8b09c54bb4e9e1a179/server/src/testflinger/static/assets/css/testflinger.css#L127

I was playing around and setting it to `0rem` seems to at least not display that white border and it seems more inline with the tabs

server/pyproject.toml
```diff
@@ -20,6 +20,7 @@ dependencies = [
     "owasp-logger>=0.1.5",
     "prometheus-client>=0.21.0",
     "prometheus-flask-exporter>=0.23.1",
+    "pygments>=2.18.0",
```
rene-oromtz: Can you please specify: Pygments>=2.20.0? While a low sev vulnerability this is to address https://github.com/canonical/testflinger/security/dependabot/62
Given we are now explicitly including pygments, we can start with the recommended version

```suggestion
    "pygments>=2.20.0",
```

server/src/testflinger/static/assets/js/clipboard.js
rene-oromtz: Hmmm I tried Firefox and Chromium and I do see the Copied! and able to paste it (only tested with Ctrl + V) not sure if its something in the code needed or perhaps the local environment not loading properly?

rene-oromtz (commented):
I really like how its getting displayed in the job view! Thanks for taking care of this! I just added a comment regarding pygments that might as well do it now given we are explicitly including it as dependency.

server/src/testflinger/static/assets/js/clipboard.js
mz2: I am not able to reproduce the problem either, tested both Firefox and Chrome. Could this be some Wayland session, XWayland based browser thing? I don't have a precise description for what I mean but I have sometimes got GNOME into a state where clipboard only works from Wayland based apps to Wayland based apps, and between XWayland to XWayland (again, not 100% sure this is the pattern). @ajzobro can you try in a fresh GNOME session to rule out a desktop bug?

server/pyproject.toml
rene-oromtz: @mz2 I forgot this will require: `just server lock` to update the lock file

server/src/testflinger/static/assets/js/clipboard.js
ajzobro: Honestly it seems to be happening more and more today (copy paste not working with Chrome) -- I closed a bunch of tabs and that seems to have helped; I can certainly try another browser/reboot/all the things.

server/src/testflinger/templates/job_detail.html
ajzobro: turns out it's the button style that we get :-(
Non-issue IMO, let's move past this. If it's really bothersome we can adjust things later.

rene-oromtz (dismissed):
LGTM

rene-oromtz (approved):

ajzobro (approved):

PR #1149, opened 2026-06-14
Fix stale docs URL in CLI auth warning

### LINKED ISSUE #961: Update authorization warning; points to page no longer available.
`testflinger-cli submit --poll reserve_manta.yaml `
Received 401 error from server with reason: 
```text
Authentication required for setting timeout: 172800
You are attempting to use a feature that requires client authorisation without using client credentials. 
See https://testflinger.readthedocs.io/en/latest/how-to/authentication.html for more details
```

Consider pointing to either of:

- https://canonical-testflinger.readthedocs-hosted.com/latest/how-to/authentication/
- https://canonical-testflinger.readthedocs-hosted.com/latest/explanation/authentication/

Note: there is no explaination of how long the "cached" credentials will be cached and the two pages don't link both ways to eachother.

mvanhorn:
## Description

When the CLI returns a 401 because a feature requiring client authorisation was used without client credentials, it prints a help link pointing at `https://testflinger.readthedocs.io/...`, which is no longer the live docs host. This updates the warning to `https://canonical-testflinger.readthedocs-hosted.com/en/latest/how-to/authentication/`, matching the host already used elsewhere in the repo (server/agent charms and the `submit` GitHub action). The path is unchanged; only the stale host is corrected.

## Resolved issues

Fixes #961

## Documentation

No documentation changes. This points an existing user-facing message at the live docs page rather than the stale one.

## Web service API changes

None.

## Tests

`grep -rn "testflinger.readthedocs.io" cli/testflinger_cli/` returns no matches after the change, and the concatenated URL resolves to the live `canonical-testflinger.readthedocs-hosted.com` authentication page. The string is not asserted by the existing CLI suite, so no test changes are needed.

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -1316,7 +1316,11 @@ class TestflingerCli:
                     f"{exc.msg}\n"
                     "You are attempting to use a feature "
                     "that requires client authorisation "
-                    "without using client credentials."
+                    "without using client credentials. \n"
+                    "See "
+                    "https://documentation.ubuntu.com/"
+                    "testflinger/latest/ "
+                    "for more details"
                 )
             # This shouldn't happen, so let's get more information
             sys.exit(


cli/testflinger_cli/__init__.py
```diff
@@ -1324,7 +1324,8 @@ def submit_job_data(self, data: dict):
                             "You are attempting to use a feature "
                             "that requires client authorisation "
                             "without using client credentials. \n"
-                            "See https://testflinger.readthedocs.io/en/latest"
+                            "See https://canonical-testflinger."
+                            "readthedocs-hosted.com/en/latest"
```
ajzobro:
```suggestion
                            "See https://documentation.ubuntu.com/testflinger/latest/"
```

May have to add proper line breaks.  @tang-mm this is where they are now hosted, is this the correct url?

ajzobro (changes requested):

cli/testflinger_cli/__init__.py
tang-mm: At this moment, yes. However, we are currently in the middle of a second migration to the domain `ubuntu.com/docs/testflinger`, and I expect the new redirect to take place within weeks. Please continue to use the current URL for now. Once the new URL is stable, I will do a scan to update all references

mvanhorn:
Applied in 96647cd. The 401 auth warning now points to https://documentation.ubuntu.com/testflinger/latest/ as suggested. Thanks @tang-mm for confirming the current URL - makes sense to leave the ubuntu.com/docs/testflinger switch to your later sweep.

mvanhorn:
Applied — kept the `documentation.ubuntu.com/testflinger/latest/` URL @tang-mm confirmed as current, and cleaned up the string wrapping. Happy to switch to `ubuntu.com/docs/testflinger` once that migration/redirect lands.

mvanhorn:
Rebased onto main. The refactor in #1168 restructured `submit_job_data` and dropped the docs pointer from the 401 warning entirely, so the original stale-URL replacement no longer applied. This PR now re-adds the pointer to the refactored message using the current docs location (https://documentation.ubuntu.com/testflinger/latest/), which is what #961 was asking for. Single commit now, `ruff check` passes.

PR #1152, opened 2026-06-16
feat(server/cli): add IDP authentication via Device Authentication Flow

rene-oromtz:
## Description
This PR adds the Device authentication Flow to be leveraged by the CLI. 

If OIDC is configured server side (and if the provider supports the Device Flow) authentication with `--method oidc` is now supported. 

This PR:
1. Adds two need server endpoints under `/oidc` route. With this endpoints, the Testflinger server can act as a proxy and redirect the request to the IDP provider. 
2. `auth-init`: returns a unique request ID to the user instead of the `device-code` (server holds this data for further validation) along with other relevant information for user login via browser
3. `auth-poll`: uses the `request_id` and the previously stored `device-code` to forward the request to the IDP and validate if user is already authenticated
4. Upon successful authentication, user is issued Testflinger access and refresh tokens (NOT IDP tokens)
5. Refresh token is stored so it can be reused until expiration (6 days)

This also removes the previous web_clients collections in favor of storing all user permissions on client_permissions collection. 

## Resolved issues
Resolves [CERTTF-1196](https://warthogs.atlassian.net/browse/CERTTF-1196)

## Documentation
For review simplicity, I will add the documentation on IDP usage on a follow up PR

## Web service API changes
This introduces a new route through `/oidc/<endpoints>` instead of `/v1/<endpoint`> given OIDC is optional

## Tests
Tested locally:

https://github.com/user-attachments/assets/ecc96246-fff9-41f2-80a2-d7aad45b3847

Tested on staging:
```
uv run testflinger-cli --server $SERVER login --method oidc
Please visit https://xxxxxxx/oauth2/device/verify and enter code xxxxxx to login.
Successfully authenticated as 'xxxxxxx@canonical.com'
```

Validate with admin credentials:
```
testflinger-cli --server $SERVER admin get client-permissions --testflinger-client-id 'xxxxxxx@canonical.com'
{"client_id": "xxxxxxx@canonical.com", "role": "contributor"}
```

Grant permissions:
```
testflinger-cli --server $SERVER admin set client-permissions --testflinger-client-id 'xxxxxxx@canonical.com' --max-reservation '{"audino": 36000}'
Updated permissions for client 'xxxxxxx@canonical.com'
```

Submit with extended reservation (8hrs):
```
uv run testflinger-cli --server $SERVER submit -p ~/Desktop/Testflinger/staging.yaml 
Job submitted successfully!
job_id: 53b9087f-1a60-4291-a12b-406923cb004f
This job will be picked up after the current job is complete (it is next in line)
...
Current time:           [2026-06-17T19:19:47.906408+00:00]
Reservation expires at: [2026-06-18T03:19:47.906458+00:00]
Reservation will automatically timeout in 28800 seconds

```

[CERTTF-1196]: https://warthogs.atlassian.net/browse/CERTTF-1196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -149,6 +149,7 @@ class TestflingerCli:
             or os.environ.get("TESTFLINGER_ERROR_THRESHOLD")
             or consts.TESTFLINGER_ERROR_THRESHOLD
         )
+        auth_method = getattr(self.args, "method", "credentials")
 
         # Allow config subcommand without worrying about server or client
         if hasattr(self.args, "func") and self.args.func == self.configure:
@@ -160,7 +161,7 @@ class TestflingerCli:
             )
         self.client = client.Client(server, error_threshold=error_threshold)
         self.auth = TestflingerCliAuth(
-            self.client_id, self.secret_key, self.client
+            auth_method, self.client_id, self.secret_key, self.client
         )
 
     def run(self):
@@ -290,6 +291,17 @@ class TestflingerCli:
             help="Authenticate with server",
         )
         parser.set_defaults(func=self.login)
+        parser.add_argument(
+            "--method",
+            nargs="?",
+            choices=["credentials", "oidc"],
+            default="credentials",
+            help=(
+                "Authentication method to use. "
+                "'credentials' uses client-id and secret-key (default). "
+                "'oidc' opens a browser for IDP-based login."
+            ),
+        )
         self._add_auth_args(parser)
 
     def _add_poll_args_generic(self, parser):
@@ -1859,9 +1871,9 @@ class TestflingerCli:
         # Clear refresh token if user is attempting reauthentication
         self.auth.clear_refresh_token()
         try:
-            # Since refresh token was cleared, credentials are always needed
+            # Since refresh token was cleared, we need to re-authenticate
             if self.auth.authenticate():
-                print(f"Successfully authenticated as user '{self.client_id}'")
+                print(f"Successfully authenticated as '{self.auth.client_id}'")
             else:
                 # authenticate can return None if no credentials were provided
                 sys.exit("Please provide credentials and reattempt login")
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -161,7 +161,7 @@ class TestflingerCli:
             )
         self.client = client.Client(server, error_threshold=error_threshold)
         self.auth = TestflingerCliAuth(
-            auth_method, self.client_id, self.secret_key, self.client
+            self.client, auth_method, self.client_id, self.secret_key
         )
 
     def run(self):
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -149,7 +149,6 @@ class TestflingerCli:
             or os.environ.get("TESTFLINGER_ERROR_THRESHOLD")
             or consts.TESTFLINGER_ERROR_THRESHOLD
         )
-        auth_method = getattr(self.args, "method", "credentials")
 
         # Allow config subcommand without worrying about server or client
         if hasattr(self.args, "func") and self.args.func == self.configure:
@@ -161,7 +160,7 @@ class TestflingerCli:
             )
         self.client = client.Client(server, error_threshold=error_threshold)
         self.auth = TestflingerCliAuth(
-            self.client, auth_method, self.client_id, self.secret_key
+            self.client, self.client_id, self.secret_key
         )
 
     def run(self):
@@ -187,6 +186,7 @@ class TestflingerCli:
         parser.add_argument(
             "--server", default=None, help="Testflinger server to use"
         )
+        parser.set_defaults(client_id=None, secret_key=None)
         subparsers = parser.add_subparsers()
         self.admin_cli = TestflingerAdminCLI(subparsers, self)
         self._add_artifacts_args(subparsers)
@@ -213,6 +213,12 @@ class TestflingerCli:
         except SnapPrivateFileError as exc:
             parser.error(exc)
         self.help = parser.format_help()
+
+        # Early validation of credentials
+        if bool(self.args.client_id) != bool(self.args.secret_key):
+            parser.error(
+                "Both --client-id and --secret-key must be provided together"
+            )
 
     def _add_auth_args(self, parser):
         parser.add_argument(
@@ -291,17 +297,6 @@ class TestflingerCli:
             help="Authenticate with server",
         )
         parser.set_defaults(func=self.login)
-        parser.add_argument(
-            "--method",
-            nargs="?",
-            choices=["credentials", "oidc"],
-            default="credentials",
-            help=(
-                "Authentication method to use. "
-                "'credentials' uses client-id and secret-key (default). "
-                "'oidc' opens a browser for IDP-based login."
-            ),
-        )
         self._add_auth_args(parser)
 
     def _add_poll_args_generic(self, parser):
--- a/cli/testflinger_cli/auth.py
+++ b/cli/testflinger_cli/auth.py
@@ -18,7 +18,6 @@
 
 import configparser
 import contextlib
-import sys
 import time
 import webbrowser
 from functools import cached_property, wraps
@@ -34,6 +33,7 @@ from testflinger_cli.errors import (
     AuthenticationError,
     AuthorizationError,
     InvalidTokenError,
+    OidcError,
 )
 
 
@@ -43,11 +43,9 @@ class TestflingerCliAuth:
     def __init__(
         self,
         tf_client: client.Client,
-        method: str = "credentials",
         client_id: str | None = None,
         secret_key: str | None = None,
     ):
-        self.method = method
         self.client_id: Optional[str] = client_id
         self.secret_key = secret_key
         self.client = tf_client
@@ -75,20 +73,29 @@ class TestflingerCliAuth:
         :raises InvalidTokenError: Refresh token already expired
         :return: string with access token (jwt token)
         """
-        # If OIDC method is selected, initialize OIDC device auth flow
-        if self.method == "oidc":
-            return self._authenticate_with_oidc()
-
-        # Authenticate with credentials if provided in args or env file
+        # 1. Attempt to authenticate with credentials from args or envvars
+        # Higher priority to honor users intent when credentials present
         if self.client_id and self.secret_key:
             return self._authenticate_with_credentials()
 
-        # Authenticate with refresh token if available
+        # 2. Authenticate with refresh token if available
+        # This covers persistent login for both credentials and OIDC logins
         refresh_token = self.get_stored_refresh_token()
+        token_expired: InvalidTokenError | None = None
         if refresh_token:
-            return self._authenticate_with_refresh_token(refresh_token)
+            try:
+                return self._authenticate_with_refresh_token(refresh_token)
+            except InvalidTokenError as exc:
+                token_expired = exc
 
-        return None
+        # 3. Authenticate with OIDC if no credentials or refresh token
+        # This should fail gracefully if OIDC is not enabled on the server
+        access_token = self._authenticate_with_oidc()
+        if not access_token and token_expired:
+            # If neither authentication method worked and refresh token invalid
+            # raise the InvalidTokenError to force reauthentication
+            raise token_expired
+        return access_token
 
     def _authenticate_with_credentials(self) -> str | None:
         """Authenticate using client_id and secret_key."""
@@ -113,7 +120,9 @@ class TestflingerCliAuth:
         except client.HTTPError as exc:
             if exc.status == HTTPStatus.BAD_REQUEST:
                 self.clear_refresh_token()
-                raise InvalidTokenError(exc.msg) from exc
+                raise InvalidTokenError(
+                    "Invalid or expired refresh token."
+                ) from exc
             self._handle_auth_error(exc)
             return None
 
@@ -227,7 +236,13 @@ class TestflingerCliAuth:
         try:
             init_data = self.client.oidc_auth_initiate()
         except client.HTTPError as exc:
-            sys.exit(f"Failed to initiate OIDC login: {exc.msg}")
+            if exc.status == HTTPStatus.NOT_FOUND:
+                # Server does not have OIDC enabled
+                # Need to return None to fail gracefully
+                return None
+            raise OidcError(
+                f"Failed to initiate OIDC login: {exc.msg}"
+            ) from exc
 
         verification_uri = init_data["verification_uri"]
         user_code = init_data["user_code"]
@@ -264,16 +279,20 @@ class TestflingerCliAuth:
                     case "slow_down":
                         interval += int(exc.headers.get("Retry-After", 5))
                     case "access_denied":
-                        sys.exit("Authentication failed: access denied")
+                        raise OidcError(
+                            "Authentication failed: access denied"
+                        ) from exc
                     case "expired_token":
-                        sys.exit("Authentication timed out. Please try again.")
+                        raise OidcError(
+                            "Authentication timed out. Please try again."
+                        ) from exc
                     case _:
-                        sys.exit(
+                        raise OidcError(
                             "Unexpected error during OIDC "
                             f"authentication: {exc.msg}"
-                        )
+                        ) from exc
 
-        sys.exit("Authentication timed out. Please try again.")
+        raise OidcError("Authentication timed out. Please try again.")
 
 
 def require_role(*roles):
--- a/cli/testflinger_cli/auth.py
+++ b/cli/testflinger_cli/auth.py
@@ -18,6 +18,9 @@
 
 import configparser
 import contextlib
+import sys
+import time
+import webbrowser
 from functools import cached_property, wraps
 from http import HTTPStatus
 from typing import Optional
@@ -37,7 +40,14 @@ from testflinger_cli.errors import (
 class TestflingerCliAuth:
     """Class to handle authentication and authorisation for Testflinger CLI."""
 
-    def __init__(self, client_id: str | None, secret_key: str, tf_client):
+    def __init__(
+        self,
+        method: str = "credentials",
+        client_id: str | None = None,
+        secret_key: str | None = None,
+        tf_client: client.Client = None,
+    ):
+        self.method = method
         self.client_id: Optional[str] = client_id
         self.secret_key = secret_key
         self.client = tf_client
@@ -65,6 +75,10 @@ class TestflingerCliAuth:
         :raises InvalidTokenError: Refresh token already expired
         :return: string with access token (jwt token)
         """
+        # If OIDC method is selected, initialize OIDC device auth flow
+        if self.method == "oidc":
+            return self._authenticate_with_oidc()
+
         # Authenticate with credentials if provided in args or env file
         if self.client_id and self.secret_key:
             return self._authenticate_with_credentials()
@@ -168,22 +182,23 @@ class TestflingerCliAuth:
             else None
         )
 
-    def decode_jwt_token(self) -> Optional[dict]:
+    def decode_jwt_token(self, token: str | None = None) -> Optional[dict]:
         """Decode JWT token without verifying signature.
 
         This is not for security but for quick screening,
         server will still enforce the JWT token validation.
 
+        :param token: optional token to decode, defaults to self.jwt_token
         :return: Dict with the decoded JWT
         """
-        if not self.is_authenticated() or not self.jwt_token:
+        token_to_decode = token or self.jwt_token
+        if not token_to_decode:
             return None
 
         try:
-            decoded_jwt = jwt.decode(
-                self.jwt_token, options={"verify_signature": False}
+            return jwt.decode(
+                token_to_decode, options={"verify_signature": False}
             )
-            return decoded_jwt
         except jwt.exceptions.DecodeError:
             return None
 
@@ -206,6 +221,59 @@ class TestflingerCliAuth:
         # If there is a decoded token, the user was authenticated
         # Default role for legacy client_ids is contributor
         return permissions.get("role", ServerRoles.CONTRIBUTOR)
+
+    def _authenticate_with_oidc(self) -> str | None:
+        """Authenticate using OIDC device flow."""
+        try:
+            init_data = self.client.oidc_auth_initiate()
+        except client.HTTPError as exc:
+            sys.exit(f"Failed to initiate OIDC login: {exc.msg}")
+
+        verification_uri = init_data["verification_uri"]
+        user_code = init_data["user_code"]
+        interval = init_data.get("interval", 5)
+        expires_in = init_data.get("expires_in", 300)
+        request_id = init_data["request_id"]
+
+        print(
+            f"Please visit {verification_uri} and enter code "
+            f"{user_code} to login."
+        )
+
+        # Attempt to launch the browser for the user to login
+        webbrowser.open(verification_uri)
+
+        code_expiration = time.monotonic() + expires_in
+        while time.monotonic() < code_expiration:
+            time.sleep(interval)
+            try:
+                auth_data = self.client.oidc_auth_poll(request_id)
+                decoded = self.decode_jwt_token(auth_data["access_token"])
+                self.client_id = (
+                    decoded.get("permissions", {}).get("client_id")
+                    if decoded
+                    else None
+                )
+                if self.client_id:
+                    self.store_refresh_token(auth_data["refresh_token"])
+                return auth_data["access_token"]
+            except client.HTTPError as exc:
+                match exc.error_code:
+                    case "authorization_pending":
+                        pass
+                    case "slow_down":
+                        interval += int(exc.headers.get("Retry-After", 5))
+                    case "access_denied":
+                        sys.exit("Authentication failed: access denied")
+                    case "expired_token":
+                        sys.exit("Authentication timed out. Please try again.")
+                    case _:
+                        sys.exit(
+                            "Unexpected error during OIDC "
+                            f"authentication: {exc.msg}"
+                        )
+
+        sys.exit("Authentication timed out. Please try again.")
 
 
 def require_role(*roles):
--- a/cli/testflinger_cli/auth.py
+++ b/cli/testflinger_cli/auth.py
@@ -42,10 +42,10 @@ class TestflingerCliAuth:
 
     def __init__(
         self,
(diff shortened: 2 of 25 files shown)


cli/testflinger_cli/auth.py
```diff
@@ -37,7 +40,14 @@
...
     """Class to handle authentication and authorisation for Testflinger CLI."""
 
-    def __init__(self, client_id: str | None, secret_key: str, tf_client):
+    def __init__(
+        self,
+        method: str = "credentials",
+        client_id: str | None = None,
+        secret_key: str | None = None,
+        tf_client: client.Client = None,
+    ):
```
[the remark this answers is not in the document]
rene-oromtz: This actually needs to be a required param as it can't be None

server/src/testflinger/api/auth.py
```diff
@@ -31,6 +31,19 @@
 from testflinger import database
 from testflinger.owasp import OWASPLogger
 
+DEFAULT_REFRESH_TOKEN_EXPIRATION = 6 * 24 * 60 * 60  # 6 days in seconds
+
```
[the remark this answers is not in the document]
rene-oromtz: This is intended, security is harden so the refresh token all now expiring after 6 days

server/devel/dex-config.yaml
```diff
@@ -9,11 +9,18 @@ staticClients:
       - 'http://testflinger-server:5000/auth/callback'
       - 'http://localhost:5000/auth/callback'
       - 'http://127.0.0.1:5000/auth/callback'
+      - '/device/callback'
     name: 'Testflinger Dev'
-    secret: my-oidc-secret
+    public: true
+    grantTypes:
 enablePasswordDB: true
```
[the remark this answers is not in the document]
rene-oromtz: This is intended, device flow in  dex expects `/device/callback` as redirect_uri

cli/testflinger_cli/__init__.py
```diff
@@ -149,6 +149,7 @@ def __init__(self):
             or os.environ.get("TESTFLINGER_ERROR_THRESHOLD")
             or consts.TESTFLINGER_ERROR_THRESHOLD
         )
+        auth_method = getattr(self.args, "method", "credentials")
```
ajzobro: Do we need this? Could the client have a service endpoint to query to know if authentication is required?

cli/testflinger_cli/auth.py
```diff
@@ -65,6 +75,10 @@ def authenticate(self) -> str | None:
         :raises InvalidTokenError: Refresh token already expired
         :return: string with access token (jwt token)
         """
+        # If OIDC method is selected, initialize OIDC device auth flow
+        if self.method == "oidc":
```
ajzobro: I believe that it would be safe to always try internal authentication (quick to check, first) and fall back to OIDC if that fails.

cli/testflinger_cli/auth.py
```diff
@@ -207,6 +222,59 @@ def get_user_role(self) -> str | None:
         # Default role for legacy client_ids is contributor
         return permissions.get("role", ServerRoles.CONTRIBUTOR)
 
+    def _authenticate_with_oidc(self) -> str | None:
+        """Authenticate using OIDC device flow."""
+        try:
+            init_data = self.client.oidc_auth_initiate()
+        except client.HTTPError as exc:
+            sys.exit(f"Failed to initiate OIDC login: {exc.msg}")
```
ajzobro: we should not call sys.exit here -- if we are in a loop and lose network, we should be able to continue retrying.

An HTTP code that says we are no longer authenticated should kick us out to a re-auth and we should prompt the user again.

cli/testflinger_cli/auth.py
```diff
@@ -207,6 +222,59 @@ def get_user_role(self) -> str | None:
...
+                        sys.exit("Authentication failed: access denied")
+                    case "expired_token":
+                        sys.exit("Authentication timed out. Please try again.")
+                    case _:
+                        sys.exit(
+                            "Unexpected error during OIDC "
+                            f"authentication: {exc.msg}"
+                        )
+
+        sys.exit("Authentication timed out. Please try again.")
```
ajzobro: Again, sys.exit should be reserved for the main application, should raise here.

cli/testflinger_cli/client.py
```diff
@@ -68,10 +70,14 @@ def _handle_response_error(self, req: requests.Request):
                 raise VPNError from exc
```
ajzobro, line 70: not for now, but this should gain an enhancement of only declaring it to be a 403 due to "VPN" if trying to access a known-to-not-exist (ought to be 404) url and still getting a 403.

cli/testflinger_cli/client.py
```diff
@@ -68,10 +70,14 @@ def _handle_response_error(self, req: requests.Request):
...
             # Raise HTTPError with clear text message if any other ValueError
-            raise HTTPError(status=req.status_code, msg=req.text) from exc
+            raise HTTPError(
+                status=req.status_code, msg=req.text, headers=req.headers
+            ) from exc
 
         # flask `abort` returns a JSON object with error message
         error_message = error_json.get("message", req.text)
+        # Oauth2 error responses may include an "error" field
+        error_code = error_json.get("error")
```
ajzobro: move down to where it's used?

cli/testflinger_cli/client.py
```diff
@@ -495,3 +506,23 @@ def delete_client_permissions(
...
+
+        :return: dict containing required OIDC parameters for client to
+            authenticate with provider
+        """
+        endpoint = "/oidc/auth-init"
+        response = self.post(endpoint, data={})
+        return json.loads(response)
+
+    def oidc_auth_poll(self, request_id: str) -> dict:
+        """Poll for OIDC authentication result based on request ID.
```
ajzobro, line 521: Is this typically done through polling and not a server event?

ajzobro: actually makes sense in case the client disappears

cli/tests/test_cli_auth.py
```diff
@@ -17,9 +17,11 @@
 
 import configparser
 import json
+import os
```
ajzobro: rather than use our real OS should we instead mock?

cli/tests/test_cli_auth.py
```diff
@@ -309,3 +296,147 @@ def test_refresh_token_expired(tmp_path, requests_mock, monkeypatch):
     history = requests_mock.request_history
     assert len(history) == 2
     assert history[1].path == "/v1/oauth2/refresh"
+
+
+def test_cli_login_defaults_credentials(auth_fixture, capsys):
+    """Test login command with default method authentication."""
+    # This loads credentials as environment variables for the test
+    auth_fixture(ServerRoles.CONTRIBUTOR)
+    client_id = os.environ.get("TESTFLINGER_CLIENT_ID")
```
ajzobro: We should be mocking the environment not using it for testing.

server/devel/openapi_app.py
```diff
@@ -39,3 +40,7 @@ class OpenAPIConfig:
 
 # Create and expose the app in TESTING mode for Flask CLI to use
 app = create_flask_app(OpenAPIConfig)
+
+# Register OIDC blueprint unconditionally so /oidc endpoints appear in the
```
ajzobro, line 44: But why?

server/src/testflinger/api/auth.py
```diff
@@ -343,6 +356,59 @@ def generate_refresh_token(client_id: str, expires_in: int | None) -> str:
...
+    Handles token generation, refresh token expiration, and OWASP logging.
+    Used by both credential-based and OIDC-based authentication flows.
+
+    :param client_id: Testflinger client ID for which to issue tokens
+    :param allowed_resources: Permissions dict to encode into the JWT
+    :param refresh_token_ttl: Optional expiration time for refresh token
+    :return: Dict with access_token, token_type, expires_in, refresh_token
+    """
+    secret_key = os.environ.get("JWT_SIGNING_KEY")
```
ajzobro: Can we name the variable as "signing" instead of "secret"?

server/src/testflinger/api/auth.py
```diff
@@ -343,6 +356,59 @@ def generate_refresh_token(client_id: str, expires_in: int | None) -> str:
...
+    access_token = generate_access_token(allowed_resources, secret_key)
+
+    role = ServerRoles(allowed_resources.get("role", ServerRoles.CONTRIBUTOR))
+    refresh_expires_in = (
+        None
+        if role in (ServerRoles.ADMIN, ServerRoles.MANAGER)
```
ajzobro: I don't think we should have this exception.

server/src/testflinger/api/auth.py
```diff
@@ -343,6 +356,59 @@ def generate_refresh_token(client_id: str, expires_in: int | None) -> str:
...
+        else refresh_token_ttl
+    )
+    refresh_token = generate_refresh_token(
+        client_id, expires_in=refresh_expires_in
+    )
+
+    current_app.owasp_logger.authn_login_success(
+        userid=client_id,
+        description=(
+            f"Client {client_id} successfully authenticated with role {role}"
```
ajzobro: we don't appear to be actually "authenticating" in this function -- I think that this belongs where tyhe client-id is verified to exist and have the right password.

ajzobro: This code used to be in /oauth2/token which was indeed after a successful login, I think we should move this appropriately.

server/src/testflinger/oidc/__init__.py
```diff
@@ -48,7 +48,8 @@ def app_register_oidc(tf_app: APIFlask) -> OAuth | None:
     secret_key = tf_app.config["SECRET_KEY"]
 
     # If provider is not set, app will run without web authentication
-    if not all([client_id, client_secret, issuer, secret_key]):
+    # client_secret is optional: public clients may have none configured
+    if not all([client_id, issuer, secret_key]):
         logger.info("OIDC disabled due to missing configuration parameters")
```
ajzobro, line 53: elsewhere we crash out for this same reason -- which will happen first?  Will we crash out due to not having all parameters or will we simply log some info?

elsewhere is: server/charm/src/config.py‎:80

server/src/testflinger/oidc/api.py
```diff
@@ -0,0 +1,226 @@
+# Copyright (C) 2022-2025 Canonical
```
ajzobro:
```suggestion
# Copyright (C) 2026 Canonical
```

server/src/testflinger/oidc/api.py
```diff
@@ -0,0 +1,226 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""Testflinger OIDC API."""
+
+import logging
+import secrets
+from http import HTTPStatus
+
+import requests
+from apiflask import APIBlueprint, abort
```
ajzobro, line 23: flask and apiflask now both offer `abort` and we seem to be using different things in different places.

Should we be switching more or less over to `apiflask` and/or could we make `flask` work for our purposes and avoid clouding up the water?

server/tests/conftest.py
```diff
@@ -116,6 +116,7 @@ def oidc_app(oidc_client, mongo_app, iam_server, monkeypatch):
     monkeypatch.setenv("OIDC_CLIENT_SECRET", oidc_client.client_secret)
     monkeypatch.setenv("OIDC_PROVIDER_ISSUER", iam_server.url)
     monkeypatch.setenv("WEB_SECRET_KEY", "my_web_secret_key")
```
ajzobro, line 118: I wonder what this is for...

server/tests/test_oidc.py
```diff
@@ -103,8 +103,8 @@ def test_client_refresh_login_if_exists(mock_current_app, oidc_app, user):
     )
 
     client_entry = {
-        "openid_sub": "1234",
-        "email": user.emails[0],
+        "client_id": user.emails[0],
```
ajzobro, line 106: We should talk about the concept that the IDP might give back more than one email address and whether we can know that order and that that order will NEVER change.

What would the implications be if that list were to change?

server/tests/test_oidc_api.py
```diff
@@ -0,0 +1,411 @@
...
+    """Test that oidc_metadata returns the metadata from the OIDC provider."""
+    app, _ = oidc_app
+    with app.app_context():
+        metadata = helpers.oidc_metadata()
+        assert isinstance(metadata, dict)
+        assert "issuer" in metadata
+        # iam_server fixture configuration may have a trailing slash
+        assert metadata["issuer"] == app.config["OIDC_PROVIDER_ISSUER"].rstrip(
+            "/"
+        )  # noqa: E501
```
ajzobro: since this got reformatted i don't think the noqa comment applies anymore
```suggestion
        )
```

server/docker-compose.yml
```diff
@@ -26,8 +26,8 @@ services:
       # TESTFLINGER_VAULT_URL: http://vault:8200
       WEB_SECRET_KEY: my_web_secret_key
       OIDC_CLIENT_ID: testflinger-client
-      OIDC_CLIENT_SECRET: my-oidc-secret
-      OIDC_PROVIDER_ISSUER: http://dex:5556/dex
+      OIDC_CLIENT_SECRET: ""
```
ajzobro, line 29: I would like to understand this change.

ajzobro (changes requested):
OK, quick overview of the application code, did not try to focus on the tests. Let's discuss the overall design first.

cli/testflinger_cli/__init__.py
ajzobro:
```mermaid
sequenceDiagram
    actor User
    participant CLI as testflinger-cli
    participant Server as testflinger-server
    participant OIDC@{ "type": "control" } as OIDC

    rect rgb(235, 245, 255)
    Note over User,Server: Have client-id + secret-key: basic auth
    User->>CLI: Run command with client-id + secret-key
    CLI->>CLI: Validate both parameters are present
    CLI->>Server: Request with Basic Auth<br/>base64(client-id:secret-key)
    Server->>Server: Validate client credentials
    Server-->>CLI: 200 OK / authorized response
    CLI-->>User: Show result
    end
```

ajzobro:
```mermaid
sequenceDiagram
    actor User
    participant CLI as testflinger-cli
    participant Server as testflinger-server
    participant OIDC@{ "type": "control" } as OIDC

    rect rgb(235, 255, 235)
    Note over User,Server: Neither client-id nor secret-key refresh token accepted auth as token user
    User->>CLI: Run command without client-id + secret-key
    CLI->>CLI: Load stored refresh/bearer token
    CLI->>Server: Request with Bearer refresh token
    Server->>Server: Validate token
    Server-->>CLI: 200 OK / authenticated as token-associated user
    CLI-->>User: Show result
    end
```

ajzobro:
```mermaid
sequenceDiagram
    actor User
    participant CLI as testflinger-cli
    participant Server as testflinger-server
    participant OIDC@{ "type": "control" } as OIDC

    rect rgb(255, 245, 235)
    Note over User,OIDC: Neither client-id nor secret-key, token rejected, server uses OIDC
    User->>CLI: Run command without client-id + secret-key
    CLI->>CLI: Load stored refresh/bearer token
    CLI->>Server: Request with Bearer refresh token
    Server->>Server: Token expired or rejected
    Server-->>OIDC: Initiate OIDC auth flow
    OIDC-->>Server: Auth code
    Server->>CLI: 401 Unauthorized, try this instead: Auth code
    CLI->>CLI: Delete stored token
    CLI->>User: Display url and code
    User->>OIDC: Complete the handshake
    CLI->>Server: Polling for auth completion
    Server-->>OIDC: Is this user cool?
    OIDC-->>Server: Yes + User identity (email)
    Server->>CLI: Issue refresh token
    User-->CLI: Potential user re-issue of command?
    CLI->>Server: Retry request with new Bearer token
    Server-->>CLI: 200 OK / authenticated response
    CLI-->>User: Show result
    end
```

ajzobro:
```mermaid
sequenceDiagram
    actor User
    participant CLI as testflinger-cli
    participant Server as testflinger-server
    participant OIDC@{ "type": "control" } as OIDC

    rect rgb(255, 235, 235)
    Note over User,Server: Neither client-id nor secret-key, token rejected, server does not use OIDC
    User->>CLI: Run command without client-id + secret-key
    CLI->>CLI: Load stored refresh/bearer token
    CLI->>Server: Request with Bearer refresh token
    Server->>Server: Token expired or rejected
    Server-->>CLI: 401 Unauthorized, delete stored token
    CLI->>CLI: Delete stored token
    CLI-->>User: Authentication failed
    end
```

ajzobro:
```mermaid
sequenceDiagram
    actor User
    participant CLI as testflinger-cli
    participant Server as testflinger-server
    participant OIDC@{ "type": "control" } as OIDC

    rect rgb(245, 245, 245)
    Note over User,Server: Neither client-id nor secret-key, no token, server does not require auth
    User->>CLI: Run command without credentials or token
    CLI->>CLI: No stored bearer/refresh token found
    CLI->>Server: Request without Authorization header
    Server->>Server: Auth not required
    Server-->>CLI: 200 OK / anonymous user response
    CLI-->>User: Show result
    end
```

cli/testflinger_cli/__init__.py
```diff
@@ -149,6 +149,7 @@ def __init__(self):
...
```
ajzobro, line 152:
```mermaid
sequenceDiagram
    actor User
    participant CLI as testflinger-cli
    participant Server as testflinger-server
    participant OIDC@{ "type": "control" } as OIDC

    rect rgb(245, 235, 255)
    Note over User,OIDC: Neither client-id nor secret-key, no token, server uses OIDC
    User->>CLI: Run command without credentials or token
    CLI->>CLI: No stored bearer/refresh token found
    CLI->>Server: Request without Authorization header
    Server-->>OIDC: Initiate OIDC auth flow
    OIDC-->>Server: Auth code
    Server->>CLI: 401 Unauthorized, try this instead: Auth code
    CLI->>User: Display URL and code
    User->>OIDC: Complete the handshake
    CLI->>Server: Poll for auth completion
    Server-->>OIDC: Is this user cool?
    OIDC-->>Server: Yes + user identity (email)
    Server->>CLI: Issue refresh token
    User-->>CLI: Potential user re-issue of command?
    CLI->>Server: Retry request with new Bearer token
    Server-->>CLI: 200 OK / authenticated response
    CLI-->>User: Show result
    end
```

server/docker-compose.yml
rene-oromtz: For some reason, Dex does not let me use client secret and client id for using the device flow. Not sure if a bug upstream of a misconfiguration from my side but I've been reviewing their examples and it should work with providing credentials...

For testing locally, only alternative I could found was to make Dex "public" (this does not require client_secret).

server/tests/test_oidc.py
rene-oromtz: Okay, I was researching this, seems that the iam_server fixture is using the [SCIM](https://www.rfc-editor.org/rfc/rfc7643) (System for Cross-domain Identity Management) schema that allows the multi-value `emails`. 

The IDP provider on the other hand, needs to conform to the [OIDC Spec](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) and that one specifies the email to be a string which should be the user preferred email:
> email 	string 	End-User's preferred e-mail address.

server/tests/conftest.py
rene-oromtz: The WEB_SECRET_KEY? That is required so Flask app can sign the web session cookies

server/src/testflinger/oidc/api.py
rene-oromtz: If we want a JSON response, we should use apiflask. I think we mostly are using apiflask from the API as this is easier to parse

server/src/testflinger/oidc/__init__.py
rene-oromtz: Charm will block first, so this code should not be reacheable on charm deployments. only local deployment should get here

server/devel/openapi_app.py
rene-oromtz: This is so they are available in the schema. The schema generation script runs a "simple" flask application that does not define OIDC variables. Given OIDC routes are still meant to be consumed by the CLI, we need to include them so they are documented

server/src/testflinger/api/auth.py
rene-oromtz: Moved to the endpoint definition in [e6d40e1](https://github.com/canonical/testflinger/pull/1152/commits/e6d40e1b34ced0f9bfafe7963435aa65610bda6b)

server/src/testflinger/api/auth.py
rene-oromtz: I agree with you given we need a more strict security, I ended up removing it in [af49a2e](https://github.com/canonical/testflinger/pull/1152/commits/af49a2e22e47d942e7e046b86f9d8a534153fd74)
But @pedro-avalos and @diohe0311 needs to be aware of this change given this impacts HEXR (and lab manager?)

server/src/testflinger/api/auth.py
rene-oromtz: Sure thing! Done in [e6d40e1](https://github.com/canonical/testflinger/pull/1152/commits/e6d40e1b34ced0f9bfafe7963435aa65610bda6b)

cli/tests/test_cli_auth.py
rene-oromtz: the auth_fixture is mocking the envvar so this is already taken care of by monkeypatch. I ended up replacing this anyway to validate the TEST_CLIENT_ID given this is more explicit anyway
Ref [e6d40e1](https://github.com/canonical/testflinger/pull/1152/commits/e6d40e1b34ced0f9bfafe7963435aa65610bda6b)

cli/tests/test_cli_auth.py
rene-oromtz: Done in [e6d40e1](https://github.com/canonical/testflinger/pull/1152/commits/e6d40e1b34ced0f9bfafe7963435aa65610bda6b)

cli/testflinger_cli/client.py
rene-oromtz: Moved it down in [e6d40e1](https://github.com/canonical/testflinger/pull/1152/commits/e6d40e1b34ced0f9bfafe7963435aa65610bda6b)

cli/testflinger_cli/auth.py
rene-oromtz: Raising with a new OIDC Error in [280f3ef](https://github.com/canonical/testflinger/pull/1152/commits/280f3efccfd506df7eb102255ca94925900556c9)

server/src/testflinger/oidc/api.py
rene-oromtz: Done in [e6d40e1](https://github.com/canonical/testflinger/pull/1152/commits/e6d40e1b34ced0f9bfafe7963435aa65610bda6b)

cli/testflinger_cli/auth.py
rene-oromtz: Agree, this was done in [280f3ef](https://github.com/canonical/testflinger/pull/1152/commits/280f3efccfd506df7eb102255ca94925900556c9)

cli/testflinger_cli/auth.py
rene-oromtz: This is just for the initial OIDC dance, I change the sys exit for a raise in [280f3ef](https://github.com/canonical/testflinger/pull/1152/commits/280f3efccfd506df7eb102255ca94925900556c9)

server/tests/test_oidc_api.py
rene-oromtz: Fixed in [17c1249](https://github.com/canonical/testflinger/pull/1152/commits/17c12497a692fb5027e74b765e66f030d7df97e5)

rene-oromtz:
@ajzobro thanks for your review! I modified the PR with the design decisions we agree on, please let me know if I capture those effectively!

server/docker-compose.yml
ajzobro: ty

server/tests/test_oidc.py
ajzobro: My only concern then, would be if a user changes their preferred email address.

barbara@example.com -> barb@canonical.com

Doesn't matter because we are using the sub in the client permissions Here we are just setting the email (preferred email).

server/src/testflinger/api/auth.py
diohe0311: Thanks for flagging this.

This will affect C3. We currently use a Testflinger admin token for restricted-queue automation (used by QA, HWE, and PE), and we've been relying on the refresh token not expiring so we don't need to handle token renew.

If admin refresh tokens are going to expire as well, we'll need to figure out what changes are needed on the C3 side.

ajzobro:
Service accounts ... need to switch to using service account credentials.
We can talk more about this in the C3 context offline.

server/src/testflinger/oidc/api.py
```diff
@@ -101,10 +103,20 @@ def _register_and_issue_tokens(userinfo: dict, request_id: str) -> dict:
 
     # Delete device code after all operations have succeeded
     database.delete_oidc_device_code(request_id)
```
ajzobro, line 105: Do unused device codes get cleaned up after a bit? What is the lifetime of a device code? We are cleaning up, here, after success -- where do we deal with unused/failed device codes?

rene-oromtz: Yes, devices codes does gets cleaned up by the lifetime defined by the IDP provider (usually 300 seconds). 

IDP:
- Set the expiration time to 300 seconds after which, the IDP does no longer recognize this code as valid

Testflinger:
- Given the auth is no longer valid, this gets deleted from our database via [TTL](https://github.com/canonical/testflinger/pull/1152/changes#diff-36d910d54d1305414a4e92e63df96a1c561cf9136e3c1b1aa344934e1cc6d781R122)
- If authentication is successful, it will manually remove it bypassing the TTL
- If authentication is unsuccessful, it will also manually remove the device code from out database

ajzobro: Roger that: it's the TTL that I wanted to hear about. Thank you.

ajzobro (commented):
LGTM

ajzobro (approved):
Approved

PR #1154, opened 2026-06-17
fix(server-charm): add pydantic validation to CSFLE master key

rene-oromtz:
## Description
This add a pydantic validation at charm level so MongoDB CSFLE encryption key is valid base64 with 96-byte length. 
When unset, this will also clear the stored CSFLE key in charm. 

## Resolved issues
NA

## Documentation
NA

## Tests
Added unit tests for this behavior

### DIFF
--- a/server/charm/src/charm.py
+++ b/server/charm/src/charm.py
@@ -294,6 +294,14 @@ class TestflingerCharm(ops.CharmBase):
         new_key = self.typed_config.testflinger_secrets_master_key
         stored_key = self._stored.previous_master_key
 
+        # If the key was unset, clear the stored key so the next
+        # config-changed with a valid key is treated as a fresh deployment.
+        if not new_key:
+            logger.info("Master key unset, clearing stored key")
+            self._stored.previous_master_key = ""
+            self._update_layer_and_restart()
+            return
+
         # Rotate only if master key was previously defined, has changed,
         # and this is the leader unit.
         if stored_key and new_key != stored_key and self.unit.is_leader():
--- a/server/charm/tests/unit/test_charm.py
+++ b/server/charm/tests/unit/test_charm.py
@@ -533,3 +533,24 @@ def test_conflict_cleared_on_route_changed(ctx):
         state_in,
     )
     assert state_out.unit_status == testing.ActiveStatus()
+
+
+def test_config_changed_clears_stored_key_on_unset(ctx, make_state, caplog):
+    """Test that stored key is cleared when config key is unset."""
+    old_key = helpers.generate_b64_key()
+    state_in = make_state(
+        config={"testflinger_secrets_master_key": ""},
+        stored_key=old_key,
+    )
+    state_out = ctx.run(ctx.on.config_changed(), state_in)
+
+    # Verify log message
+    assert "Master key unset, clearing stored key" in caplog.text
+
+    # Verify stored key is cleared
+    stored = next(
+        state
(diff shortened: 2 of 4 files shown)


server/charm/src/charm.py
```diff
@@ -294,6 +294,14 @@ def _on_config_changed(self, _: ops.framework.EventBase) -> None:
         new_key = self.typed_config.testflinger_secrets_master_key
         stored_key = self._stored.previous_master_key
 
+        # If the key was unset, clear the stored key so the next
+        # config-changed with a valid key is treated as a fresh deployment.
+        if not new_key:
+            logger.info("Master key unset, clearing stored key")
```
ajzobro, line 300: Do we really want to support on-the-fly removal of the secret master key?

Is this also going to disable secrets when/if someone clears out the master key?

server/charm/src/charm.py
```diff
@@ -294,6 +294,14 @@ def _on_config_changed(self, _: ops.framework.EventBase) -> None:
...
+            self._stored.previous_master_key = ""
+            self._update_layer_and_restart()
+            return
+
         # Rotate only if master key was previously defined, has changed,
         # and this is the leader unit.
```
ajzobro, line 306: Here we only change the key if we are the leader unit...  should we consider that when we are clearing the key as well?

server/charm/tests/unit/test_charm.py
```diff
@@ -533,3 +533,24 @@ def test_conflict_cleared_on_route_changed(ctx):
...
+    # Verify log message
+    assert "Master key unset, clearing stored key" in caplog.text
+
+    # Verify stored key is cleared
+    stored = next(
+        state
+        for state in state_out.stored_states
+        if state.owner_path == "TestflingerCharm" and state.name == "_stored"
+    )
+    assert stored.content["previous_master_key"] == ""
```
ajzobro, line 556: This is verifying the "next" state has a "previous" master key that is blank; shouldn't we be verifying the current key is empty?

ajzobro (commented):
Looks good, minor questions.

server/charm/tests/unit/test_charm.py
rene-oromtz: The `previous_master_key` is the value of the key internally stored by Juju which needs to be cleared if it was unset

server/charm/src/charm.py
rene-oromtz: Removing it will disable the secret store but also additional cleanup is needed (manually removing the relation), this is also needed because the stored state can't be modified easily, so you need to unset to not force a rotation

server/charm/src/charm.py
rene-oromtz: Good question, normally this is required when sending changes directly to a DB (such as Mongo). Given this is internal state, perhaps its also a DB :thinking: I can add this, doesn't hurt to have it

rene-oromtz: ah perhaps I was wrong? It seems that the data may be per unit if I get this right:
https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.StoredState

> Data is stored alongside the charm (in the charm container for Kubernetes sidecar charms, and on the machine for machine charms).

If I get this right, each container have its own data store, so we need to clear them from all units

ajzobro (approved):
Thanks for the explanations and further thought with respect to my comments, LGTM.

PR #1161, opened 2026-06-22
Docs/auth flows and detail

ajzobro:
## Description

WIP, only the pictures

### DIFF
--- a/docs/explanation/authentication.rst
+++ b/docs/explanation/authentication.rst
@@ -22,9 +22,14 @@ The following sequence diagrams show the authentication flow for each of the
 authentication methods and the interactions between the user, the CLI and the
 server.
 
+.. note::
+
+    For simplicity, all of the diagrams are targetting the ``/v1/jobs`` endpoint
+    on Testflinger server side but the same flow applies to any other endpoint
+    that requires authentication.
+
 Basic Auth
 ^^^^^^^^^^
-
 
 Regardless of whether OIDC is enabled or not, Basic Auth will be used if both the
 ``client_id`` and ``secret_key`` are provided.
@@ -43,10 +48,12 @@ Regardless of whether OIDC is enabled or not, Basic Auth will be used if both th
         User->>CLI: Run command with client-id + secret-key
         CLI->>CLI: Validate both parameters are present
         CLI->>Server: Request with Basic Auth<br/>base64(client-id:secret-key)
+        Note over CLI,Server: POST /v1/oauth2/token
         Server->>Server: Validate client credentials
         Server-->>CLI: 200 OK / return access and refresh tokens
         CLI-->>Server: use access token to complete the request
-        Server->>Server: Validate client permissions
+        Note over CLI,Server: POST /v1/jobs
+        Server->>Server: Validate access token signature and client permissions
         Server-->>CLI: 200 OK / authenticated response
         CLI-->>User: Show result
 
@@ -93,17 +100,19 @@ OIDC is enabled on the server:
         User->>CLI: Run command without client-id + secret-key
         CLI->>CLI: Load stored refresh token
         CLI->>Server: Request with refresh token
+        Note over CLI,Server: POST /v1/oauth2/refresh
         Server->>Server: Validate token
         Server-->>CLI: 200 OK / return access and refresh tokens
         CLI-->>Server: use access token to complete the request
-        Server->>Server: Validate client permissions
+        Note over CLI,Server: POST /v1/jobs
+        Server->>Server: Validate access token signature and client permissions
         Server-->>CLI: 200 OK / authenticated response
         CLI-->>User: Show result
 
 Token expired or rejected
 .........................
 
-When the refresh token is rejected, the behavior differs depending on whether
+When the refresh token is rejected, the behaviour differs depending on whether
 the server has OIDC enabled.
 
 With OIDC Enabled
@@ -125,21 +134,27 @@ initiates an OIDC authentication flow:
 
         Note over User,OIDC: Neither client-id nor secret-key, token rejected, server uses OIDC
         User->>CLI: Run command without client-id + secret-key
-        CLI->>CLI: Load stored refresh/bearer token
-        CLI->>Server: Request with Bearer refresh token
+        CLI->>CLI: Load stored refresh token
+        CLI->>Server: Request with refresh token
+        Note over CLI,Server: POST /v1/oauth2/refresh
         Server->>Server: Token expired or rejected
-        Server-->>OIDC: Initiate OIDC auth flow
-        OIDC-->>Server: Auth code
-        Server->>CLI: 401 Unauthorized, try this instead: user code + URL for OIDC handshake
-        CLI->>CLI: Delete stored token
+        Server-->>CLI: 400 Bad Request
+        CLI->>CLI: Delete stored refresh token
+        CLI->>Server: Initiate OIDC auth flow
+        Note over CLI,Server: POST /oidc/auth-init
+        Server-->>OIDC: Proxy request to OIDC provider
+        OIDC-->>Server: device_code + user_code + URL for OIDC handshake
+        Server->>CLI: request_id + user code + URL for OIDC handshake
         CLI->>User: Display URL and code
         User->>OIDC: Complete the handshake
         CLI->>Server: Poll for auth completion
+        Note over CLI,Server: POST /oidc/auth-poll/<request_id>
         Server-->>OIDC: Is this user cool?
         OIDC-->>Server: Yes + user identity (email)
         Server->>CLI: Issue refresh and access token
         CLI->>Server: Send request with access token
-        Server->>Server: Validate client permissions
+        Note over CLI,Server: POST /v1/jobs
+        Server->>Server: Validate access token signature and client permissions
         Server-->>CLI: 200 OK / authenticated response
         CLI-->>User: Show result
 
@@ -161,17 +176,18 @@ the server returns a 401 and the CLI removes the stored token:
 
         Note over User,Server: Neither client-id nor secret-key, token rejected, server does not use OIDC
         User->>CLI: Run command without client-id + secret-key
-        CLI->>CLI: Load stored refresh/bearer token
-        CLI->>Server: Request with Bearer refresh token
+        CLI->>CLI: Load stored refresh token
+        CLI->>Server: Request with refresh token
+        Note over CLI,Server: POST /v1/oauth2/refresh
         Server->>Server: Token expired or rejected
-        Server-->>CLI: 401 Unauthorized, delete stored token
-        CLI->>CLI: Delete stored token
+        Server-->>CLI: 400 Bad Request
+        CLI->>CLI: Delete stored refresh token
         CLI-->>User: Authentication failed
 
 No credentials or stored token
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-When neither credentials nor a stored token are available, the behavior
+When neither credentials nor a stored token are available, the behaviour
 depends on whether the server has OIDC enabled.
 
 If the server has OIDC enabled, it initiates an OIDC authentication flow:
@@ -190,18 +206,21 @@ If the server has OIDC enabled, it initiates an OIDC authentication flow:
         Note over User,OIDC: Neither client-id nor secret-key, no token, server uses OIDC
         User->>CLI: Run command without credentials or token
         CLI->>CLI: No stored bearer/refresh token found
-        CLI->>Server: Request without Authorization header
-        Server-->>OIDC: Initiate OIDC auth flow
-        OIDC-->>Server: Auth code
-        Server->>CLI: Return user code and URL for OIDC handshake
+        CLI->>Server: Initiate OIDC auth flow
+        Note over CLI,Server: POST /oidc/auth-init
+        Server-->>OIDC: Proxy request to OIDC provider
+        OIDC-->>Server: device_code + user_code + URL for OIDC handshake
+        Server->>CLI: request_id + user code + URL for OIDC handshake
         CLI->>User: Display URL and code
         User->>OIDC: Complete the handshake
         CLI->>Server: Poll for auth completion
+        Note over CLI,Server: POST /oidc/auth-poll/<request_id>
         Server-->>OIDC: Is this user cool?
         OIDC-->>Server: Yes + user identity (email)
-        Server->>CLI: Issue refresh token
-        CLI->>Server: Retry request with new Bearer token
-        Server->>Server: Validate client permissions
+        Server->>CLI: Issue refresh token and access token
+        CLI->>Server: Send request with access token
+        Note over CLI,Server: POST /v1/jobs
+        Server->>Server: Validate access token signature and client permissions
         Server-->>CLI: 200 OK / authenticated response
         CLI-->>User: Show result
 
@@ -222,6 +241,7 @@ as an anonymous user:
         User->>CLI: Run command without credentials or token
         CLI->>CLI: No stored bearer/refresh token found
         CLI->>Server: Request without Authorization header
+        Note over CLI,Server: POST /v1/jobs
         Server->>Server: Auth not required
         Server-->>CLI: 200 OK / anonymous user response
         CLI-->>User: Show result
--- a/docs/explanation/authentication.rst
+++ b/docs/explanation/authentication.rst
@@ -163,7 +163,7 @@ When the refresh token is not accepted:
         Server->>Server: Token expired or rejected
         Server-->>CLI: 401 Unauthorized, delete stored token
         CLI->>CLI: Delete stored token
-        CLI-->>User: Authentication faile
+        CLI-->>User: Authentication failed
     
 When the refresh token is not available:
 ........................................
--- a/docs/explanation/authentication.rst
+++ b/docs/explanation/authentication.rst
@@ -12,8 +12,7 @@ OIDC is *not* enabled, Testflinger supports anonymous logins. However, when OIDC
 each request must be authenticated. Both operating modes support use of self-signed tokens
 for authentication, and these are what are used in typical CI/CD flows.
 
-Since Basic Auth can be used in both configurations, it is prioritized first, when
-Basic Auth is used it will be handled within the server before other authentication methods.
+Since Basic Auth can be used in both configurations, it is prioritized and will be handled within the server before other authentication methods.
 
 Authentication Flows
 --------------------
@@ -51,8 +50,8 @@ Regardless of whether OIDC is enabled or not, Basic Auth will be used if both th
 ``client_id`` and ``secret_key`` are provided.
 
 .. mermaid::
-    :alt: Sequence diagram showing BASIC Auth is used when client_id and secret_key are provided
-    :caption: Sequence diagram showing BASIC Auth is used when client_id and secret_key are provided
+    :alt: Sequence diagram showing Basic Auth is used when client_id and secret_key are provided
+    :caption: Sequence diagram showing Basic Auth is used when client_id and secret_key are provided
     :align: center
 
     sequenceDiagram
@@ -129,8 +128,7 @@ OIDC is enabled on the server:
 Token expired or rejected
 .........................
 
-If the server does not have OIDC enabled and the refresh token is rejected,
-the server returns a 401 and the CLI removes the stored token. If the CLI does not
+The server returns a 401 and the CLI removes the stored token. If the CLI does not
 automatically retry the operation anonymously it will signal to the user that the
 request was denied so the user may try again.
 
@@ -182,7 +180,6 @@ as an anonymous user:
 OIDC Enabled
 ~~~~~~~~~~~~
 
-When Testflinger is configured to use OpenID Connect (OIDC) for authentication, the order is:
 When Testflinger is configured to use OpenID Connect (OIDC) for authentication, the order is:
 
 1. Basic Auth using ``client_id`` and ``secret_key``
--- a/docs/explanation/authentication.rst
+++ b/docs/explanation/authentication.rst
@@ -1,7 +1,7 @@
 .. _authentication:
 
 Authentication and Authorisation
---------------------------------
+================================
 
 Authentication requires a client_id and a secret_key. These credentials can be
 obtained by contacting the server administrator with the queues you want priority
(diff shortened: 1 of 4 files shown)


ajzobro:
This is just the pictures -- we need words as well, but I wanted to get the mermaid diagrams out there to share for development.

docs/explanation/authentication.rst
```diff
@@ -22,4 +22,182 @@ until the cached credentials expire.
...
+.. mermaid::
+    :alt: Sequence diagram showing a refresh token being rejected and the server initiating OIDC process
+    :caption: Sequence diagram showing a refresh token being rejected and the server initiating OIDC process
+    :align: center
+
+    sequenceDiagram
+        actor User
+        participant CLI as testflinger-cli
+        participant Server as testflinger-server
+
```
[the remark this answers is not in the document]
ajzobro: @copilot; update from OIDC to Server, don't add participants.

docs/explanation/authentication.rst
```diff
@@ -22,4 +22,182 @@ until the cached credentials expire.
...
+        Server-->>CLI: 200 OK / authenticated response
+        CLI-->>User: Show result
+   
+When the refresh token is not available:
+........................................
+
+.. mermaid::
+    :alt: Sequence diagram showing a refresh token being used and accepted
+    :caption: Sequence diagram showing a refresh token being used and accepted
+    :align: center
```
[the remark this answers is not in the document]
ajzobro: @copilot: revert all your changes to all files.

docs/explanation/authentication.rst
```diff
@@ -22,4 +22,182 @@ until the cached credentials expire.
...
+        participant Server as testflinger-server
+
+        Note over User,Server: Neither client-id nor secret-key, token rejected, server does not use OIDC
+        User->>CLI: Run command without client-id + secret-key
+        CLI->>CLI: Load stored refresh/bearer token
+        CLI->>Server: Request with Bearer refresh token
+        Server->>Server: Token expired or rejected
+        Server-->>CLI: 401 Unauthorized, delete stored token
+        CLI->>CLI: Delete stored token
+        CLI-->>User: Authentication faile
```
ajzobro:
```suggestion
        CLI-->>User: Authentication failed
```

docs/explanation/authentication.rst
```diff
@@ -22,4 +22,182 @@ until the cached credentials expire.
 
 .. tip::
     You can also run ``testflinger-cli login`` without command line arguments if your credentials
-    are located in a ``.env`` file as mentioned in :doc:`Authentication using Testflinger CLI <../how-to/authentication>`
\ No newline at end of file
+    are located in a ``.env`` file as mentioned in :doc:`Authentication using Testflinger CLI <../how-to/authentication>`
+
+Authorization Flows Based on Configuration and Provided Auth Parameters
+=======================================================================
```
tang-mm: Is this heading supposed to be an H2 heading beneath the same title? The current marker is at the same level as "Authentication and Authorisation", which makes the document appear as a new topic in the sidebar, and thus the sub-headings are not rendered in the right side panel in the [preview](https://canonical-ubuntu-documentation-library--1161.com.readthedocs.build/testflinger/1161/explanation/authentication/#)

I'd assume this level should use dash `--------`, and H3-heading with `~~~`: 

```
Authorization Flows Based on Configuration and Provided Auth Parameters
------------------------------------------------------------------------

When the client_id and secret_key are provided:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

docs/explanation/authentication.rst
```diff
@@ -22,4 +22,182 @@ until the cached credentials expire.
...
+        CLI->>CLI: Load stored refresh/bearer token
+        CLI->>Server: Request with Bearer refresh token
+        Server->>Server: Validate token
+        Server-->>CLI: 200 OK / authenticated as token-associated user
+        CLI-->>User: Show result
+    
+When OIDC is enabled:
+^^^^^^^^^^^^^^^^^^^^^
+
+When the refresh token is not accepted:
```
tang-mm: reword to something like:
```suggestion
Token refresh fails
```

docs/explanation/authentication.rst
```diff
@@ -22,4 +22,182 @@ until the cached credentials expire.
...
+        CLI->>Server: Poll for auth completion
+        Server-->>OIDC: Is this user cool?
+        OIDC-->>Server: Yes + user identity (email)
+        Server->>CLI: Issue refresh token
+        User-->>CLI: Potential user re-issue of command?
+        CLI->>Server: Retry request with new Bearer token
+        Server-->>CLI: 200 OK / authenticated response
+        CLI-->>User: Show result
+   
+When the refresh token is not available:
```
tang-mm:
```suggestion
First-time login or missing token
```

tang-mm (commented):
Nice diagrams! they really make the flow explanation very clear. 

Some thoughts about the general structure: 

Currently, the heading structure reflects a decision tree, which matches the implementation. However, it may complicate document scanning as it requires people to start triaging from the first section. I'd suggest reorganizing the sections by the auth mechanism, for example:
- Basic auth
- Bearer auth
- OIDC re-authentication  (it seems that OIDC is only activated when the two previous auths fail)
- No authentication (or anonymous access)
- Invalid credential combinations (maybe part of Basic auth?)

In general, if the precedence is defined as "Basic > Bearer > OIDC", it would be helpful to add a short overview paragraph explaining how the backend selected the auth method. This way, the branching logic does not need to be repeated in the headings.

Similarly, the differences between OIDC enabled/disabled feel more like a system configuration layer rather than an auth mechanism, so those could probably be separated out as a distinct subsection. 

(PS: I know these are WIP for development usage, but I'd leave the note for user-facing publication later. Happy to discuss!)

ajzobro:
@tang-mm Thank you for the review.  Yes, I agree that the order could possibly be improved. We actually would probably benefit from a section at the top with a genuine flowchart capturing:
```
if we have (client-id + secret-key)
  a) basic auth client-id:secret-key


if we do NOT have (client-id + secret-key)
  a) refresh token (bearer)
    i) Refresh token Accepted
        A) auth'd as user assoc w/Token
    ii) Token Expired / Rejected
        A) server uses OIDC
          1) 401 cli told to delete token
          2) initiate OIDC auth flow
        B) else
          1) cli told to delete token


  b) no bearer or refresh token
    1) server no required auth
      i) Anonymous user
    2) server uses OIDC
      ii) initiate OIDC auth flow


If the CLI receives only one of client-id or secret-key it rejects it as invalid combination of parameters
```

rene-oromtz:
@ajzobro as discussed, I took the liberty to add some words to the current diagram flows (and also update the how-to for login). 

From the proposed flow, the only thing I changed (if I'm correct) is the case where one of the arguments were provided as I think it also makes sense adding it under the "basic" section, but given this is still a WIP, feel free to let me know what you think or if you have any other modification, please feel free to push any other change you think is needed!

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,280 @@
 .. _authentication:
 
 Authentication and Authorisation
```
ajzobro, line 3:
```suggestion
Authentication and Authorization
```

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,280 @@
...
+
+    sequenceDiagram
+        actor User
+        participant CLI as testflinger-cli
+        participant Server as testflinger-server
+
+        Note over User,Server: Have client-id + secret-key, use basic auth
+        User->>CLI: Run command with client-id + secret-key
+        CLI->>CLI: Validate both parameters are present
+        CLI->>Server: Request with Basic Auth<br/>base64(client-id:secret-key)
```
ajzobro, line 64: I would like to add the endpoint, since it is explicitly the authorization endpoint: /oauth2/token

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,280 @@
...
+        Server->>Server: Validate client credentials
+        Server-->>CLI: 200 OK / return access and refresh tokens
+        CLI-->>Server: use access token to complete the request
```
ajzobro, line 68: In this case, we would send the request to the intended endpoint, e.g. `/jobs/<job_id>`; the "Validate client permissions" could be:
a) Validate token authenticity (token sig)
b) Validate permissions for endpoint

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,280 @@
...
+    :align: center
+
+    sequenceDiagram
+        actor User
+        participant CLI as testflinger-cli
+        participant Server as testflinger-server
+
+        Note over User,Server: Neither client-id nor secret-key, refresh token accepted
+        User->>CLI: Run command without client-id + secret-key
+        CLI->>CLI: Load stored refresh/bearer token
```
ajzobro:
```suggestion
        CLI->>CLI: Load stored refresh token
```

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,280 @@
...
+        CLI->>Server: Request with Bearer refresh token
```
ajzobro:
```suggestion
        CLI->>Server: Request with refresh token
```

ajzobro: add a note about which endpoint: `/oauth2/refresh` (vs `token` in the Basic case)

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,280 @@
...
+        Server->>Server: Validate token
+        Server-->>CLI: 200 OK / return access and refresh tokens
+        CLI-->>Server: use access token to complete the request
```
ajzobro, line 121: add endpoint note; e.g. /jobs/<job_id>

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,280 @@
...
+        Server-->>OIDC: Initiate OIDC auth flow
+        OIDC-->>Server: Auth code
+        Server->>CLI: 401 Unauthorized, try this instead: user code + URL for OIDC handshake
+        CLI->>CLI: Delete stored token
+        CLI->>User: Display URL and code
+        User->>OIDC: Complete the handshake
+        CLI->>Server: Poll for auth completion
+        Server-->>OIDC: Is this user cool?
+        OIDC-->>Server: Yes + user identity (email)
+        Server->>CLI: Issue refresh token
```
ajzobro:
```suggestion
        Server->>CLI: Issue refresh and access token
```

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,280 @@
...
+        CLI->>Server: Retry request with new Bearer token
```
ajzobro:
```suggestion
        CLI->>Server: Send request with access token
```
Add note about uri -- now we finally use the access token to reach the desired endpoint

ajzobro (commented):
Here are some more comments, after tinkering with the diagrams again I'll review the whole thing again.

docs/explanation/authentication.rst
rene-oromtz: Its been a while since I read the recommendations, but as far as I remember, UK spelling was preferred over US spelling. @tang-mm perhaps you may know which is the latest recommendation? I honestly don't remember where the documentation was

diohe0311:
Thanks for putting these diagrams together, Andy. They make the auth flow really easy to follow.

From the C3 side, this looks pretty straightforward. Rather than storing a long-lived refresh token, we'd just keep the client ID and secret and request an access token when needed. So the 6 days refresh token expiry shouldn't be an issue for us.

Other than that, I don't see any impact on C3. I'll make the necessary changes in C3, but nothing here would block us. Thanks again!

docs/explanation/authentication.rst
ajzobro: There was a recent (since I've been here) announcement that US English was to be used for new projects.  There aren't many of these, and we can leave this one until and if we decide to do the whole thing; but I do like to see my `z`'s when I can.

rene-oromtz: Yeah, once we clarify this we can check which other words are in UK English and replace them... I saw the UK recommendation when I was starting at Canonical so I might be the one outdated!

docs/explanation/authentication.rst
rene-oromtz: I modified all the headings so this should be already fixed

docs/explanation/authentication.rst
rene-oromtz: I think this is now: `Token expired or rejected`. I'll mark this as resolved for now, but please let me know if that also needs a clarification

docs/explanation/authentication.rst
rene-oromtz: I think this is now: `No credentials or stored token` I'll mark this as resolved for now, but please let me know if this perhaps is clearer or preferred as suggested!

docs/explanation/authentication.rst
ajzobro: no change for now.

ajzobro:
One more thing that we might want to consider is that while the present set of diagrams is ordered and presented to describe all the potential auth flows with their priorities as well (e.g. basic before refresh before OIDC); which is certainly important to know since all three apply in the case where OIDC auth is enabled.

However; the application itself can be deployed in two distinct configurations: Without OIDC enabled, and With OIDC enabled.

In the case where we have enabled OIDC, the following flows apply, in order:
 - Basic Auth -> return access + refresh token
 - Access Token -> @rene-oromtz  our conversation earlier makes me think that our diagrams are not complete as it sounded to me like if we have an access token we will use that elusively until it is expired and we have no such diagram of direct access token usage.
 - Refresh Token -> return access token
 - No available token -> OIDC Auth device flow -> access + refresh token
 
 In the case where we have NOT enabled OIDC, the following flows apply, in order:
 - Basic Auth -> return access + refresh token
 - Access Token -> @rene-oromtz  our conversation earlier makes me think that our diagrams are not complete as it sounded to me like if we have an access token we will use that elusively until it is expired and we have no such diagram of direct access token usage.
 - Refresh Token -> return access token
 - No available token -> Anonymous CONTRIBUTOR access
 
 I feel that the overall document should have more headings (one per flow, e.g. above; and larger main headings for each of No OIDC and With OIDC).  The purpose of the two main headings is to help focus the material to the situation that applies to any given deployment, since both are not applicable at once and users/administrators and potential users/admins can see what the difference will be based on their choice of configuration.

rene-oromtz:
@ajzobro yes, I think that order makes sense and perhaps more in line to the suggestion by @tang-mm. Given OIDC is a feature, perhaps user wise would be to split them in OIDC vs No OIDC so can we direct the focus to only one section given those are mutually exclusive. 

Perhaps:
Basic Auth - both diagrams (success and failure) given this is applicable for both setups (2)
Refresh Token success - Also works the same for both setups (1)
With OIDC: 
- Token expired or rejected (1)
- First login (1)

Without OIDC:
- Token expired or rejected (1)
- Anonymus login (1) 
I think this cover the existing 7 diagrams. 

As for your question regarding the access token:
> access token we will use that elusively until it is expired and we have no such diagram of direct access token usage.

That is true but only within the same session, as far as I remember only `poll` and `submit` will reuse the access_token as they issue more than 1 request to the server within the same session. I think that perhaps the diagrams are clear as they are as they are only showing a single request to the server using an access_token. If we were to add this logic, perhaps the diagram will add complexity that could be harder to read but please let me know what you think!

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,331 @@
...
+.. _basic-auth-is-oidc-agnostic:
+
+Basic Auth
+''''''''''
+
+Regardless of whether OIDC is enabled or not, Basic Auth will be used if both the
+``client_id`` and ``secret_key`` are provided.
+
+.. mermaid::
+    :alt: Sequence diagram showing BASIC Auth is used when client_id and secret_key are provided
```
ajzobro:
```suggestion
    :alt: Sequence diagram showing Basic Auth is used when client_id and secret_key are provided
```

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,331 @@
...
+    :caption: Sequence diagram showing BASIC Auth is used when client_id and secret_key are provided
```
ajzobro:
```suggestion
    :caption: Sequence diagram showing Basic Auth is used when client_id and secret_key are provided
```

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,331 @@
...
-queues that you need access to.
+Authentication
+--------------
 
-These credentials can be :doc:`set using the Testflinger CLI <../how-to/authentication>`. 
+Testflinger can be configured to use OpenID Connect (OIDC) for authentication. The way that
+authorization working in the Testflinger server depends on whether OIDC is enabled or not. When
+OIDC is *not* enabled, Testflinger supports anonymous logins. However, when OIDC is enabled
+each request must be authenticated. Both operating modes support use of self-signed tokens
+for authentication, and these are what are used in typical CI/CD flows.
```
rene-oromtz: I feel that this is not accurate for CI/CD flows as "self-signed tokens" is what happens under the hood. Also, I feel like we perhaps should not need to specify that those are for CI/CD as this is mostly related on how we use it. 

Perhaps more accurate would be something like this:

> Both operating modes supports the use of a Testflinger ``client_id`` and ``secret_key`` pair that can be generated by an administrator and used as part of the Basic Auth. 

This also relates directly to the next paragraph, what do you think?

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,331 @@
...
 
-Additionally, you can also login to the server by running the following command:
+Since Basic Auth can be used in both configurations, it is prioritized first, when
+Basic Auth is used it will be handled within the server before other authentication methods.
```
rene-oromtz: I think we can remove one mentioned to the Basic Auth, something like:

```suggestion
Since Basic Auth can be used in both configurations, it is prioritized and will be handled within the server before other authentication methods.
```

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,331 @@
...
+        Note over CLI,Server: POST /v1/jobs
+        Server->>Server: Validate access token signature and client permissions
+        Server-->>CLI: 200 OK / authenticated response
+        CLI-->>User: Show result
+
+Token expired or rejected
+.........................
+
+If the server does not have OIDC enabled and the refresh token is rejected,
+the server returns a 401 and the CLI removes the stored token. If the CLI does not
```
rene-oromtz: Perhaps it doesn't hurt, but I think this is now redundant given we are the "OIDC disabled" section. 

It can perhaps directly be simplified to:

```suggestion
The server returns a 401 and the CLI removes the stored token. If the CLI does not
```

This is different as the cases above where I think it has some value to say that it works regardless OIDC is enabled or not as we are referencing in the below.

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,331 @@
...
+        Server->>Server: Token expired or rejected
+        Server-->>CLI: 400 Bad Request
+        CLI->>CLI: Delete stored refresh token
+        CLI-->>User: Authentication failed
+
+No credentials or stored token
+''''''''''''''''''''''''''''''
+
+If the server does not require authentication, the request is allowed through
+as an anonymous user:
```
rene-oromtz: Given this is a OIDC disabled section, this can also be reworded/simplified but I don't think as necessarily as the other [comment](https://github.com/canonical/testflinger/pull/1161/changes#r3617801912)

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,331 @@
...
+        Note over CLI,Server: POST /v1/jobs
+        Server->>Server: Auth not required
+        Server-->>CLI: 200 OK / anonymous user response
+        CLI-->>User: Show result
+
+OIDC Enabled
+~~~~~~~~~~~~
+
+When Testflinger is configured to use OpenID Connect (OIDC) for authentication, the order is:
+When Testflinger is configured to use OpenID Connect (OIDC) for authentication, the order is:
```
rene-oromtz, line 182:
```suggestion
```

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,331 @@
...
+Refresh Token
+'''''''''''''
+
+Exactly the same as without OIDC, see :ref:`refresh-auth-is-oidc-agnostic`.
+
+Token expired or rejected
+.........................
+
+If the server has OIDC enabled and the refresh token is rejected, the server
+initiates an OIDC authentication flow:
```
rene-oromtz, line 202: Same as this [comment](https://github.com/canonical/testflinger/pull/1161/changes#r3617801912) perhaps redundant to say it has OIDC enabled  as we are already in this section

rene-oromtz (changes requested):
Just added some comments for reduce redundancy or add some clarification. Besides that LGTM

docs/explanation/authentication.rst
ajzobro:
```suggestion
each request must be authenticated.  Both operating modes support use of basic auth accounts for authentication which are suitable for use with CI/CD flows.
```

docs/explanation/authentication.rst
ajzobro: I prefer to keep the verbosity as these are long sections and being able to glance down and have the paragraphs clearly stating with OIDC, etc.  In this case, I agree that it can be re-worded, but I would only make it more technically correct:


```suggestion
If a given endpoint does not require authentication, the request is allowed through
as an anonymous user:
```

docs/explanation/authentication.rst
ajzobro: I like the verbosity, arguments above.

ajzobro:
@tang-mm Can I get you to look at this again when you have a chance, please.  I think it's looking pretty good overall.

docs/explanation/authentication.rst
ajzobro: If you want to propose dropping the second "OIDC" I can do that.  Please add your comment/suggestion

docs/explanation/authentication.rst
rene-oromtz: Alright, I can be sold that idea and the rewording makes sense and its more accurate, if you are to "Apply suggestion" just note that there is a "the" before the "a given endpoints", perhaps this is

> If a given endpoint

docs/explanation/authentication.rst
rene-oromtz: Its okay, if you prefer to keep the initial OIDC the second also makes sense. I can only see `the authentication flow` but its more explicit as-is to indicate this starts the OIDC-specific process

docs/explanation/authentication.rst
rene-oromtz: The rewording makes sense but still the "self-signed" tokens is the one its bugging me a bit. Specially because the only signed is the `access_token` and CI/CD flows don't use this

rene-oromtz (dismissed):
Thanks for the changes! LGTM! 

Could you please help us take a look @tang-mm? As both Andy and I committed to this branch we could use your expertise to see if this can be improved or is good as-is

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,327 @@
...
+        Note over CLI,Server: POST /oidc/auth-init
+        Server-->>OIDC: Proxy request to OIDC provider
+        OIDC-->>Server: device_code + user_code + URL for OIDC handshake
+        Server->>CLI: request_id + user code + URL for OIDC handshake
+        CLI->>User: Display URL and code
+        User->>OIDC: Complete the handshake
+        CLI->>Server: Poll for auth completion
+        Note over CLI,Server: POST /oidc/auth-poll/<request_id>
+        Server-->>OIDC: Is this user cool?
+        OIDC-->>Server: Yes + user identity (email)
```
tang-mm:
```suggestion
        Server-->>OIDC: Validate authorization status
        OIDC-->>Server: Authorized + user identity (email)
```
Let's update this wording to use a more formal tone

docs/explanation/authentication.rst
```diff
@@ -1,25 +1,327 @@
...
```
tang-mm:
```suggestion
        Server-->>OIDC: Validate authorization status
        OIDC-->>Server: Authorized + user identity (email)
```
same

tang-mm (dismissed):
Great work! Thank you for taking the suggestions into account and adding so many valuable updates. Just one minor comment on the wording below. Happy to see this landed :rocket:

ajzobro:
Excellent!  Thanks @tang-mm back to you for the final approval and we can get it merged if there are no other blocking issues.

rene-oromtz (approved):

PR #1165, opened 2026-06-29
Add: `WEBHOOK_DOMAINS` to support multiple webhook domains

LaiderLai:
## Description

Add: `WEBHOOK_DOMAINS` to support multiple webhook domains.

## Resolved issues

https://warthogs.atlassian.net/browse/OEIOT-846
To free the runner resources for polling the test results from the testflinger, a webhook to get the test result notification would be better.
The Mattermost incoming webhook is a good place to get notifications.

## Web service API changes

No open APIs have changed.
For backward compatibility, add `WEBHOOK_DOMAINS` to support this feature.

## Tests

 1. ✅ test_agents_status_put_with_domainlist - PASSED
 2. ✅ test_agents_status_put_with_domainlist_rejected - PASSED
 3. ✅ test_agents_status_put_domainlist_multiple_domains - PASSED

### DIFF
--- a/server/API.md
+++ b/server/API.md
@@ -448,6 +448,16 @@ Receive job status updates from an agent and posts them to the specified webhook
 The `job_status_webhook` parameter is required for this endpoint. Other
 parameters included here will be forwarded to the webhook.
 
+### Webhook URL Validation
+
+The webhook URL is validated against one of two possible configurations:
+
+1. **Multiple Mode** (recommended): Set `WEBHOOK_DOMAINS` environment variable
+   with a comma-separated list of allowed webhook domains. Example: `webhook.example.com, webhook2.example.com`
+
+2. **Legacy Mode**: If `WEBHOOK_DOMAINS` is not set, falls back to strict mode
+   where the webhook URL must match the server-configured `WEBHOOK_URL`.
+
 Parameters:
 
 - `job_id` (UUID): test job identifier
@@ -461,14 +471,19 @@ Status Codes:
 
 - `HTTP 200 (OK)`
 - `HTTP 400 (Bad request)`: The arguments could not be processed by the server
+- `HTTP 403 (Forbidden)`: Webhook URL is not in the configured domain list
+- `HTTP 502 (Bad Gateway)`: Webhook unreachable
 - `HTTP 504 (Gateway Timeout)`: The webhook URL timed out
 
-Example:
+Examples:
 
 ```shell
 curl -X POST \
   -H "Content-Type: application/json" \
-  -d '{"agent_id": "agent-00", "job_queue": "myqueue", "job_status_webhook": "http://mywebhook", "events": [{"event_name": "started_provisioning", "timestamp": "2024-05-03T19:11:33.541130+00:00", "detail": "my_detailed_message"}]}' http://localhost:8000/v1/job/00000000-0000-0000-0000-000000000000/events
+  -d '{"agent_id": "agent-00", "job_queue": "myqueue", "job_status_webhook": "http://webhook.example.com/xxxxxxxxxxxxxxxx", "events": [{"event_name": "started_provisioning", "timestamp": "2024-05-03T19:11:33.541130+00:00", "detail": "my_detailed_message"}]}' http://localhost:8000/v1/job/00000000-0000-0000-0000-000000000000/events
+
+# Multiple domains can be configured
+# WEBHOOK_DOMAINS="webhook.example.com, webhook2.example.com"
 ```
 
 ## `[GET] /v1/queues/wait_times`
--- a/server/src/testflinger/api/v1.py
+++ b/server/src/testflinger/api/v1.py
@@ -779,29 +779,26 @@ def agents_status_post(job_id, json_data):
     # Webhook specified in the job definition
     job_webhook = json_data.pop("job_status_webhook")
 
-    # Webhook base URL configured on the server
+    # Validate webhook URL against configured domain list or legacy WEBHOOK_URL
+    domain_list = os.environ.get("WEBHOOK_DOMAINS", "").strip()
     authorized_webhook_url = os.environ.get("WEBHOOK_URL")
-    if not authorized_webhook_url:
-        # Abort if no webhook configured server-side
+
+    if not domain_list and not authorized_webhook_url:
+        # Abort if no webhook configuration found
         abort(
             HTTPStatus.INTERNAL_SERVER_ERROR,
             message="Webhook URL not configured",
         )
 
-    # Parse both URLs; authorized_webhook_url is the only permitted
-    # endpoint to send requests to.
-    parsed_job_url = urlparse(job_webhook)
-    parsed_authorized_url = urlparse(authorized_webhook_url)
-
-    # Validate job_webhook is referring to the server-configured webhook URL
-    if (
-        parsed_job_url.scheme != parsed_authorized_url.scheme
-        or parsed_job_url.netloc != parsed_authorized_url.netloc
-    ):
+    # Validate job_webhook is allowed
+    if not is_webhook_url_allowed(job_webhook):
         abort(
             HTTPStatus.FORBIDDEN,
             message="Invalid job_status_webhook URL specified",
         )
+
+    # Parse job URL for later use (scheme detection for retry adapter)
+    parsed_job_url = urlparse(job_webhook)
 
     # Attempt to get optional authentication header from server configuration
     auth_headers = {}
@@ -816,9 +813,7 @@ def agents_status_post(job_id, json_data):
                     allowed_methods=frozenset(["PUT"]),
                 )
             )
-            webhook_session.mount(
-                f"{parsed_authorized_url.scheme}://", retry_adapter
-            )
+            webhook_session.mount(f"{parsed_job_url.scheme}://", retry_adapter)
             response = webhook_session.put(
                 job_webhook,
                 json=json_data,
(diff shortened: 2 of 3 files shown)


server/src/testflinger/api/v1.py
```diff
@@ -779,30 +779,27 @@ def agents_status_post(job_id, json_data):
     # Webhook specified in the job definition
     job_webhook = json_data.pop("job_status_webhook")
 
-    # Webhook base URL configured on the server
+    # Validate webhook URL against configured domain list or legacy WEBHOOK_URL
+    domain_list = os.environ.get("WEBHOOK_DOMAINS", "")
```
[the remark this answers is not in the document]
LaiderLai: Resolved.

server/API.md
```diff
@@ -461,14 +471,18 @@ Status Codes:
 
 - `HTTP 200 (OK)`
 - `HTTP 400 (Bad request)`: The arguments could not be processed by the server
+- `HTTP 403 (Forbidden)`: Webhook URL is not in the configured domain list
 - `HTTP 504 (Gateway Timeout)`: The webhook URL timed out
```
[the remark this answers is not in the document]
LaiderLai: Resolved.

server/tests/test_v1.py
```diff
@@ -1206,6 +1206,85 @@ def test_agents_status_put_webhook_auth_failure(
...
+    assert output.status_code == HTTPStatus.FORBIDDEN
+    assert "Invalid job_status_webhook URL specified" in output.get_data(
+        as_text=True
+    )
+
+
+def test_agents_status_put_domainlist_multiple_domains(
+    mongo_app, requests_mock, monkeypatch
+):
+    """Test domainslist supports multiple domains."""
```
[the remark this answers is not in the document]
LaiderLai: Resolved.

server/tests/test_v1.py
```diff
@@ -1206,6 +1206,85 @@ def test_agents_status_put_webhook_auth_failure(
     assert "Webhook authentication failed" in output.get_data(as_text=True)
 
 
+def test_agents_status_put_with_domainlist(mongo_app, requests_mock, monkeypatch):
+    """Test webhook URL validation with domanlist mode."""
```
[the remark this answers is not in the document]
LaiderLai: Resolved.

server/src/testflinger/api/v1.py
```diff
@@ -844,6 +841,37 @@ def agents_status_post(job_id, json_data):
...
+    """
+    parsed_webhook = urlparse(webhook_url)
+    webhook_domain = parsed_webhook.netloc
+
+    # Check domain list
+    domain_list = os.environ.get("WEBHOOK_DOMAINS", "")
+    if domain_list:
+        allowed_domains = [d.strip() for d in domain_list.split(",")]
+        return webhook_domain in allowed_domains
+
```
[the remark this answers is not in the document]
LaiderLai: It's out of scope.

server/src/testflinger/api/v1.py
```diff
@@ -779,30 +779,27 @@ def agents_status_post(job_id, json_data):
     # Webhook specified in the job definition
     job_webhook = json_data.pop("job_status_webhook")
 
-    # Webhook base URL configured on the server
+    # Validate webhook URL against configured domain list or legacy WEBHOOK_URL
+    domain_list = os.environ.get("WEBHOOK_DOMAINS", "").strip()
```
ajzobro, line 783: Appears we are getting the "list" as a string but we don't assert that the domains themselves are properly formatted.

If we are going to support the DOMAIN list, we don't need the current WEBHOOK_URL since they both serve the same purpose.

server/src/testflinger/api/v1.py
```diff
@@ -779,30 +779,27 @@ def agents_status_post(job_id, json_data):
     # Webhook specified in the job definition
     job_webhook = json_data.pop("job_status_webhook")
```
ajzobro, line 780: Seems like we're still only supporting one web-hook per job with this solution, is that the goal?  I thought we wanted to support multiple.

server/src/testflinger/api/v1.py
ajzobro, line 820: We need for the auth_headers to be partitioned off by domain as well -- When we had one webhook url we had one set of auth parameters.

We cannot send all domains the same auth paramters and there should be an auth parameter per domain.

server/src/testflinger/api/v1.py
```diff
@@ -844,6 +839,37 @@ def agents_status_post(job_id, json_data):
         abort(HTTPStatus.BAD_GATEWAY, message="Webhook unreachable")
 
 
+def is_webhook_url_allowed(webhook_url):
+    """Validate webhook URL against configured domain list.
+
+    Checks if the webhook URL is allowed based on:
+    1. WEBHOOK_DOMAINS env var (comma-separated list of allowed domains)
+    2. Falls back to legacy WEBHOOK_URL check for backward compatibility
```
ajzobro, line 847: Again, as we modify this, we should avoid duplication -- to support multiple possible webhook domains, we need to support multiple domains, each with restrictions on or values for:
 - uri
 - auth parameters

Avoid sending the wrong auth params to the wrong domains or the wrong uri's

server/tests/test_v1.py
```diff
@@ -1687,3 +1768,17 @@ def test_job_get_no_auth_headers(mongo_app):
...
+    # Skip: /v1/result/<job_id>/webhook is not a valid endpoint
+    # Webhook handling is done via result posting
+    pass
+
+
+def test_webhook_unreachable(mongo_app, requests_mock):
+    """Test webhook unreachable handling."""
+    # Skip: /v1/result/<job_id>/webhook is not a valid endpoint
+    # Webhook handling is done via result posting
+    pass
```
ajzobro, line 1784: While good to test the impacts of these potential third-party services timing out of giving bad statuses, the testflinger components themselves (agent/server) don't need to concern themselves with proper functionality to perform their tasking -- third-party webhook support is NOT something that should hinder testflinger agent/server performance and we should guard against it.

Optionally we could report on such failures in the job results (new section related to webhooks, scrubbed of any auth data) for debugging.

Webhooks are non-essential to testflinger's operation.

ajzobro (changes requested):
Thanks for putting forth the effort to get this started! Let's talk through some of the points I raise and how we can make this better and secure to the point where it can be used!

ajzobro:
@rene-oromtz Please review this.

I want to also consider the potential use of these "events" for first-party use and not only have webhooks benefit from the oppportunity for collecting metrics as the agents run tests.

server/src/testflinger/api/v1.py
LaiderLai: Hi @ajzobro,

Thanks for reviewing. I added `DOMAINS` without touching `URL` because I learned that' WEBHOOK_URL` is only used for an internal service.

It's good to know we can extend the `WEBHOOK_URL` to `WEBHOOK_URLS` directly.

server/src/testflinger/api/v1.py
LaiderLai: It's the same. I assume the job_webhook is only used for the internal service.
So I only tried to extend the server-side to support multiple webhook URLs.

It's also good to know that we can expand the target for the server/client sides to support multiple webhook URLs.

server/src/testflinger/api/v1.py
LaiderLai: Indeed, but it's hard to adopt all cases in this PR.
I suppose to support Mattermost (simple, only needs payload) and GitHub (with header and data) webhooks at first.

If any new cases, a PR to extend each one. What do you think?

server/src/testflinger/api/v1.py
LaiderLai: Indeed, let's focus on extending:
1.  job_webhook =>  job_webhooks
2. WEBHOOK_URLS

server/tests/test_v1.py
LaiderLai: Thanks for pointing this out. You're right. Timeout and unreachable tests will be removed when I submit a new commit to this PR.

rene-oromtz (changes requested):
Thanks @LaiderLai for putting this up and thanks @ajzobro for the initial review! 

I think Andy raised really valid concerns that needs to be addressed regarding authentication of the webhooks. Addingl to the comments he already pointed out I just wanted to mentioned that charm-side modification are also required for this PR. 

For context, the prod server is deployable via Charm which is setting all the required envvars. Given `WEBHOOK_DOMAINS` is newly added server-side, the charm also needs to define it and provide it to the underlying app environment. 

Additionally for authentication, this is the important part because you will need to add a mapping to indicate webhookA needs authA. 
Perhaps you will need to define something as `WEBHOOK_AUTH_MAP` that is configurable at charm level as a JSON string e.g.
`webhook_auth_map: '{"test-observer.example.com": "token"}'`

This way you add add the tokens that are required for specific webhooks and ignore from this mapping if they don't require auth. 

Relevant files for this purpose: [charmcraft.yaml](https://github.com/canonical/testflinger/blob/main/server/charm/charmcraft.yaml) (definition), [config.py](https://github.com/canonical/testflinger/blob/main/server/charm/src/config.py) (validation), [charm.py](https://github.com/canonical/testflinger/blob/main/server/charm/src/charm.py) (passing envvars to app)

Please feel free to reach out if any questions remain!

server/src/testflinger/api/v1.py
rene-oromtz: Just added a comment about it but if this is changes to `WEBHOOK_URLS` the same modification is needed charm side.

PR #1167, opened 2026-06-29
refactor(agent): add authorization headers to all requests.

rene-oromtz:
## Description
This PR adds authentication header to all requests by using a shared requests.Session. 

Implementation:

1. Uses the `__call__` method from `AuthBase` to add the access token as authorization header
2. Adds this header to the `session.auth` object 
3. Caches the access token (JWT token) so it can be reused with the same session to reduce network calls. 
4. On access token expiration, it automatically handles token expiration by refreshing the access token and reattempting the same request that initially failed. 
5. session cookies are now implicit, this happens at agent registration and is guaranteed to be kept given the agent now uses the same session for all requests. 
6. Also reduces updating the refresh token to 3 days given the expiration is now 6 days server side

## Resolved issues
Resolves [CERTTF-1234](https://warthogs.atlassian.net/browse/CERTTF-1234)

## Tests
Existing unit tests are passing
session.cookies unit tests are cleaned up given this is now implicit, no cookies refresh is happening agent side.

[CERTTF-1234]: https://warthogs.atlassian.net/browse/CERTTF-1234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/src/testflinger_agent/client.py
+++ b/agent/src/testflinger_agent/client.py
@@ -146,7 +146,8 @@ class TestflingerClient:
             if getattr(response.request, "_auth_retry", False):
                 return response
 
-            # Drain before releasing the connection
+            # Consume the response body to return the active connection to
+            # the pool for reuse
             _ = response.content
 
             del self.access_token
--- a/agent/src/testflinger_agent/client.py
+++ b/agent/src/testflinger_agent/client.py
@@ -69,8 +69,8 @@ class ClientAuth(AuthBase):
         :param req: The prepared request object to modify
         :return: The modified request object with Authorization header
         """
-        if self.client.access_token:
-            req.headers["Authorization"] = f"Bearer {self.client.access_token}"
+        if access_token := self.client.access_token:
+            req.headers["Authorization"] = f"Bearer {access_token}"
         return req
 
 
@@ -147,15 +147,15 @@ class TestflingerClient:
                 return response
 
             # Drain before releasing the connection
-            response.content  # noqa: B018
+            _ = response.content
 
             del self.access_token
-            if not self.access_token:
+            if not (access_token := self.access_token):
                 return response
 
             new_request = response.request.copy()
             new_request.headers["Authorization"] = (
-                f"Bearer {self.access_token}"
+                f"Bearer {access_token}"
             )
             new_request._auth_retry = True
             new_response = response.connection.send(new_request, **kwargs)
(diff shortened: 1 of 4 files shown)


agent/src/testflinger_agent/client.py
```diff
@@ -102,21 +130,38 @@ def _configure_influx(self):
...
+        if response.status_code == HTTPStatus.UNAUTHORIZED:
+            if getattr(response.request, "_auth_retry", False):
+                return response
+
+            # Consume the response body to return the active connection to
+            # the pool for reuse
+            _ = response.content
+
+            del self.access_token
+            if not (access_token := self.access_token):
```
ajzobro, line 154: We are deleting and then using self.access_token.

agent/tests/test_client.py
```diff
@@ -644,111 +623,38 @@ def test_get_job_missing_token_header(self, client, requests_mock):
...
-            f"{client.server}/v1/oauth2/refresh",
-            json={"access_token": "test-token"},  # noqa: S106
-        )
-        requests_mock.post(
-            f"{client.server}/v1/agents/data/test_agent",
-            json={"job_id": ""},
-        )
 
-        # Cookies are empty, so post_agent_data should be called
-        assert not client.session.cookies
```
ajzobro, line 715: Are we still going to have cookies? I believe the server is still expecting cookies for subsequent agent identification and that we should still be checking appropriate and expected cookie bahavior.

agent/src/testflinger_agent/client.py
ajzobro: Ah, this must just clear the cached value.

agent/tests/test_client.py
rene-oromtz: Correct, agent will have its cookies on agent registration and every time it gets a job via `post_agent_data`. Given we are using the same session for all requests, the cookies are guaranteed (afaik those are not expiring server side)

agent/src/testflinger_agent/client.py
rene-oromtz: Correct, deleting the cached valued and re-cache again

ajzobro (approved):
LGTM, let's go try this out on staging.

PR #1168, opened 2026-06-29
refactor(cli): add authorization headers to all requests

rene-oromtz:
## Description
This PR adds authentication header to all CLI requests by using a shared requests.Session.

Implementation:
1. This decouples a circular dependency where the Auth module required the client module and viceversa
2. CLI auth module now uses a bare requests method call given its not necessary to inject Authentication headers. 
3. The client module now requires the CLI Auth as an "auth manager" this is so the client can handle all requests (via a shared session) while the auth module is in charge of performing the logic based on authentication mechanism.
4. Given this centralized the Auth header (via a subclass of `AuthBase`), there is no need to include headers in any requests method or build it manually. 
5. The JWT token was already cached so we are just ensuring to actually use it on all requests. In case of expiration, refresh is made automatically via session hooks. 
6. This also increases the access token expiration to 10 minutes given 30 seconds is to short. This reduces the amount of network calls that will be required for agents/users to refresh the access tokens.  

## Resolved issues
Resolves [CERTTF-1235](https://warthogs.atlassian.net/browse/CERTTF-1235)

## Tests
Existing unit tests are passing but needed to be modified. 
- all: now uses a centralized URL from conftest
- client test: this is the one that was modified more heavily given client now requires an auth manager. This now is using a shared fixture

[CERTTF-1235]: https://warthogs.atlassian.net/browse/CERTTF-1235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/testflinger_cli/auth.py
+++ b/cli/testflinger_cli/auth.py
@@ -16,18 +16,19 @@
 
 """Testflinger CLI Auth module."""
 
+import base64
 import configparser
-import contextlib
 import time
+import urllib
 import webbrowser
 from functools import cached_property, wraps
 from http import HTTPStatus
 from typing import Optional
 
 import jwt
+import requests
 from xdg_base_dirs import xdg_config_home
 
-from testflinger_cli import client
 from testflinger_cli.enums import ServerRoles
 from testflinger_cli.errors import (
     AuthenticationError,
@@ -36,19 +37,22 @@ from testflinger_cli.errors import (
     OidcError,
 )
 
+DEFAULT_AUTH_TIMEOUT = 15  # seconds
+
 
 class TestflingerCliAuth:
     """Class to handle authentication and authorisation for Testflinger CLI."""
 
     def __init__(
         self,
-        tf_client: client.Client,
+        server_url: str,
         client_id: str | None = None,
         secret_key: str | None = None,
     ):
-        self.client_id: Optional[str] = client_id
+        """Initialize the Testflinger CLI Auth class."""
+        self.server_url = server_url
+        self.client_id = client_id
         self.secret_key = secret_key
-        self.client = tf_client
 
         # Refresh token relevant configuration
         config_home = xdg_config_home()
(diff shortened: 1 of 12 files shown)


cli/testflinger_cli/auth.py
```diff
@@ -98,39 +102,65 @@ def authenticate(self) -> str | None:
...
-        """Authenticate using client_id and secret_key."""
+        """Authenticate using client_id and secret_key.
+
+        :return: string with access token (jwt token)
+        """
+        url = urllib.parse.urljoin(self.server_url, "/v1/oauth2/token")
+        id_key_pair = f"{self.client_id}:{self.secret_key}"
+        encoded_id_key_pair = base64.b64encode(
+            id_key_pair.encode("utf-8")
+        ).decode("utf-8")
```
ajzobro: The result of base64 encoding ought not contain utf-8 characters making ASCII the right thing to decode to but it hardly matters because it won't contain non-ascii chars

cli/testflinger_cli/auth.py
```diff
@@ -98,39 +102,65 @@ def authenticate(self) -> str | None:
...
+
         try:
-            response = self.client.refresh_authentication(refresh_token)
-            return response["access_token"]
-        except client.HTTPError as exc:
-            if exc.status == HTTPStatus.BAD_REQUEST:
+            response = requests.post(
+                url,
+                json={"refresh_token": refresh_token},
+                timeout=DEFAULT_AUTH_TIMEOUT,
```
ajzobro, line 157: I think this applies elsewhere also:
We account for a timeout in the call, but we don't catch `except requests.exceptions.Timeout:` to be able to handle it.

cli/testflinger_cli/auth.py
```diff
@@ -98,39 +102,65 @@ def authenticate(self) -> str | None:
...
+            )
+            response.raise_for_status()
+            response_data = response.json()
+            return response_data["access_token"]
+        except requests.exceptions.HTTPError as exc:
+            if exc.response.status_code == HTTPStatus.BAD_REQUEST:
                 self.clear_refresh_token()
```
ajzobro, line 164: handle timeouts; place this into finally.

cli/testflinger_cli/auth.py
```diff
@@ -233,16 +262,23 @@ def get_user_role(self) -> str | None:
 
     def _authenticate_with_oidc(self) -> str | None:
         """Authenticate using OIDC device flow."""
+        auth_init_url = urllib.parse.urljoin(
+            self.server_url, "/oidc/auth-init"
+        )
```
ajzobro: We could wrap this with a method:

```suggestion
        auth_init_url = self.get_url("/oidc/auth-init")
```

Why? it would allow for a re-usable pattern to be used in-line everywhere that a url may appear and it will ensure that the server host is always accounted for.

{minor, goldplating}

cli/testflinger_cli/client.py
```diff
@@ -47,13 +48,72 @@ def __init__(self, status, msg="", headers=None, error_code=None):
...
+
+        When the server signals token expiry with a 401, this hook
+        invalidates the cached token, fetches a fresh one, and replays
+        the original exactly once.
+
+        :param response: The response object from the request
+        :return: The replayed response on expiry, otherwise the original
+        """
+        if response.status_code == HTTPStatus.UNAUTHORIZED:
+            if getattr(response.request, "_auth_retry", False):
```
ajzobro, line 99: Add comment about mysterious "_auth_retry" and where it is set

cli/tests/test_client.py
```diff
@@ -20,23 +20,22 @@
...
-def test_get_error_threshold(caplog, requests_mock):
+def test_get_error_threshold(caplog, requests_mock, client):
     """Test that a warning is logged when error_threshold is reached."""
     caplog.set_level(logging.WARNING)
-    requests_mock.get(
-        "http://testflinger/test", exc=requests.exceptions.ConnectionError
-    )
-    client = Client("http://testflinger", error_threshold=3)
+    requests_mock.get(f"{URL}/test", exc=requests.exceptions.ConnectionError)
+    client.error_threshold = 3
```
ajzobro: isn't this the default?

rene-oromtz: Yep, just removed it as this is not needed

cli/testflinger_cli/auth.py
rene-oromtz: All auth routes are now using `self.get_url`

cli/testflinger_cli/auth.py
rene-oromtz: Added timeout handling, but the finally is not needed, the refresh token only need to be removed on BAD_REQUEST as its no longer valid. Also the finally will remove it on success

ajzobro (approved):
Thanks for the updates!  I'm happy to see those sys.exit calls get removed in favor of better exceptions.

Issue #292, opened 2024-06-26
Anyone can cancel any job


joalif:
Currently, I can go here https://testflinger.canonical.com/jobs and cancel any job.
I'm not saying that people will do it on purpose but it could be accidental, e.g. I want to cancel a job of mine and accidentally copy another job id.

PR #1169, opened 2026-06-30
breaking+feature/api lockdown when OIDC enabled

ajzobro:
# Lock down the v1 API with role-based authorization

## Summary

This PR makes role-based access control mandatory across the Testflinger v1
API. Previously many endpoints were unauthenticated; now every protected
endpoint is guarded by an explicit `@require_role(...)`, and the role model
behaves consistently whether or not OIDC is enabled. It also fixes a 500-on-
auth bug for OIDC-registered clients and removes the deprecated legacy
result/output endpoints and schemas.

## Motivation

The v1 API exposed mutating and data-read endpoints (jobs, results,
attachments, agents, queues, images, etc.) with no authorization. This change
ties each endpoint+method to the roles allowed to use it, closing those gaps
and giving us a single, testable source of truth for API permissions.

### Authorization model (`api/auth.py`)
- **Anonymous access** is now treated as `CONTRIBUTOR` **only when OIDC is
  disabled**. When OIDC is enabled, unauthenticated requests to protected
  endpoints are rejected (`401`).
- `require_role` enforces role checks when OIDC is enabled and the caller is
  unauthenticated / has no `client_id`.
- **Bug fix:** OIDC-registered clients have no `client_secret_hash` (they
  authenticate via the web flow). Credential-based login at
  `/v1/oauth2/token` now rejects them with `401` instead of raising a `500`.
- Bearer-token parsing tidied so a missing token no longer assumes a stripped
  prefix.

### Endpoint protection (`api/v1.py`)
- Added `@authenticate` + `@require_role(...)` to all protected v1 endpoints
  (jobs, results, artifacts, logs, attachments, events, agents data/queues/
  images, provision logs, job position/search, queue listings, secrets).
- `secrets` PUT/DELETE now return `401` (not `403`) when no authenticated
  `client_id` is present.

### Role enum (`common/.../enums.py`)
- `ServerRoles._missing_` enables case-insensitive role lookup (e.g.
  `"contributor"` → `ServerRoles.CONTRIBUTOR`).
- Clarified docstring on how the role hierarchy is (and isn't) used.

### Legacy removal
- Removed legacy endpoints `GET/POST /v1/result/<job_id>/output` and
  `GET/POST /v1/result/<job_id>/serial_output`.
- Removed the polymorphic `ResultSchema`/`ResultLegacy` and renamed
  `ResultPost` → `ResultSchema`. `result_get` no longer special-cases the
  legacy on-disk format.
- Removed the `POST /v1/client-permissions` collection endpoint from docs.
- Updated `API.md` and `schemas/openapi.json` accordingly.

### Tests
- **New permissions matrix:** `tests/permissions.json` declares the roles
  allowed for each endpoint+method, and `tests/test_v1_authorization_roles.py`
  parametrizes over it to assert, for both OIDC-enabled and OIDC-disabled
  modes, that allowed roles get `200`, forbidden roles get `403`, and
  anonymous gets `401`/CONTRIBUTOR as appropriate.
- **Coverage guard:** `test_all_v1_endpoints_have_permissions_coverage`
  enumerates the live URL map and fails if any `/v1` route+method is missing
  from `permissions.json`, unless it is listed in
  `DOCUMENTED_PERMISSION_EXCEPTIONS`. The exception set documents the only
  endpoints intentionally excluded from the role matrix (they have no role
  gate by design): `GET /v1/`, `POST /v1/oauth2/token`, and
  `POST /v1/oauth2/refresh`.
- **Shared helpers:** new `tests/utilities.py`
  (`get_access_token_header`, `get_basic_auth_header`, `get_access_token`,
  `get_refresh_token`) replaces ad-hoc auth helpers duplicated across test
  modules.
- **Fixtures:** added `app_with_store`, `role_clients_factory`, and
  `webhook_fixture`; `JWT_SIGNING_KEY` is now set consistently in fixtures.
- Updated existing tests to send `Authorization: Bearer <token>` (previously
  the raw token) and to pass agent auth headers where now required.
- Removed tests for the deleted legacy endpoints/format.

## Behavioral / breaking changes

- Clients that previously called protected v1 endpoints anonymously will now
  receive `401`/`403` when OIDC is enabled. With OIDC disabled, anonymous
  callers retain the `CONTRIBUTOR` role.
- The `Authorization` header must use the `Bearer ` scheme for access tokens.
- Legacy `output` / `serial_output` result endpoints are gone; use
  `/v1/result/<job_id>/log/<log_type>`.

## Testing

- `just lint server` and `just format server` pass clean.
- New and updated unit tests cover the full role matrix (OIDC on/off), the
  coverage guard, and the OIDC-client token `401` regression.

# Table Overview

Method | URI | AGENT | CONTRIBUTOR | MANAGER | ADMIN | Used/Delete
-- | -- | -- | -- | -- | -- | --
GET | / | x | * | * | * | *ping, Anonymous OK!
POST | /agents/provision_logs/<agent_name> | x |   |   |   |  
GET | /job | x |   |   |   |  
GET | /job/<job_id>/attachments | x |   |   |   |  
POST | /job/<job_id>/events | x |   |   |   | one-way
POST | /result/<job_id> | x |   |   |   |  
POST | /result/<job_id>/artifact | x |   |   |   |  
POST | /result/<job_id>/log/<log_type:log_type> | x |   |   |   |  
POST | /agents/data/<agent_name> | x |   | * | x | *needed in MANAGER role for LAB MAINT
POST | /agents/images | x |   | * | x | *needed in MANAGER role for LAB MAINT
POST | /agents/queues | x |   | * | x | *needed in MANAGER role for LAB MAINT
GET | /agents/data/<agent_name> | x | x | x | x |  
POST | /oauth2/refresh | x | x | x | x | MANAGER = CONTRIBUTOR + client-permissions (not agents!)
POST | /oauth2/token | * | x | x | x | *only called by agent charm, could leave agent stranded
GET | /result/<job_id> | x | x | x | x |  
GET | /agents/data |   | x | x | x |  
GET | /agents/images/<queue> |   | x | x | x |  
GET | /agents/queues |   | x | x | x |  
GET | /queues/<queue_name>/agents |   | x | x | x | get agents for queue
GET | /queues/<queue_name>/jobs |   | x | x | x |  
GET | /queues/wait_times |   | x | x | x | not used outside of tests
POST | /job | * | x | x | x | *Note: multidevice will need to solve how to act as a contributor from the device connector perspective to avoid exposing AGENT credentials. Marian Gasparovic is the main user of this for now and may be actively depending on it
GET | /job/<job_id> |   | x | x | x |  
POST | /job/<job_id>/action |   | x | x | x | e.g. cancel job
POST | /job/<job_id>/attachments |   | x | x | x | one-way
GET | /job/<job_id>/position |   | x | x | x | not accurate or terribly useful information due to queue layout
GET | /job/search |   | x | x | x |  
GET | /result/<job_id>/artifact |   | x | x | x |  
GET | /result/<job_id>/log/<log_type:log_type> |   | x | x | x |  
DELETE | /secrets/<client_id>/<path> | * | x | x | x | * not restricted today, not needed at all, however.
PUT | /secrets/<client_id>/<path> | * | x | x | x | * not restricted today, not needed at all, however.
GET | /restricted-queues |   | x | x | x | C3 / lab manager is the user of this
GET | /restricted-queues/<queue_name> |   | x | x | x | C3 / lab manager is the user of this
GET | /client-permissions/<client_id> |   |   | x | x |  
GET | /client-permissions |   |   | x | x |  
PUT | /client-permissions/<client_id> |   |   | x | x |  
DELETE | /restricted-queues/<queue_name> |   |   | * | x | *needed in MANAGER role for LAB MAINT
POST | /restricted-queues/<queue_name> |   |   | * | x | *needed in MANAGER role for LAB MAINT
DELETE | /client-permissions/<client_id> |   |   |   | x | Today only ADMIN
POST | /oauth2/revoke |   |   |   | x | Today only ADMIN
GET | /result/<job_id>/output |   |   |   |   | removed as part of this activity
POST | /result/<job_id>/output |   |   |   |   | removed as part of this activity
GET | /result/<job_id>/serial_output |   |   |   |   | removed as part of this activity
POST | /result/<job_id>/serial_output |   |   |   |   | removed as part of this activity

### DIFF
--- a/server/src/testflinger/api/auth.py
+++ b/server/src/testflinger/api/auth.py
@@ -371,8 +371,8 @@ def issue_tokens(
     :param refresh_token_ttl: Optional expiration time for refresh token
     :return: Dict with access_token, token_type, expires_in, refresh_token
     """
-    secret_key = os.environ.get("JWT_SIGNING_KEY")
-    access_token = generate_access_token(allowed_resources, secret_key)
+    signing_key = os.environ.get("JWT_SIGNING_KEY")
+    access_token = generate_access_token(allowed_resources, signing_key)
 
     role = ServerRoles(allowed_resources.get("role", ServerRoles.CONTRIBUTOR))
     refresh_expires_in = (
@@ -384,13 +384,6 @@ def issue_tokens(
         client_id, expires_in=refresh_expires_in
     )
 
-    current_app.owasp_logger.authn_login_success(
-        userid=client_id,
-        description=(
-            f"Client {client_id} successfully authenticated with role {role}"
-        ),
-        **OWASPLogger.get_request_metadata(request),
-    )
     current_app.owasp_logger.authn_token_created(
         userid=client_id,
         entitlements=[role],
--- a/server/src/testflinger/api/auth.py
+++ b/server/src/testflinger/api/auth.py
@@ -260,23 +260,31 @@ def authenticate(func):
         g.permissions = {}
         g.is_authenticated = False
 
-        # If no token, continuing as regular user.
-        if not auth_token:
+        # Anonymous login IS allowed when OIDC is not enabled:
+        if current_app.oauth is None and not auth_token:
+            g.client_id = None  # Expressly None; no secrets allowed
+            g.role = ServerRoles.CONTRIBUTOR
+            g.permissions = {}
+            # Role and permissions are known; client_id can be trusted (None),
+            # hence "is_authenticated" is True
+            g.is_authenticated = True
             return func(*args, **kwargs)
-
-        if auth_token.startswith("Bearer "):
-            auth_token = auth_token[len("Bearer ") :]
+        # else, role will be None until we figure out who they are
 
         # If there is a token available, attempt to retrieve information
-        secret_key = os.environ.get("JWT_SIGNING_KEY")
-        decoded_jwt = decode_jwt_token(auth_token, secret_key)
-        permissions = decoded_jwt.get("permissions", {})
+        if auth_token:
+            auth_token = auth_token[len("Bearer ") :]
+            secret_key = os.environ.get("JWT_SIGNING_KEY")
+            decoded_jwt = decode_jwt_token(auth_token, secret_key)
+            permissions = decoded_jwt.get("permissions", {})
 
-        # Store auth state if decoding was successful
-        g.client_id = permissions["client_id"]
-        g.role = ServerRoles(permissions.get("role", ServerRoles.CONTRIBUTOR))
-        g.permissions = permissions
-        g.is_authenticated = True
+            # Store auth state if decoding was successful
+            g.client_id = permissions["client_id"]
+            g.role = ServerRoles(
+                permissions.get("role", ServerRoles.CONTRIBUTOR)
+            )
+            g.permissions = permissions
+            g.is_authenticated = True
 
         return func(*args, **kwargs)
 
@@ -289,7 +297,9 @@ def require_role(*roles):
     def decorator(func):
         @wraps(func)
         def wrapper(*args, **kwargs):
-            if not g.is_authenticated:
+            if current_app.oauth is not None and (
+                not g.is_authenticated or g.client_id is None
+            ):
                 current_app.owasp_logger.authz_fail(
                     userid="unauthenticated",
                     resource=request.path,
--- a/server/src/testflinger/api/auth.py
+++ b/server/src/testflinger/api/auth.py
@@ -274,8 +274,8 @@ def authenticate(func):
             return func(*args, **kwargs)
         # else, role will be None until we figure out who they are
 
-        # If there is a token available, attempt to retrieve information
-        if auth_token:
+        # If there is a bearer token available, attempt to retrieve information
+        if auth_token and auth_token.startswith("Bearer "):
             auth_token = auth_token[len("Bearer ") :]
             secret_key = os.environ.get("JWT_SIGNING_KEY")
             decoded_jwt = decode_jwt_token(auth_token, secret_key)
--- a/server/src/testflinger/api/auth.py
+++ b/server/src/testflinger/api/auth.py
@@ -66,9 +66,12 @@ def validate_client_key_pair(client_id: str, client_key: str) -> dict | None:
     client_key_bytes = client_key.encode("utf-8")
     client_permissions_entry = database.get_client_permissions(client_id)
 
-    if not client_permissions_entry or not bcrypt.checkpw(
+    # OIDC-registered clients have no client_secret_hash since they
+    # authenticate through the web flow, so reject credential-based logins
+    secret_hash = (client_permissions_entry or {}).get("client_secret_hash")
+    if not secret_hash or not bcrypt.checkpw(
         client_key_bytes,
-        client_permissions_entry["client_secret_hash"].encode("utf8"),
+        secret_hash.encode("utf8"),
     ):
         return None
 
--- a/server/src/testflinger/api/auth.py
+++ b/server/src/testflinger/api/auth.py
@@ -31,6 +31,19 @@ from testflinger_common.enums import ServerRoles
 from testflinger import database
 from testflinger.owasp import OWASPLogger
 
+DEFAULT_REFRESH_TOKEN_EXPIRATION = 6 * 24 * 60 * 60  # 6 days in seconds
+
+# Fields from client_permissions to include in the Testflinger JWT
+PERMISSIONS_FIELDS = frozenset(
+    {
+        "role",
+        "max_priority",
+        "allowed_queues",
+        "max_reservation_time",
+        "client_id",
+    }
+)
+
 
 def hash_secret(secret: str):
     """Securely hash a secret before storing in database.
@@ -317,13 +330,12 @@ def require_role(*roles):
     return decorator
 
 
-def generate_refresh_token(client_id: str, expires_in: int | None) -> str:
+def generate_refresh_token(client_id: str, expires_in: int) -> str:
     """
     Generate opaque string as a new refresh token.
 
     :param client_id: Client ID associated with this refresh token.
-    :param expires_in: Expiration time in seconds (default 30 days).
-                       Set to None for non-expiring token.
+    :param expires_in: Expiration time in seconds.
 
     :return: Refresh token string.
     """
@@ -333,14 +345,53 @@ def generate_refresh_token(client_id: str, expires_in: int | None) -> str:
         "client_id": client_id,
         "refresh_token": refresh_token,
         "issued_at": now,
-        "expires_at": None
-        if expires_in is None
-        else now + timedelta(seconds=expires_in),
+        "expires_at": now + timedelta(seconds=expires_in),
         "revoked": False,
         "last_accessed": now,
     }
     database.add_refresh_token(token_data)
     return refresh_token
+
+
+def issue_tokens(
+    client_id: str,
+    allowed_resources: dict,
+    refresh_token_ttl: int = DEFAULT_REFRESH_TOKEN_EXPIRATION,
+) -> dict:
+    """Issue Testflinger access and refresh tokens for an authenticated client.
+
+    Handles token generation, refresh token expiration, and OWASP logging.
+    Used by both credential-based and OIDC-based authentication flows.
+
+    :param client_id: Testflinger client ID for which to issue tokens
+    :param allowed_resources: Permissions dict to encode into the JWT
+    :param refresh_token_ttl: Expiration time in seconds for the refresh token
+    :return: Dict with access_token, token_type, expires_in, refresh_token
+    """
+    signing_key = os.environ.get("JWT_SIGNING_KEY")
+    access_token = generate_access_token(allowed_resources, signing_key)
+
+    role = ServerRoles(allowed_resources.get("role", ServerRoles.CONTRIBUTOR))
+    refresh_token = generate_refresh_token(
+        client_id, expires_in=refresh_token_ttl
+    )
+
+    current_app.owasp_logger.authn_token_created(
+        userid=client_id,
+        entitlements=[role],
+        description=(
+            f"JWT access token and refresh token issued for "
+            f"client {client_id} with role: {role}."
+        ),
+        **OWASPLogger.get_request_metadata(request),
+    )
+
+    return {
+        "access_token": access_token,
+        "token_type": "Bearer",
+        "expires_in": 30,
+        "refresh_token": refresh_token,
+    }
 
 
 def validate_refresh_token(token: str) -> dict:
--- a/server/src/testflinger/api/auth.py
+++ b/server/src/testflinger/api/auth.py
@@ -31,6 +31,19 @@ from testflinger_common.enums import ServerRoles
 from testflinger import database
 from testflinger.owasp import OWASPLogger
 
+DEFAULT_REFRESH_TOKEN_EXPIRATION = 6 * 24 * 60 * 60  # 6 days in seconds
+
+# Fields from client_permissions to include in the Testflinger JWT
+PERMISSIONS_FIELDS = frozenset(
+    {
+        "role",
+        "max_priority",
+        "allowed_queues",
+        "max_reservation_time",
+        "client_id",
+    }
+)
+
 
 def hash_secret(secret: str):
     """Securely hash a secret before storing in database.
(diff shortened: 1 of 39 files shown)


server/src/testflinger/api/schemas.py
```diff
@@ -461,11 +461,12 @@ class ResultGet(Schema):
     cleanup_status = fields.Integer(required=False)
     cleanup_output = fields.String(required=False)
     cleanup_serial = fields.String(required=False)
+
     device_info = fields.Dict(required=False)
     job_state = fields.String(required=False)
 
 
-class ResultPost(Schema):
+class ResultSchema(Schema):
```
rene-oromtz: I personally prefer the `ResultPost` given we have a `ResultGet` and the docstrings matches more clearly but we can use either one. It is only used once in `v1.py` so the changes are minimal. I'll let you decide based on your personal preference

server/src/testflinger/api/v1.py
```diff
@@ -392,6 +407,8 @@ def artifacts_get(job_id):
     :return:
         send_file stream of artifact tarball to download
     """
+    # TODO: consider restrictions to artifacts to original job owner (or group?
```
rene-oromtz, line 408: Both of this TODO are actually a really nice idea! Thanks for adding it here so we can revisit this later

server/src/testflinger/api/v1.py
```diff
@@ -485,6 +506,8 @@ def log_post(job_id: str, log_type: LogType, json_data: dict) -> str:
 
 
 @v1.post("/result/<job_id>")
+@authenticate
+@require_role(ServerRoles.AGENT)
 @v1.input(schemas.ResultSchema, location="json")
```
rene-oromtz: If you choose to use `ResultPost` this is the only place where it needs modification afaik

server/src/testflinger/api/v1.py
```diff
@@ -506,6 +529,13 @@ def result_post(job_id: str, json_data: dict) -> str:
 
 
 @v1.get("/result/<job_id>")
+@authenticate
+@require_role(
+    ServerRoles.ADMIN,
+    ServerRoles.MANAGER,
+    ServerRoles.CONTRIBUTOR,
+    ServerRoles.AGENT,
+)
```
rene-oromtz: Given this is applicable to all roles, perhaps this will make it a bit cleaner

```suggestion
@authenticate
@require_role(*ServerRoles)
```

server/src/testflinger/api/v1.py
```diff
@@ -289,6 +294,8 @@ def attachment_get(job_id):
     :return:
         send_file stream of attachment tarball to download
     """
+    # TODO: if we want attachments to be downloadable by the job owner,
+    #      consider similar TODO treatment as for artifacts: job page
```
rene-oromtz: Given this is Agent only, perhaps is not entirely needed, Perhaps we can assume that given this are the attachments from the user, the user already have them?

server/src/testflinger/api/v1.py
```diff
@@ -521,24 +551,22 @@ def result_get(job_id: str):
...
+@authenticate
+@require_role(ServerRoles.ADMIN, ServerRoles.MANAGER, ServerRoles.CONTRIBUTOR)
 @v1.input(schemas.ActionIn, location="json")
 def action_post(job_id, json_data):
     """Take action on the job status for a specified job ID.
 
     :param job_id:
         UUID as a string for the job
     """
+    # TODO: limit to job owner (or greater) if auth enabled (so job owner known
```
rene-oromtz: yes, please!
Also, perhaps we can reference this old bug here as comment?: https://github.com/canonical/testflinger/issues/292
(I just reopen it)
Also, seems the TODO is cropped?

server/src/testflinger/api/v1.py
```diff
@@ -647,6 +685,13 @@ def agents_get_all():
 
 
 @v1.get("/agents/data/<agent_name>")
+@authenticate
+@require_role(
+    ServerRoles.ADMIN,
+    ServerRoles.MANAGER,
+    ServerRoles.CONTRIBUTOR,
+    ServerRoles.AGENT,
+)
```
rene-oromtz: Same as the other comment, perhaps more clean with:

```suggestion
@require_role(*ServerRoles)
```

server/src/testflinger/api/v1.py
```diff
@@ -1412,12 +1471,15 @@ def delete_client_permissions(client_id: str) -> str:
...
 @authenticate
+@require_role(ServerRoles.CONTRIBUTOR, ServerRoles.MANAGER, ServerRoles.ADMIN)
 @v1.input(schemas.SecretIn, location="json")
 @v1.output(schemas.SecretOut)
 def secrets_put(client_id, path, json_data):
     """Store a secret value for the specified client_id and path."""
     if current_app.secrets_store is None:
         abort(HTTPStatus.BAD_REQUEST, message="No secrets store")
+    if not g.client_id:
+        abort(HTTPStatus.UNAUTHORIZED)
```
rene-oromtz: Perhaps we should add a message? I mean, its pretty obvious the Unauthorized but perhaps a more meaningful message would make API clearer (the CLI already handles missing client_id)

server/src/testflinger/api/v1.py
```diff
@@ -1449,10 +1511,13 @@ def secrets_put(client_id, path, json_data):
 
 @v1.delete("/secrets/<client_id>/<path:path>")
 @authenticate
+@require_role(ServerRoles.CONTRIBUTOR, ServerRoles.MANAGER, ServerRoles.ADMIN)
 def secrets_delete(client_id, path):
     """Remove a secret value for the specified client_id and path."""
     if current_app.secrets_store is None:
         abort(HTTPStatus.BAD_REQUEST, message="No secrets store")
+    if not g.client_id:
+        abort(HTTPStatus.UNAUTHORIZED)
```
rene-oromtz: Same as the other comment about adding a message

server/tests/test_results.py
```diff
@@ -139,78 +148,19 @@ def test_job_get_result_no_data(mongo_app):
...
-
-    # Retrieve results and verify it matches what was posted
-    response = app.get(f"/v1/result/{job_id}")
-    result_data = response.json
-    for key, value in legacy_data.items():
-        assert result_data[key] == value
-
-
-def test_result_post_payload_too_large(mongo_app, monkeypatch):
+def test_result_post_payload_too_large(mongo_app, agent_auth_header):
```
rene-oromtz, line 151: I think this whole test can also be removed. This was needed because the old post result could send a bunch of data. Right now, the job_state is a simple string with the state and the device_info is also guarantee to be a very small payload

server/tests/utilities.py
```diff
@@ -0,0 +1,81 @@
...
+import jwt
+from testflinger_common.enums import ServerRoles
+
+
+def get_access_token_header(client_id: str, role: ServerRoles) -> dict:
+    """Create a Bearer token header for the specified role."""
+    secret_key = os.environ.get("JWT_SIGNING_KEY")
+    expiration_time = datetime.now(timezone.utc) + timedelta(
+        seconds=30000000
+    )  # TODO: extended timeout for debug
```
rene-oromtz: Is this TODO needed? Or was it just used while debugging?

server/tests/test_v1_authorization_roles.py
```diff
@@ -0,0 +1,643 @@
...
+from pathlib import Path
+
+import pytest
+from testflinger_common.enums import ServerRoles
+
+from tests.utilities import get_access_token_header, get_refresh_token
+
+ALL_ROLES = {
+    ServerRoles(r) for r in ("AGENT", "CONTRIBUTOR", "MANAGER", "ADMIN")
+}
```
rene-oromtz: Just in case we ever add a new role this makes it cleaner:

```suggestion
ALL_ROLES = set(ServerRoles)
```

server/src/testflinger/api/auth.py
```diff
@@ -260,23 +263,31 @@ def decorator(*args, **kwargs):
...
+            decoded_jwt = decode_jwt_token(auth_token, secret_key)
+            permissions = decoded_jwt.get("permissions", {})
+
+            # Store auth state if decoding was successful
+            g.client_id = permissions["client_id"]
+            g.role = ServerRoles(
+                permissions.get("role", ServerRoles.CONTRIBUTOR)
+            )
+            g.permissions = permissions
+            g.is_authenticated = True
```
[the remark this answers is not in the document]
ajzobro: The only acceptable type of authentication allowed is 'Bearer ' and yes, it should be included in a check; But if not found to be what is expected, we must proceed as if the token is not available.

server/src/testflinger/api/schemas.py
ajzobro: revert to `RestultPost`

server/src/testflinger/api/v1.py
ajzobro: remove TODOs

server/src/testflinger/api/v1.py
ajzobro: I need to check how and where anonymous logins are handled.
`/secrets/` are the only thing that required a client_id to exist above and beyond also having a role. Enforcement of a role (and/or allowing anonymous) is done in `require_role` and so I believe we need to keep this to disallow anonymous access as would be appropriate if OIDC were enabled.

server/src/testflinger/api/v1.py
ajzobro: yeah -- ruff formatting won't allow the one extra char so I chopped it quickly since this was a TODO.  (It was keeping me from being able to commit due to pre-commit hooks that I have in place)

```suggestion
    # TODO: limit to job owner (or greater) if auth enabled so job owner known
```

server/src/testflinger/api/v1.py
ajzobro: Indeed, when removed the test to verify that no anonymous users can connect when OIDC is enabled does catch that, as well:

```
FAILED tests/test_v1_authorization_roles.py::TestAllEndPoints::test_permissions_with_oidc_enabled[None-GET:_v1_result_<job_id>:AGENT+CONTRIBUTOR+MANAGER+ADMIN] - 
AssertionError: Role contributor, Method GET, URI /v1/result/<job_id>: was expected to be unauthorized (401) not 200 OK b'{"job_sta...
```

server/src/testflinger/api/v1.py
ajzobro: See above, re: anonymous users + OIDC; When removed this also triggers test failures.

server/src/testflinger/api/v1.py
ajzobro:
```suggestion
        abort(HTTPStatus.UNAUTHORIZED, message="A login (known client_id) is required to securely store information.")
```

server/src/testflinger/api/v1.py
ajzobro:
```suggestion
        abort(HTTPStatus.UNAUTHORIZED, message="A login (known client_id) is required to securely store information.")
```

server/tests/test_results.py
ajzobro: I kept it working because we do have special code on the server to handle the large payload rejection as well.  If we keep that server code, I think we keep this test.

If we get rid of that server code -- well, we have other questions to ask like: what other endpoints might this apply to? Do we have any vulnerabilities to massive payloads?

server/tests/utilities.py
```diff
@@ -0,0 +1,81 @@
...
+    token_payload = {
```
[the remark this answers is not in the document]
ajzobro: not necessary.  Only affects testing; this logic exists in ~10 places

server/tests/utilities.py
ajzobro: mistake, TODO added to ensure that I reverted it -- I found it while you were reviewing and fixed it already.

server/src/testflinger/api/schemas.py
rene-oromtz: Thanks for the considering my suggestion! This also needs a schema refresh

server/tests/test_results.py
rene-oromtz: Honestly I don't know... That was the only endpoint I was aware of given the agent potentially sending a massive payload

rene-oromtz (approved):
LGTM! Thanks for updating the server to limit what users can or can't do with both OIDC enabled and disabled.

PR #1171, opened 2026-06-30
feat/update api docs generation

ajzobro:
# Auto-generate the server OpenAPI spec

**NOTE: Builds on PR #1169; not to be reviewed and merged until after.**

## Summary

This PR makes the single source of truth for API documentation into the OpenAPI spec, which is extracted from the app, committed as `server/schemas/openapi.json`.

## What changed

**Generation scripts (`server/scripts/`)**
- `generate_openapi_schema.py` — extracts the OpenAPI spec from the Flask app
  without needing MongoDB/OIDC, and injects `x-permission-roles` metadata from
  the `@require_role` decorators. Supports `--output` and `--diff`.

**Spec extraction harness**
- `server/devel/openapi_app.py` — minimal `TESTING` Flask app that registers all
  blueprints (including OIDC) so every endpoint appears in the spec.

**Committed artifacts**
- `server/schemas/openapi.json` — generated spec, single source of truth.

## Validation

```shell
just schema         # regenerate server/schemas/openapi.json
just check-schema   # verify the committed spec matches the code
```

- Generation is deterministic (`sort_keys`, stable ordering); re-running produces
  no diff.
- The generators are validated by their committed output plus the CI diff checks
  and manual review; dedicated unit tests are intentionally omitted at this time.


ajzobro:
@tang-mm I added this not knowing if we want to dispose of API.md altogether or continue to maintain it.

If we are going to let it go, we should delete it.

If we are going to maintain it, this proposal could help, but I would suggest we also move it to the documentation.

I am in favor of letting it go and focusing our efforts on the open api spec.

rene-oromtz:
@ajzobro whenever you have the chance can you please pull the latest changes from main to this branch?
I'm with you, we should probably remove `API.md` as its just a pain to maintain, hopefully the autogenerated OpenAPI schema and the `x-permission-roles` helps to map endpoints to roles clearly enough for the users

ajzobro:
@rene-oromtz updated; @copilot re-review

server/scripts/generate_openapi_schema.py
```diff
@@ -45,15 +45,69 @@
 from devel.openapi_app import app
 
 
+def _get_endpoint_roles(view_func):
+    """Extract @require_role roles from endpoint function wrapper.
+    
+    The decorator stores _role_requirements on the wrapper for introspection.
+    """
+    return getattr(view_func, '_role_requirements', [])
```
rene-oromtz, line 55: This is still missing the server side changes, currently the require_role does not store the _role_requirements on the wrapper

server/scripts/generate_openapi_schema.py
```diff
@@ -45,15 +45,69 @@
...
+        _inject_role_metadata(spec)
+        return spec
+
+
+def _inject_role_metadata(spec: dict):
+    """Add x-permission-roles field to endpoint operations based on @require_role decorators.
+    
+    This is a documentation-only extension, not used during request processing.
+    """
+    import re
```
rene-oromtz: Would it be possible to place this import at top level of the file?

rene-oromtz (changes requested):
Overall LGTM, thanks for this changes! I was playing around and there are two changes needed to display the roles in the OpenAPI reference docs
1. [Comment](https://github.com/canonical/testflinger/pull/1171/changes#r3667551160) regarding the require_role decorator
2. docs/reference/openapi.rst will also need a minor change to display this x-permission-roles. `showExtensions` is disabled by default as per Swagger [Docs](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/)

server/scripts/generate_openapi_schema.py
rene-oromtz: I think its just a one line change according to copilot e.g:
```
            return func(*args, **kwargs)
        wrapper._role_requirements = list(roles)
        return wrapper
```

server/scripts/generate_openapi_schema.py
ajzobro: Of course! This is what I get for letting AI help.

ajzobro: I'm surprised we don't have scripts to check/enforce good ordering of imports already.

rene-oromtz: We do but this path its currently not included:
https://github.com/canonical/testflinger/blob/104ae943440bb43c5975273a982c3fbfaca36b5c/server/pyproject.toml#L52-L58

Not that you need to in this PR, but if you want and have the chance, you can include this path to get all the benefits of `just server format`
e.g.
include = [
```
  "app.py",
  "src/**/*.py",
  "scripts/**/*.py",
  "tests/**/*.py",
]
```

rene-oromtz:
Thanks for the changes @ajzobro I see that the x-permission-roles are now in the schema. Can you please take a look at point 2?
> docs/reference/openapi.rst will also need a minor change to display this x-permission-roles. showExtensions is disabled by default as per Swagger [Docs](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/)

Specifically here:
https://github.com/canonical/testflinger/blob/104ae943440bb43c5975273a982c3fbfaca36b5c/docs/reference/openapi.rst?plain=1#L21-L28

After adding `showExtensions: true` the permissions will be visible in the docs as well

ajzobro:
> 1. [Comment](https://github.com/canonical/testflinger/pull/1171/changes#r3667551160) regarding the require_role decorator

See auth.py.

rene-oromtz:
@ajzobro yes, point 1 is already addressed in auth.py, pending comment is regarding point 2

> docs/reference/openapi.rst will also need a minor change to display this x-permission-roles. showExtensions is disabled by default as per Swagger [Docs](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/)

The line that needs to be added is in this [comment](https://github.com/canonical/testflinger/pull/1171#issuecomment-5109921388)

ajzobro:
Yes, understood. In progress and new docs are being inspected now. :-)

rene-oromtz (dismissed):
LGTM! Thanks for taking care of this

rene-oromtz (approved):

PR #1172, opened 2026-06-30
Add table of roles to reference documentation

ajzobro:
## Description

Adds a new reference page documenting the role-based access control (RBAC) permissions for every Testflinger server API endpoint.

**Changes**
 - docs/reference/api-roles.rst (new): Comprehensive reference table listing all API endpoints grouped by the roles permitted to call them. Covers four roles — AGENT, CONTRIBUTOR, MANAGER, and ADMIN — with a separate table for unauthenticated endpoints and a full permissions matrix for authenticated ones.
-  docs/.sphinx/_static/table-roles.css (new): Custom CSS to visually highlight permission markers (`:x:`) in the roles table with a distinct green badge style.
 - docs/conf.py: Registers the new CSS file and defines the `:x:` RST role, mapping it to the cell-x CSS class used for the permission badge rendering.
 - docs/reference/index.rst: Adds api-roles to the reference section toctree so the new page is included in the built docs.
 - docs/CONTRIBUTING.md: Adds a missing newline at end of file.

**Notes:**
The CONTRIBUTOR role note clarifies that when OIDC is disabled, CONTRIBUTOR-level endpoints also accept anonymous (unauthenticated) connections — a useful callout for operators evaluating their deployment's authentication posture.

## Resolved issues

N/A

## Documentation

This **is** documentation :-)

## Web service API changes

See #1169 for associated API changes.

## Tests

Visually inspected resulting HTML page.


docs/.sphinx/_static/table-roles.css
```diff
@@ -0,0 +1,10 @@
+/* Cell-x: highlight table cells that contain a permission marker */
+td .cell-x {
+    display: block;
+    background-color: #2d7d46;
```
ajzobro: I'm not sold on the color.

docs/reference/api-roles.rst
```diff
@@ -0,0 +1,302 @@
...
+   * - ``POST``
+     - ``/v1/oauth2/token``
+     - Exchanges client credentials (HTTP Basic) for a JWT access token
+       and a refresh token.  Cannot require a Bearer token because this
+       is how a client *obtains* one.
+   * - ``POST``
+     - ``/v1/oauth2/refresh``
+     - Exchanges a valid refresh token for a new access token.  No
+       Bearer token is required; the refresh token itself is the
+       credential.
```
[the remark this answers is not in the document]
ajzobro: @copilot be sure to compare this implementation against the the server code from PR #1169 and not the API from this branch itself.

docs/reference/api-roles.rst
```diff
@@ -0,0 +1,302 @@
...
+
+   * - Method
+     - URI
+     - AGENT
+     - CONTRIBUTOR
+     - MANAGER
+     - ADMIN
+   * - ``POST``
+     - ``/v1/agents/provision_logs/{agent_name}``
+     - :x:`x`
```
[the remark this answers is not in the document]
ajzobro: @copilot see above.  Compare this permission table against the roles as defined in PR #1169

docs/reference/api-roles.rst
```diff
@@ -0,0 +1,302 @@
...
+     -
+     - :x:`x`
+     - :x:`x`
+     - :x:`x`
+   * - ``DELETE``
+     - ``/v1/secrets/{client_id}/{path}``
+     -
+     - :x:`x`
+     - :x:`x`
+     - :x:`x`
```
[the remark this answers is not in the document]
ajzobro: @copilot see above. Compare this permission table against the roles as defined in PR #1169

docs/reference/api-roles.rst
```diff
@@ -0,0 +1,302 @@
...
+   * - ``PUT``
+     - ``/v1/secrets/{client_id}/{path}``
+     -
+     - :x:`x`
+     - :x:`x`
+     - :x:`x`
```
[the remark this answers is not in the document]
ajzobro: @copilot see above. Compare this permission table against the roles as defined in PR #1169

docs/.sphinx/_static/table-roles.css
```diff
@@ -0,0 +1,10 @@
+/* Cell-x: highlight table cells that contain a permission marker */
+td .cell-x {
```
tang-mm: You can use the icon set provided by the `Sphinx-design` extension (already enabled in the docs but very implicit). They support adding colored icons using RST syntax, such as 
``:octicon:`check-circle-fill;1em` `` and ``:octicon:`x-circle-fill;1em` `` 

https://sphinx-design.readthedocs.io/en/latest/badges_buttons.html#octicon-icons

docs/.sphinx/_static/table-roles.css
tang-mm: The sphinx-design extension provides default colors for green (sd-text-success) and red (sd-text-danger), but if we adhere to [Vanilla design](https://vanillaframework.io/docs/patterns/icons), the color for success is `#0e8420`

(Sadly, the vanilla icon classes cannot be used directly here)

docs/reference/api-roles.rst
```diff
@@ -0,0 +1,302 @@
...
+     - ADMIN
+   * - ``POST``
+     - ``/v1/agents/provision_logs/{agent_name}``
+     - :x:`x`
+     -
+     -
+     -
+   * - ``GET``
+     - ``/v1/job``
+     - :x:`x`
```
tang-mm: For accessibility considerations: 
This table uses "X" marks for "allowed", but it isn't obvious that the blank cells mean "denied", so they may be mistaken for the opposite meaning. Additionally, we cannot rely on colors as the only differentiator of key information. 

So my suggestion is:
If you keep a single symbol, switch "X" to a checkmark (the success icon). For the denied ones, add an explicit marker for "denied" such as a "/" or "-" so we know they are not inadvertently missed out

tang-mm (changes requested):
This is a great reference, but I'm concerned about maintainability. I see that #1171 is WIP to inject `x-permission-roles` metadata into the OpenAPI schema, which ensures the openapi.json is aligned with the @require_role implementation. Could we include the roles in the Swagger-rendered endpoints, and only keep the role definitions here? Alternatively, if the roles cannot be rendered, can we automate generating the role-endpoint table from the json file? This could be a step alongside the existing check-schema validation, so the table is produced from openapi.json and a CI diff check keeps it in sync.

(The automation part can be split into a follow-up PR. I also left some comments about the accessibility of the new symbol.)

docs/reference/api-roles.rst
ajzobro: Can do.

rene-oromtz:
I agree with @tang-mm I think on the long run maintaining the table is going to be hard. Perhaps we should prioritize #1171 now that the server endpoint restriction is live so we can validate how the OpenAPI schema generates the required roles and if we like it, perhaps use this PR for adding the reference to the roles?

ajzobro:
Thanks for the feedback, both of you.

I have changes for this one at 95%, I was just looking over some of the final formatting myself before submitting.  It will be here soon, probably before you look again.

But the points about auto-generating this from the code is a something I have a couple feelings about:

1) Not today, please.  I still think this is useful, and that we can transition to maintaining it from the openapi.json file. However, it started off as requirements. This was originally a capture of what **should** we do and drove the implementation.

Now that it's implemented, we can flip that around and it can show the implementation in a strategic way.

I would like to emphasize that where possible, I have grouped and clumped endpoints together such that all AGENT roles are together, etc. and that while the endpoints themselves might not be in a grouped order, the roles are presented as such here.

So, if we could accept this as accurate today then we can move to auto-generating it later.

docs/.sphinx/_static/table-roles.css
ajzobro: See latest code, I think it's better.

rene-oromtz (approved):
I like this approach better! I guess that if we want this approach, we later can revisit the automation part for generating this table. 

For now, LGTM but would want to see @tang-mm inputs on this changes

tang-mm:
@ajzobro Thanks for considering the previous feedback and adding the updates! The current version looks good to go :rocket: 

For later: I think an automated setup is still preferred to ensure long-term maintainability, but I get your point about grouping the roles together. When it happens, let's find a way to automate the table in the current format. Thanks for putting this together!

tang-mm (approved):

PR #1173, opened 2026-07-01
chore: Update justfiles to modular layout

pedro-avalos:
## Description

- Update `justfile`s to use a modular layout consistent with our other projects
- Use `just` in CI
- Add `.pre-commit-config.yaml` to be used with `prek` for a workspace/monorepo layout

## Documentation

- Update `CONTRIBUTING.md`

## Tests

Run `just` locally.


server/docker-compose.yml
```diff
@@ -33,6 +33,7 @@ services:
       - 5000:5000
     depends_on:
       - mongo
+      - dex
```
pedro-avalos, line 36: @rene-oromtz @ajzobro am I correct in assuming this should be required now that auth is required in testflinger?

rene-oromtz: Yes, I mean, its not a hard requirement as the server supports having both OIDC enabled and disabled but default is enabled

server/justfile
```diff
@@ -0,0 +1,91 @@
...
+
+[doc('Run all checks.')]
+[group('test')]
+check: lint format test check-schema
+    @echo 'Server checks done!'
+
+[arg('detach', long='detach', short='d', value='true', help='Run Testflinger in the background')]
+[doc('Run the server.')]
+[group('deploy')]
+serve detach='false':
```
rene-oromtz: Perhaps its just me, but I prefer the detach to default to true. I normally just build the server and continue in the same terminal. While you can detach afterwards, perhaps its more simple just to always run the detach?

agent/charms/testflinger-agent-host-charm/justfile
```diff
@@ -0,0 +1,78 @@
...
+[doc('Run integration tests.')]
+[group('test')]
+[positional-arguments]
+integration *args: charm
+    @uvx tox run -e integration -- "$@"
+
+[doc('Build the charm package')]
+[group('package')]
+charm:
+    @charmcraft pack
```
rene-oromtz: This is me probably nitpicking, but given this can also be called as: `just agent charm charm` if attempting to do it locally, perhaps a different name to reflect the action?
```suggestion
integration *args: pack
    @uvx tox run -e integration -- "$@"

[doc('Build the charm package')]
[group('package')]
pack:
    @charmcraft pack
```

I think `pack` or `package` could be appropiate

server/charm/justfile
```diff
@@ -0,0 +1,78 @@
...
+[doc('Run integration tests.')]
+[group('test')]
+[positional-arguments]
+integration *args: charm
+    @uvx tox run -e integration -- "$@"
+
+[doc('Build the charm package')]
+[group('package')]
+charm:
+    @charmcraft pack
```
rene-oromtz, line 78: Same comment as for the agent charm, perhaps this could benefit for a "pack" or "package" name for the recipe

rene-oromtz (changes requested):
Thank you for taking the time to add this to the repo! I really like how its structured. I only added some nitpick but besides that, LGTM and all tests are passing. 

Can you also please resolve the merge conflicts?

server/justfile
pedro-avalos: Yeah, I think you're right, it should detach by default also because if any other recipes want to deploy as a dependency, it should detach as well

rene-oromtz (approved):
LGTM! Thanks for adding the changes.

PR #1184, opened 2026-07-02
fix(agent-charm): reauthenticate on secret change

### LINKED ISSUE #1163: bug: updating tf-scripts in place leads to potential failures
Description
----
While updating a tf-script e.g ([tf-test](https://github.com/canonical/testflinger/blob/main/agent/charms/testflinger-agent-host-charm/src/tf-cmd-scripts/tf-test)) the modification is made in place leading to unexpected behavior if an agent is already during its respective phase (test phase for example). This in place modification is made by [update_charm_scripts](https://github.com/canonical/testflinger/blob/a277fff8b47d48dbe87b9dcb0a9a806531792110/agent/charms/testflinger-agent-host-charm/src/common.py#L66)

The file modification needs to be made atomically instead as in this case #1036 modified the file content and size and the agent that was processing the job read incorrectly the next instruction after completing the test cmds.

On next run, the agent will pick the new version so this is just a transient issue. This [analysis](https://superuser.com/a/1631617) details more appropiately on why we need to handle this atomically. 

Traceback
----

```
07:44:47 /usr/local/bin/tf-test: line 39: syntax error near unexpected token `cd'
07:44:47 /usr/local/bin/tf-test: line 39: `        bash -c "(cd /srv/testflinger-agent/$AGENT/testflinger/device-connectors && sudo pip install . &> /dev/null) && PYTHONUNBUFFERED=1 testflinger-device-connector $PROVISION_TYPE runtest -c /srv/testflinger-agent/$AGENT/default.yaml testflinger.json"'
```

rene-oromtz:
## Description
The authentication with the testflinger server needs to be updated whenever any potential reference to the Juju secret changes. This can be done whenever there is any "on_secret_changed" (the value of the secret was modified) or "config_changed" (potentially, the secret config option changed)

There is currently a bug where if `on_secret_changed` was fired, the authentication won't be triggered as the only validation being made is to check if the secret is already expired. Potentially, the secret is still within the 3 days allowed to live, the authentication won't be refreshed. 

Additionally, this fixes in place write to atomic writes so its safer for file write operations. Instead of writing in place this creates a temporary file which then replaces the original atomically. 

## Resolved issues
Resolves [CERTTF-1262](https://warthogs.atlassian.net/browse/CERTTF-1262)
Resolves #1163 

## Tests
Added unit tests for both fixes

[CERTTF-1262]: https://warthogs.atlassian.net/browse/CERTTF-1262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agent/charms/testflinger-agent-host-charm/src/charm.py
+++ b/agent/charms/testflinger-agent-host-charm/src/charm.py
@@ -244,7 +244,7 @@ class TestflingerAgentHostCharm(ops.charm.CharmBase):
 
         return content
 
-    def _update_refresh_token(self, event: ops.EventBase) -> bool:
+    def _update_refresh_token(self, event: ops.HookEvent) -> bool:
         """Update the refresh token if needed.
 
         This checks both if the refresh token is expired or if
--- a/agent/charms/testflinger-agent-host-charm/src/charm.py
+++ b/agent/charms/testflinger-agent-host-charm/src/charm.py
@@ -188,9 +188,9 @@ class TestflingerAgentHostCharm(ops.charm.CharmBase):
         self.update_testflinger_repo()
         self.unit.status = ops.ActiveStatus()
 
-    def on_start(self, _):
+    def on_start(self, event: ops.StartEvent):
         """Start the service."""
-        if not self._update_refresh_token():
+        if not self._update_refresh_token(event=event):
             return
         self.unit.status = ops.ActiveStatus()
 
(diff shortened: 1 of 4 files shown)


ajzobro (approved):
LGTM

PR #1190, opened 2026-07-06
Added new config arg - links

kdutch:
Added support for two new arguments in the config - links and control_host. Docs also updated to add new arguments, and new args added to mock config to validate post_initial_data works as intended.

## Description

This change was made to provide more information to the user via the CLI and to also allow for third party integrations. Support for a `links` section of the config has been added, and is a generic dict that can provide a variety of information about the agent, such including a link to the product on the manufacturers page or a link to the agent's cloud entry. The `agent_detail.html` template was also updated to ensure the addition of this field was reflected in the web UI. 

## Resolved issues

github.com/canonical/testflinger/issues/794

## Documentation

Docs were updated to include information about the new configuration arguments, and built to ensure they were working.

## Web service API changes
No new endpoints. Only change is that the `links` and `control_host` config arguments will be passed when the `agents/data/{agent_name}` endpoint is invoked. 

## Tests

All units were run before new arguments were added to the config, and after, to ensure that behavior was expected regardless of the presence of the optional arguments. I also had copilot use the altered `agent_detail.html` template and the example agent config to generate an example `agent_detail.html` based on the information in the config, to ensure the link section was displayed as intended.


agent/src/testflinger_agent/agent.py
```diff
@@ -132,6 +132,10 @@ def _post_initial_agent_data(self):
         }
         if identifier:
             agent_data["identifier"] = identifier
+        if links := self.client.config.get("links", None):
+            agent_data["links"] = links
+        if control_host := self.client.config.get("control_host", None):
+            agent_data["control_host"] = control_host
```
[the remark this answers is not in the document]
kdutch: Suggestion declined, as I think the walrus operator is cleaner. Please let me know if you disagree. Returning `None` is redundant but clear.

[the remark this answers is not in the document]
ajzobro: This is a good point; @kdutch what PR will have the server-side updates for the new data?  If it's WIP or planned, that's fine and we can ignore this issue for now, but if we know it we could like it here for clarity.

agent/src/testflinger_agent/schema.py
```diff
@@ -45,6 +45,8 @@
     voluptuous.Optional("output_timeout"): int,
     voluptuous.Optional("advertised_queues"): dict,
     voluptuous.Optional("advertised_images"): dict,
+    voluptuous.Optional("links"): dict,
+    voluptuous.Optional("control_host"): dict,
```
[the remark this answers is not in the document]
ajzobro: I also like the idea of tightening the specific schema as appropriate (not necessarily what copilot suggests)

agent/tests/test_agent.py
```diff
@@ -65,6 +65,8 @@ def agent(self, requests_mock):
                 "execution_basedir": self.tmpdir,
                 "logging_basedir": self.tmpdir,
                 "results_basedir": os.path.join(self.tmpdir, "results"),
+                "control_host": {"type": "bmc", "address": "127.0.0.1"},
+                "links": {"product": "fake.url_product.link"},
```
ajzobro: For example, we we are expecting the schema (http/https):// to be part of the url for the thing that will be labelled "product" for example, might be something we want to pull into this design.

docs/reference/testflinger-agent-conf.rst
```diff
@@ -54,6 +54,10 @@ The following configuration options are supported by the Testflinger Agent:
       - Command to run for the cleanup phase
     * - ``provision_type``
       - (Optional) type of device connector used. This is sometimes useful when templating the call to the external device-agent command, but is not required. For the list of supported provision types, see :doc:`device-connector-types`.
+    * - ``links``
+      - (Optional) Links to more information about the hardware in `label: info` format. Example: (`product: <Link to manufacturer hardware page>`)
+    * - ``control_host``
+      - (Optional) Management information for the hardware in `type: value` format. Example: (`bcm: <BCM address>`)
```
ajzobro: See copilot's response for this line; is it going to be 
```
{"type": "bmc", "address": "127.0.0.1"}
```
or
```
{"bmc": "127.0.0.1"}   # and if so, do we want to capture ports as well?
```

And while we're at it, is this supposed to just simply be a link like all the other links that we're able to collect and display?

agent/tests/test_agent.py
ajzobro: When we do have links, do we intend to present them in an order?  I was expecting a list of kv pairs.

If we want to control the order on the Agent side, we would need to preserve the config file order all the way through to the server; If we want the server to handle an unordered list and present them however it feels; we have that with this implementation.

ajzobro (changes requested):

agent/src/testflinger_agent/schema.py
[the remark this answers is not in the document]
kdutch: While `control_host` is intended to have a type and address fields, I felt it was better this convention was unenforced. That allows for alternate control schemes to be used if we keep it generic.

agent/src/testflinger_agent/agent.py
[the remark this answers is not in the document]
kdutch: The `server` currently takes any information provided by the agent on `post_initial_agent_data` and prodives it to anything that uses the data/agent/<agent_name> endpoint, meaning these changes are automatically propagated to the server. The server then supplies this information as is to the CLI, meaning this is propagated to the CLI as is. Or at least, it appears that way to me based on the code:

- The CLI makes a request to see agent data: https://github.com/canonical/testflinger/blob/6300157f05a7176b00f2603cbac2ecade65c6661/cli/testflinger_cli/client.py#L326
 - That accesses this endpoint in the server v1 API: https://github.com/canonical/testflinger/blob/6300157f05a7176b00f2603cbac2ecade65c6661/server/src/testflinger/api/v1.py#L684
 - Which is retrieved from the mongodb database: https://github.com/canonical/testflinger/blob/6300157f05a7176b00f2603cbac2ecade65c6661/server/src/testflinger/database.py#L419
 - Which is generated `post_initial_agent_data` based on: https://github.com/canonical/testflinger/blob/6300157f05a7176b00f2603cbac2ecade65c6661/agent/src/testflinger_agent/agent.py#L118
 - which is then posted to https://github.com/canonical/testflinger/blob/6300157f05a7176b00f2603cbac2ecade65c6661/server/src/testflinger/api/v1.py#L713 which generates/updates the entry in the database.
 
 As such, anything added to post_initial_data should be propagated to server and therefore the CLI.

docs/reference/testflinger-agent-conf.rst
kdutch: Yeah my bad the documentation needs to be updated to match. The BMC could be a link (ie. for our purposes a link to our service that has that info) but I wanted these to be generic and non-specific so it could have a wider variety of uses.

agent/tests/test_agent.py
kdutch: I wasn't intending the keys to be in any particular order - this is because some may have more links than others depending on the machine. For example, we might not have a product link for every machine in the fleet, or we might not have a resolvable cloud entry for every machine in the fleet. Some may have both or neither, and we might even want to do something like include a link to certified device information, i.e. this was intended to be generic. Anything using these links would have to account for retrieving the specific key they are attempting to interact with.

docs/reference/testflinger-agent-conf.rst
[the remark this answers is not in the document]
kdutch: I like the suggestion but I do think it's a bit to specific with control_host info - while this entry is intended to be in `type`/`address` format, it's not necessarily supposed to be restrictive, since there's schemas we may not have considered.

agent/src/testflinger_agent/agent.py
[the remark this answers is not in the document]
rene-oromtz: The server has a Marshmallow schema for both AgentIn and AgentOut: https://github.com/canonical/testflinger/blob/6300157f05a7176b00f2603cbac2ecade65c6661/server/src/testflinger/api/schemas.py#L50-L73

If the intention is for the CLI to return this information as well, you will also need to edit AgentOut to also return this new fields. If intention is just for server to be aware and later display it in the UI at least AgentIn is required. This is because the server rejects any unknown field e.g. this will be returned by server:
`{"detail": {"json": {"links": ["Unknown field."]}}, "message": "Validation error"}`

docs/reference/testflinger-agent-conf.rst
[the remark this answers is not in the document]
rene-oromtz: I do wonder if we should be restrictive on this (at least for the control hosts) given this will also require a server side schema validation. Having it restricted to "type" and "address" is more easy to model in Marshmallow, for example, you could create a ControlHost schema in server side that does something like:

```
type = fields.String(required=True)
address = fields.IPv4(required=True)
```

type could be either string to account for other potential "control hosts" but we could restrict address to be a valid IPv4. 

update: I just read this [comment](https://github.com/canonical/testflinger/pull/1190#discussion_r3531578327) so I guess perhaps we can drop the IPv4 type given this could be a link or include ports or we let the server "construct" the address based on other fields such as "protocol" and/or "port"

agent/src/testflinger_agent/agent.py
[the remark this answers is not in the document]
kdutch: Changes made.

server/src/testflinger/api/schemas.py
```diff
@@ -58,6 +58,7 @@ class AgentIn(Schema):
     queues = fields.List(fields.String(), required=False)
     state = fields.String(required=False)
     comment = fields.String(required=False)
+    links = fields.Dict(required=False)
```
ajzobro: We should check that the format of the Dict is str:str.

docs/reference/testflinger-agent-conf.rst
```diff
@@ -54,6 +54,8 @@ The following configuration options are supported by the Testflinger Agent:
       - Command to run for the cleanup phase
     * - ``provision_type``
       - (Optional) type of device connector used. This is sometimes useful when templating the call to the external device-agent command, but is not required. For the list of supported provision types, see :doc:`device-connector-types`.
+    * - ``links``
```
ajzobro, line 57: We should document the display order, test the display order, and remember to choose a good display order when configuring the agent configs (consistency).

The example should clarify the structure of the mapping, is an ordered list of key: value pairs.

server/src/testflinger/templates/agent_detail.html
```diff
@@ -55,6 +55,25 @@ <h2 class="p-muted-heading">Queues</h2>
...
+        <th>URL</th>
+      </tr>
+    </thead>
+    <tbody>
+      {% for label, url in agent.get('links', {}).items() %}
+        <tr class="searchable-row">
+          <td data-heading="Name">{{ label }}</td>
+          <td data-heading="URL">
+            <a href="{{ url }}">{{ url }}</a>
+          </td>
```
ajzobro, line 72: Add test for generated HTML and be sure to test link ordering and other tests for schema.

server/src/testflinger/api/schemas.py
kdutch: Changes made :+1:

server/src/testflinger/api/schemas.py
```diff
@@ -71,6 +72,7 @@ class AgentOut(Schema):
     job_id = fields.String(required=False)
     comment = fields.String(required=False)
     restricted_to = fields.Dict(required=False)
+    links = fields.Dict(required=False)
```
rene-oromtz: Server needs the same schema as the agent:

```suggestion
    links = fields.Dict(
        keys=fields.String(),
        values=fields.String(),
        required=False,
    )
```
I'm adding the suggestion but I don't know if ruff will complain

server/src/testflinger/templates/agent_detail.html
```diff
@@ -55,6 +55,25 @@ <h2 class="p-muted-heading">Queues</h2>
...
+  <h2 class="p-muted-heading">Links</h2>
+  <table aria-label="Links table" class="p-table--mobile-card">
+    <thead>
+      <tr>
+        <th>Name</th>
+        <th>URL</th>
+      </tr>
+    </thead>
+    <tbody>
+      {% for label, url in agent.get('links', {}).items() %}
```
rene-oromtz, line 67: I think this is all needed if we want the sorting to render any links alphanumerically:

```suggestion
      {% for label, url in agent.get('links', {}).items()|sort %}
```

Tested with:
```
curl -s -X POST \
  -H "Authorization: Bearer $TOKEN" \
  -d '{"identifier": "agent1", "links": {"c3": "http://test.com/agent1", "product": "http://test.com/agent1/product", "activity": "http://test.com/agent1/activity", "1link": "http://test.com/agent1/1link"}}' \
  http://localhost:5000/v1/agents/data/agent1
```

Its getting displayed as 1link, activity, c3, product

bladernr:
Given that this is part of a bigger design (and it's on me for not really stressing that before) can we revisit this in the context of the overall plan for this as a feature?

kdutch:
> Given that this is part of a bigger design (and it's on me for not really stressing that before) can we revisit this in the context of the overall plan for this as a feature?

Yes, I actually suggested that yesterday. I will decline this PR - we should look for a good first task :).

kdutch:
Conflict about implementation means work is postponed.

PR #1204, opened 2026-07-07
fix: do not allow adding client_secret to OIDC account

rene-oromtz:
## Description
Minor bug fix to not allow setting the client_secret to a OIDC account. 
Also changed from BAD_REQUEST to UNPROCESSABLE_ENTITY given the data is well structured and complete but is semantically wrong

## Resolved issues
Resolves [CERTTF-1294](https://warthogs.atlassian.net/browse/CERTTF-1294)

## Tests
Added unit test

[CERTTF-1294]: https://warthogs.atlassian.net/browse/CERTTF-1294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/src/testflinger/api/schemas.py
+++ b/server/src/testflinger/api/schemas.py
@@ -602,7 +602,7 @@ images_out = {
 class ClientPermissionsIn(Schema):
     """Client Permissions output schema."""
 
-    client_secret = fields.String(required=False)  # Optional for schema reuse
+    client_secret = fields.String(required=False, validate=Length(min=15))
     max_priority = fields.Dict(
         keys=fields.String(),
         values=fields.Integer(),
--- a/server/src/testflinger/api/v1.py
+++ b/server/src/testflinger/api/v1.py
@@ -1305,10 +1305,11 @@ def set_client_permissions(client_id: str, json_data: dict) -> str:
 
     client_secret = json_data.pop("client_secret", None)
     permissions = database.get_client_permissions(client_id) or {}
-    if "email" in json_data and permissions.get("sub"):
-        # Do not allow adding an email to an OIDC client id
+    if ("email" in json_data or client_secret) and permissions.get("sub"):
+        # Do not allow adding an email or client secret to an OIDC client id
         abort(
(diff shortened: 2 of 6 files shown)


server/src/testflinger/api/v1.py
```diff
@@ -1305,10 +1305,11 @@ def set_client_permissions(client_id: str, json_data: dict) -> str:
 
     client_secret = json_data.pop("client_secret", None)
     permissions = database.get_client_permissions(client_id) or {}
-    if "email" in json_data and permissions.get("sub"):
-        # Do not allow adding an email to an OIDC client id
+    if ("email" in json_data or client_secret) and permissions.get("sub"):
+        # Do not allow adding an email or client secret to an OIDC client id
```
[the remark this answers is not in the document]
rene-oromtz: a empty string should not be allowed in the first place. This should be handled by the schema

server/src/testflinger/api/schemas.py
```diff
@@ -602,7 +602,7 @@ class LogQueryParams(Schema):
 class ClientPermissionsIn(Schema):
     """Client Permissions output schema."""
 
-    client_secret = fields.String(required=False)  # Optional for schema reuse
+    client_secret = fields.String(required=False, validate=Length(min=15))
```
rene-oromtz, line 605: @ajzobro currently I'm just setting a client_secret length of 15 chars. Once we implement the self serve password change, we probably should implement a more strict validation e.g need to contain at least one number, one capitalized letter or special character. What do you think?

ajzobro: :-)  I have been going with 30
I do think that the mechanism in the future ought to create very long random keys instead of allowing them to be selected. Nobody is typing these -- and if they are: they are "doing it wrong".

ajzobro (dismissed):
This is fine for the moment.

ajzobro (approved):

PR #1205, opened 2026-07-07
fix: revert collMod given it requires dbAdmin permissions

rene-oromtz:
## Description
dbAdmin is required for collMod, we would need a different approach

## Tests
This also reverts the tests. All other should pass


ajzobro (approved):
LGTM, can try to add this back after we have a maintenance mode

PR #1253, opened 2026-07-17
Feat/add client id to web

ajzobro:
## Description

Add "Submitted By" to display the client_id on the web portal.

Added active "Job" to the Agent Detail page for the agent which has started running a given job:
image

Added "Submitted By" to the Agent list view:
image

Added the "Agent" which is currently running a job to the Job Details page:
image

Added "Submitted By" to the Jobs list view:
image

Added "Submitted By" to the Jobs list view of the Queue Detail page:
image

No changes to the Queue list page:
image

## Resolved issues

Added Client ID ("Submitted By") to the web portal in each place that the Job ID is current displayed.

## Documentation

Documentation not updated.

## Web service API changes

N/A - Client ID will be added to the API as well in a follow-on PR.

## Tests

See changes to `tests` for new tests.
Tested with new unit tests;
Tested with local workshop:

```
just workshop serve && workshop exec -- just server::populate
...
Created dev admin credential
Created OIDC admin credential 'testflinger-admin'
Created sample client 'alice@example.com' (email: alice@example.com)
Created sample client 'ci-bot-kernel' (email: alice@example.com)
Created sample client 'ci-bot-snapd' (email: bob@example.com)
Created sample client 'qa-runner-x86' (email: carol@example.com)
Created sample client 'infra-agent-arm' (email: dave@example.com)
INFO:root:Created 10 queues
INFO:root:Created 10 agents
INFO:root:Created 10 jobs
```


agent/src/testflinger_agent/agent.py
```diff
@@ -323,8 +323,6 @@ def process_jobs(self):
                 )
                 os.makedirs(rundir)
 
-                self.client.post_agent_data({"job_id": job.job_id})
```
ajzobro, line 326: This is now happening (the agent is immediately associated with the job) when the agent asks the server for more work.

If there was work to give to the agent, the server make that association at the time of job issuance.

server/src/testflinger/api/v1.py
```diff
@@ -238,6 +243,7 @@ def job_get():
         return jsonify({}), HTTPStatus.NO_CONTENT
     if (secrets := retrieve_secrets(job)) is not None:
         job["test_data"]["secrets"] = secrets
+    database.set_agent_job(agent_name, job["job_id"])
```
ajzobro, line 266: This is where the agent <-> job relationship is now set.

ajzobro: A long time ago (before the aget_name existed in the cookie and agents were required to authenticate) this couldn't be done.

server/devel/dex-config.yaml
```diff
@@ -18,7 +18,27 @@ staticPasswords:
   - email: "testflinger@example.com"
     hash: "$2a$10$FWI.y.8Zpa9PhTTcV6GDIuxPzrwYJoiVCimgCA427HoyR5FR50mBu"
     username: "testflinger-admin"
-    userID: "1234"
+    userID: "1001"
+  - email: "dev-admin@example.com"
+    hash: "$2a$10$FWI.y.8Zpa9PhTTcV6GDIuxPzrwYJoiVCimgCA427HoyR5FR50mBu"
```
ajzobro: This hash is for `testflinger`, but the dev-admin account is the only account that is given `dev-secret-for-testing` which this hash does not match.

server/src/testflinger/templates/_agents_table.html
```diff
@@ -76,6 +79,7 @@
           <td class="u-align--right u-truncate" data-heading="Job ID">
             <a href="{{ url_for('testflinger.job_detail', job_id=agent.job_id) }}">{{ agent.job_id }}</a>
           </td>
+          <td data-heading="">{{ agent.client_id or "" }}</td>
```
ajzobro: I'm changing this, not sold on how the agent is imbued with client data -- as it's really the agent's current job (if and only if there is a job) whose client should be shown.

server/src/testflinger/database.py
```diff
@@ -416,6 +421,18 @@ def get_agent_info(agent: str) -> dict:
...
 
 
+def set_agent_job(agent_name: str, job_id: str) -> None:
+    """Record the job an agent is currently running.
+
+    :param agent_name: Name of the agent.
+    :param job_id: ID of the job the agent has been assigned.
+    """
+    mongo.db.agents.update_one(
+        {"name": agent_name},
```
ajzobro, line 457: i like that we gather the job, but I don't think the agent should errantly be associated with the client_id

server/src/testflinger/views.py
```diff
@@ -57,6 +57,7 @@ def home():
 def agents():
     """Agents view."""
     agent_info = mongo.db.agents.find()
+    agent_info = enrich_agents_with_client_id(agent_info)
```
ajzobro: I rather think that the agent info is being "enriched" with the job information.

agent -> job -> client_id

server/devel/create_sample_users.py
```diff
@@ -0,0 +1,99 @@
...
+import os
+
+import bcrypt
+from pymongo import MongoClient
+
+from testflinger.database import get_mongo_uri
+from sample_users import SAMPLE_CLIENTS
+
+DEV_CLIENT_ID = "dev-admin"
+DEV_CLIENT_KEY = "dev-secret-for-testing"
```
ajzobro: Should we just change this to `testflinger` like the rest?

server/src/testflinger/database.py
```diff
@@ -136,6 +136,7 @@ def create_indexes():
     mongo.db.client_permissions.create_index("sub", sparse=True)
     mongo.db.jobs.create_index("job_id")
     mongo.db.jobs.create_index(["result_data.job_state", "job_data.job_queue"])
+    mongo.db.agents.create_index("queues")
```
ajzobro, line 139: Performance measure: indexes where used, see below line 323 and also line 463

server/src/testflinger/views.py
```diff
@@ -348,3 +371,33 @@ def seconds_to_hms(seconds: float) -> str:
...
+            agent.setdefault("active_job", None)
+        return agents
+    job_data_map = {
+        doc["job_id"]: doc
+        for doc in mongo.db.jobs.find(
+            {"job_id": {"$in": job_ids}},
+            {"job_id": 1, "client_id": 1, "_id": 0},
+        )
+    }
+    for agent in agents:
```
ajzobro: Optionally (avoids double return paths):

```suggestion
    if job_ids:
        job_data_map = {
            doc["job_id"]: doc
            for doc in mongo.db.jobs.find(
                {"job_id": {"$in": job_ids}},
                {"job_id": 1, "client_id": 1, "_id": 0},
            )
        }
    else:
        job_data_map = {}
    for agent in agents:
```

server/tests/test_v1.py
```diff
@@ -735,6 +735,37 @@ def test_resubmit_job_state(mongo_app):
     assert "waiting" == updated_data.get("job_state")
 
 
+def test_job_post_stores_client_id(mongo_app):
+    """Test that submitting a job stores client_id at top level.
+
+    When OIDC is not enabled, the client_id is None (anonymous), but
+    the key should still be present on the job document.
```
ajzobro: This is not always true, client_id *may* be specified when OIDC is not enabled!
```suggestion
    When OIDC is not enabled, the client_id can be None (anonymous), but
    the key should still be present on the job document.
```

server/tests/test_v1.py
```diff
@@ -1771,6 +1802,62 @@ def test_pop_job_respects_exclude_agents(mongo_app, agent_auth_header):
...
+
+    agent_name = "agent1"
+    app.post(
+        f"/v1/agents/data/{agent_name}",
+        json={"state": "waiting", "queues": ["test"], "location": "here"},
+        headers=agent_auth_header,
+    )
+
+    # Confirm no job_id on agent before dispatch
+    agent_record = mongo.agents.find_one({"name": agent_name})
```
ajzobro, line 1846: We should add tests to ensure that this gets properly cleared out when jobs complete/agents restart/etc.

server/devel/dex-config.yaml
rene-oromtz: If you prefer to keep dev-admin with testflinger credentials its okay, only reason I selected dev-secret-for-testing was to not get a warning about password length

server/devel/create_sample_users.py
rene-oromtz: Same as this [comment](https://github.com/canonical/testflinger/pull/1253/changes#r3826331484)

If you prefer to keep testflinger, just note there are couple of places that will required to be modified.

rene-oromtz: But also, I see  that you are also creating in MongoDB the "testflinger-admin" account so if prefer I could also see this going away. No need to keep two "admin" accounts.

server/src/testflinger/api/v1.py
```diff
@@ -822,6 +838,14 @@ def agents_provision_logs_post(agent_name, json_data):
     timestamp = datetime.now(timezone.utc)
     agent_record["updated_at"] = json_data["timestamp"] = timestamp
 
+    # Look up the client_id from the job so it can be displayed in the
+    # provision history table.
+    job = database.mongo.db.jobs.find_one(
+        {"job_id": json_data["job_id"]},
+        {"client_id": 1, "_id": 0},
+    )
+    json_data["client_id"] = job.get("client_id") if job else None
```
rene-oromtz: can we move this logic to the database module just to keep the consistency in this PR? (also would love to eventually see all DB operations be removed from this file)

server/src/testflinger/database.py
rene-oromtz: Thanks! I agree think this will help a lot to the queues page (I hope).

agent/src/testflinger_agent/agent.py
rene-oromtz: I think this can also be removed given this is cleared as well server side

rene-oromtz: Just note that for this to work "smoothly" we need to update server first. Then it doesn't matter when we update the agents

server/src/testflinger/api/v1.py
```diff
@@ -582,7 +588,17 @@ def result_post(job_id: str, json_data: dict) -> str:
...
+    # so the agent no longer appears to be running a completed/cancelled job.
+    # Read job_state before add_job_results mutates json_data in-place.
+    terminal_states = {"complete", "completed", "cancelled"}
+    job_state = json_data.get("job_state")
+
     database.add_job_results(job_id, json_data)
+
+    if job_state in terminal_states:
+        database.clear_agent_job(job_id)
+
```
rene-oromtz: I think its worth to check but perhaps its not as safe to clear the job_id here as cancellations follows a different status as this status is not POST'ed by the agent e.g. [cancel_job](https://github.com/canonical/testflinger/pull/1253/changes#diff-ebfd6678908c7fb45dd484c12c0ab5bd8fedcca2f6093ccc75952503b003a87eL997)

cancelled status is written directly to the DB and agent reads this for determining it should stop processing but it can take a while for the cancellation to kick in. Given this, clearing it here for "cancelled" will have no effect and we still need to clear it agent side anyway (current behavior)

agent/src/testflinger_agent/agent.py
rene-oromtz: Nvm, I just added a comment for this: https://github.com/canonical/testflinger/pull/1253/changes#r3826630670

Given cancellation follows a different pattern, agent will still need to clear the job_id.

server/src/testflinger/database.py
```diff
@@ -291,6 +297,26 @@ def get_queue_wait_times(queues: list[str] | None = None) -> list[dict]:
...
+                "state": 1,
+                "location": 1,
+                "updated_at": 1,
+                "job_id": 1,
+                "comment": 1,
+                "provision_streak_type": 1,
+                "provision_streak_count": 1,
+            },
+        )
+    )
```
rene-oromtz: I think this can be dropped as this already exists in [get_agents](https://github.com/canonical/testflinger/pull/1253/changes#diff-36d910d54d1305414a4e92e63df96a1c561cf9136e3c1b1aa344934e1cc6d781L447)

Same results, different way to get it (whitelist vs blacklist)

server/src/testflinger/database.py
rene-oromtz, line 508: As mentioned in this [comment](https://github.com/canonical/testflinger/pull/1253/changes#r3826681765) this was duplicated by agent_get_all. 

I just wanted to add a comment that its up to you if decide to implement in this PR or when we tackle the API change. Given the views page is attaching job_id and client_id to the agents page I think we can tackle all of this on one single MongoDB query via [aggregate](https://www.mongodb.com/docs/manual/aggregation/)

That way `get_agents` can be used for:
1. agent information
2. attach job information that the agent is currently processing

All done via single query and both views and API benefit from

server/src/testflinger/views.py
```diff
@@ -348,3 +371,32 @@ def seconds_to_hms(seconds: float) -> str:
...
+            for doc in mongo.db.jobs.find(
+                {"job_id": {"$in": job_ids}},
+                {"job_id": 1, "client_id": 1, "_id": 0},
+            )
+        }
+    else:
+        job_data_map = {}
+    for agent in agents:
+        agent.setdefault("active_job", job_data_map.get(agent.get("job_id")))
+    return agents
```
rene-oromtz, line 402: This is the portion I mentioned we could benefit from using MongoDb aggregate, this could either be removed in this PR or in the followup PR for API change. 

Given the latter is happening anyway, I'll let you choose when we should tackle this or if further discussion is needed.

rene-oromtz (changes requested):
Thanks for taking care of this! This will be a really nice addition to the UI! I added some comments, feel free to let me know if we should discuss any of those further

server/src/testflinger/api/v1.py
ajzobro: My preference is for the server to manage the relationships between jobs and agents as that information is known.

I don' know what you mean by "safe", the `clear_agent_job` function is new to this PR and simply clears the "agent -> job_id" relation ("what is the active job for this agent?"); the "job -> agent" relation remains (always will, it's "where the job was run"

Cancelled status can (today) come from the CLI (and later, perhaps the Web UI) as well as the "multi/tfclient" device connector.

The cancellation goes to the server; gets picked up by the agent's `JobCancelledChecker` and `runner.py:143` and 151 both effectively make calls *after* the server has noted the job as cancelled to call `post_output` which in turn calls `live_output_handler` and would end up here at this server endpoint.

That said, the moment that the job is cancelled is ALSO a good time for the job to be removed from the agent. We can take that to its natural end and say that "When a job status changes to any of the terminal states, the agent->job association should be set to null" -- this would remove the complicated loop and frankly is a better approach, IMO.

I will move this logic accordingly.

server/src/testflinger/api/v1.py
ajzobro: yes

agent/src/testflinger_agent/agent.py
ajzobro: I think that you're right though, and that the job should be disassociated with the agent record when the job status reaches a terminal state.

server/devel/create_sample_users.py
ajzobro: I will remove this variant.

server/src/testflinger/database.py
ajzobro: Yes, they are close enough, there's 3 fields different and some things that get added that won't be on the web ui, but neither get the logs, so that's fine.

Removed.

server/src/testflinger/views.py
ajzobro: Postponing for next PR, hot on the tails of this PR, notes captured in that worktree already.

server/src/testflinger/api/v1.py
ajzobro: moved the whole function since it was all DB work

server/src/testflinger/database.py
```diff
@@ -694,6 +703,12 @@ def add_job_results(job_id: str, json_data: dict):
 
     mongo.db.jobs.update_one({"job_id": job_id}, {"$set": json_data})
 
+    # Additionally, because the job_data may reflect that the job is now done,
+    # we need to disassociate the agent from the job if the job is done:
+    terminal_states = {"complete", "completed", "cancelled"}
```
rene-oromtz, line 705: Given cancelled can't be obtained via POST, this could be simplified:

```suggestion
    terminal_states = {"complete", "completed"}
```

server/src/testflinger/database.py
```diff
@@ -267,6 +267,34 @@ def pop_job(queue_list: list[str], agent_name: str) -> dict | None:
...
+            "result_data.job_state": {
+                "$nin": ["cancelled", "complete", "completed"]
+            },
+        },
+        {"$set": {"result_data.job_state": "cancelled"}},
+    )
+    modifications += response.modified_count
+
+    # Furthermore disassociate from the agent:
+    modifications += clear_agent_job(job_id)
```
rene-oromtz, line 290: My previous concern in this [comment](https://github.com/canonical/testflinger/pull/1253#discussion_r3826630670) is that if we clear the id immediately after cancellation, we will have an inconsistent state between agent and its state. 

Taking MAAS as an example:
1. User cancel job
2. Agent gets the cancellation but if its during provision it will defer it until it timeout
3. The current behavior will clear the job_id but the agent will still be "provisioning" without any job visible in the UI which can lead to confusion. I think the old behavior was more friendly as at least you could see the attached job is being "cancelled"


I just realized that the agent already handles this cancellation scenario without needing to clear it here. After the job gets cancelled, eventually the agent will send state "complete" so one of your current changes will already handle this cleanly: 
```
    if json_data.get("result_data.job_state") in terminal_states:
        clear_agent_job(job_id)
```

Still done server side but actually cleared whenever the agent stop processing the job

server/src/testflinger/database.py
```diff
@@ -267,6 +267,34 @@ def pop_job(queue_list: list[str], agent_name: str) -> dict | None:
...
+
+    if not modifications:
+        return "The job is already completed or cancelled", 400
```
rene-oromtz: Nitpick:

```suggestion
        return "The job is already completed or cancelled", HTTPStatus.BAD_REQUEST
```

rene-oromtz (changes requested):
Just added some final comments but overall LGTM!

server/src/testflinger/database.py
ajzobro: Oh.  I view this differently: The job has been cancelled.

The fact that the agent refuses to stop doing agent things doesn't change the fact that the job has been cancelled and that the agent has been told to cease working on it.

Agents have to clean-up their DUT, but that's not a job thing anymore -- that's an agent keeping it's DUT in a good state.

I view the job cancellation as a "ripping the rug out from under it's feet" sort of action:  Cancelled is a terminal state; the agent shouldn't even be allowed to write to that job record once it's been told to stop work.

In short: The agent's association with that job should be terminated when it is told to stop work on the job.  The job has reached termination ("cancelled").

-------

From your telling, the job is still relevant, the agent continues to do things on behalf of that job, despite the job being cancelled -- that's a poor reflection of the original job's status: cancelled.

If an agent were to ignore the server and continue running with the original parameters of that job, that is the agent acting out of turn and disobeying the server's direction to stop.

Again: the reason that the agent does so is not _for the job_ but rather _for the DUT_ to be left in a good state.

server/src/testflinger/database.py
ajzobro: line longer than 79 char.

frankly this is an HTTP response and I shouldn't have brought it over into the database module.

server/src/testflinger/database.py
ajzobro: I think that it is better database code to handle all the terminal states rather than reducing the checks to only apply to the current external callers.

However, you and I need to figure out if we consider "cancelled" to be a terminal state or a directive to the agent.

If "cancelled" is a state indicating pending resolution with the agent, then it is not a terminal state.

If it is an indication that that job has reached then end of the line then it is both a terminal state AND should not allow further data to be written to that job (by anyone, agent, server, etc.)

ajzobro (commented):
Needs some thought, let's talk through these remaining issues.

rene-oromtz (approved):
Thanks for addressing the changes. LGTM!

server/src/testflinger/database.py
rene-oromtz: As agreed on our 1:1, cancellations needs to properly be handled by the agent to display more accurately what the agent is doing e.g. when job is cancelled during provision, agent is no longer processing the job but agent needs to make server aware that is still "cleaning up"

PR #1254, opened 2026-07-17
docs: add OIDC how-to, reference and explanation

rene-oromtz:
## Description
This add all the information required from an admin perspective on OIDC related functionality:
- How-to: Enable OIDC
- Explanation: on what Testflinger requires from the OIDC provider to properly work
- Reference: on the charm configuration options

## Resolved issues
Resolves [CERTTF-1152](https://warthogs.atlassian.net/browse/CERTTF-1152)

## Documentation
Added two new pages for OIDC related content
This is somewhat related to #1161 but focuses on OIDC from a more administrative perspective. 

## Tests
Built the docs locally

[CERTTF-1152]: https://warthogs.atlassian.net/browse/CERTTF-1152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/how-to/administer-server/maintain-testflinger-server.rst
```diff
@@ -54,3 +54,32 @@ secrets so that they can be accessed by agents at job runtime.
...
   the following command: ``openssl rand -base64 96 | tr -d '\n'``
+
+.. _howto-enable-oidc:
+
+Enable OIDC
+-----------
+
+To enable OpenID Connect (OIDC) authentication for Testflinger, you will need to
+configure the Testflinger server with the required parameters for your OIDC provider.
+
```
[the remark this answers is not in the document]
rene-oromtz: No contradiction here, the table already includes which are the required parameters

docs/how-to/administer-server/maintain-testflinger-server.rst
```diff
@@ -54,3 +54,30 @@ secrets so that they can be accessed by agents at job runtime.
   The ``testflinger_secrets_master_key`` is used to encrypt the CSFLE data keys
   stored in MongoDB. Generate a secure random base64-encoded string by running
   the following command: ``openssl rand -base64 96 | tr -d '\n'``
+
+.. _howto-enable-oidc:
+
+Enable OIDC
+-----------
+
+To enable OpenID Connect (OIDC) authentication for Testflinger, you will need to
```
tang-mm, line 69: Since OIDC credentials are required parameters, I'd suggest clarifying the prerequisites first. Something like:

```
Before you start, ensure that you have registered Testflinger as an application
with your OIDC provider, and obtained the client ID, client secret, and issuer 
URL from the provider.
```

docs/how-to/administer-server/maintain-testflinger-server.rst
```diff
@@ -54,3 +54,30 @@ secrets so that they can be accessed by agents at job runtime.
...
+      oidc_provider_issuer="<issuer-url>"
+
+Additionally, you will need to set up Flask's ``SECRET_KEY`` to handle web session
+management by signing session cookies. To configure it, run the following command:
+
+.. code-block:: shell
+
+  $ juju config testflinger-k8s \
+      web_secret_key="<secret-key>"
+
```
tang-mm, line 86: It will be helpful to add a verification step after running the two config changes. Can we get the activation status through `juju status` or logs?

docs/explanation/oidc-auth.rst
```diff
@@ -0,0 +1,68 @@
...
+Testflinger supports OpenID Connect (OIDC) authentication for simplifying the
+management of user logins. This enables integration with existing enterprise
+identity systems (such as SSO providers), centralizes access control, and
+allows administrators to manage user lifecycle through the identity provider
+rather than the Testflinger server itself.
+
+When OIDC is enabled, Testflinger restricts access to its resources to
+authenticated users only. Users accessing the web interface or the CLI must
+authenticate before they can use any functionality. When OIDC is not configured,
+the server allows anonymous access.
```
tang-mm: I think more clarification is needed on the two modes, if I understand correctly:

When OIDC is not enabled, server allows anonymous access on uncontrolled resources, but also requires authentication through client ID for admin permissions.
When OIDC is enabled, basic auth access is disabled, and thus all access must go through OIDC authentication.

docs/reference/juju-oidc-config.rst
```diff
@@ -0,0 +1,45 @@
...
+     - Yes
+     - The client ID issued by the OIDC provider when registering Testflinger
+       as an application.
+   * - ``oidc_provider_issuer``
+     - Yes
+     - The issuer URL of the OIDC provider. Testflinger uses this to discover
+       provider metadata via ``<issuer>/.well-known/openid-configuration``.
+   * - ``web_secret_key``
+     - Yes
+     - A secret key used by Flask to sign web session cookies. Must be a long
```
tang-mm, line 32: Is there a max/min requirement on the length?

tang-mm (commented):
Great structure and content! The diataxis separation works well here. I added a few suggestions on clarity

docs/how-to/administer-server/maintain-testflinger-server.rst
ajzobro: You can run `juju config testfinger-k8s | grep 'web_secret_key' - A5` and see the value... but that's about it.

docs/explanation/oidc-auth.rst
rene-oromtz: Correct, I'll add this clarification

1. OIDC disabled: all non restricted resources e.g. (non admin, non-agent) are accessible, otherwise will require credentials
2. OIDC enabled: Users need to have OIDC authentication or client credentials. In our setup, client credentials are meant generally for CI usage but I think I can just leave it as `either OIDC or client credentials"

I think I can also ref to the table of roles Andy added

docs/reference/juju-oidc-config.rst
rene-oromtz: Not on Flask's side nor our side. I took that definition from Flask configuration option and the only thing they say is to be "long"...

docs/how-to/administer-server/maintain-testflinger-server.rst
rene-oromtz: `juju config testfinger-k8s web_secret_key` will also do the trick but yeah, that's the only verification that we have.

But this comment make me realize we also need to do a `juju status` as the best way to tell this changes worked is if the unit is not `BLOCKED` otherwise something is missing on the setup

docs/explanation/oidc-auth.rst
rene-oromtz: Added in [0bb3aff](https://github.com/canonical/testflinger/pull/1254/commits/0bb3affb334960a5d5e36eb564c87014296c88df), please let me know if any modification is needed to the phrasing I used

docs/how-to/administer-server/maintain-testflinger-server.rst
rene-oromtz: Added both clarifications in [0bb3aff](https://github.com/canonical/testflinger/pull/1254/commits/0bb3affb334960a5d5e36eb564c87014296c88df). Let me know what you both think!

docs/reference/juju-oidc-config.rst
ajzobro: This is a general security question and not something that our application mandates or dictates.

*If* we add checks for this, I propose to make them very long indeed, and since it's a configuration setting it would be probably best to check that at the Charm level unless we want our app to take the role of proving that it is configured well...

docs/explanation/oidc-auth.rst
```diff
@@ -0,0 +1,68 @@
...
+
+.. seealso::
+
+   :ref:`howto-enable-oidc`
```
ajzobro:
```suggestion
   How to :ref:`howto-enable-oidc`
```

docs/explanation/oidc-auth.rst
```diff
@@ -0,0 +1,68 @@
...
+      Steps to configure a provider and enable OIDC on a Testflinger server.
+
+   :doc:`../reference/oidc-config`
```
ajzobro:
```suggestion
   Reference: :doc:`../reference/oidc-config`
```

ajzobro: this one feels more contrived than the `How to` example; but the How to example only read "Enable OIDC" rather than "How To Enable OIDC"

docs/explanation/oidc-auth.rst
```diff
@@ -0,0 +1,68 @@
...
+
+Provider Capabilities
+---------------------
+
+Testflinger relies on specific OIDC capabilities that the configured provider
+must support:
+
+- **Token authentication**: Testflinger uses ``client_secret_basic`` (HTTP Basic
+  Auth in the Authorization header) for confidential clients, or falls back to
+  including the ``client_id`` in the request body for public clients.
```
ajzobro: > confidential clients, or falls back to including the ``client_id`` in the request body for public clients.

Oh! **Testflinger Server** is the client in these contexts, and the client_id in question is that of the testflinger server as a client to the OIDC service.

Basic Auth is then comprised not only of `oidc_client_secret` but also `oidc_client_id`

Rather than using `client_id` and `client_secret` (the testflinger input terms), use the following  consistently when talking about the testflinger server in its role as a client to the OIDC service:
 - `oidc_client_id`
 - `oidc_client_secret`

docs/explanation/oidc-auth.rst
```diff
@@ -0,0 +1,68 @@
...
+- **Authorization Code Flow**: required for web interface authentication, using
+  the ``authorization_code`` grant type.
+- **Device Authorization Flow**: required for CLI authentication, using the
+  ``urn:ietf:params:oauth:grant-type:device_code`` grant type.
+- **Scopes**: Testflinger requests ``openid``, ``profile``, and ``email`` scopes.
+  The ``email`` claim is used as the user's stable identity within Testflinger.
+- **Redirect URI**: the provider must specify a redirect URI with the following format
```
ajzobro:
```suggestion
- **Redirect URI**: the provider must suppport a redirect URI with the following format
```

docs/how-to/administer-server/maintain-testflinger-server.rst
```diff
@@ -54,3 +54,30 @@ secrets so that they can be accessed by agents at job runtime.
...
+configure the Testflinger server with the required parameters for your OIDC provider.
+
+.. code-block:: shell
+
+  $ juju config testflinger-k8s \
+      oidc_client_id="<client-id>" \
+      oidc_client_secret="<client-secret>" \
+      oidc_provider_issuer="<issuer-url>"
+
+Additionally, you will need to set up Flask's ``SECRET_KEY`` to handle web session
```
ajzobro, line 79: Since this portion must be done any time the oidc_* parameters are set (above):

- a) why don't we just list it above with the rest?
- b) if this token is not in-and-of-itself OIDC-related, but generally just needed all the time because we are trusting the cookies that flask is generating for the purpose of agent cookies -- then we should probably not allow this to be optional.

Thoughts?

docs/reference/juju-oidc-config.rst
```diff
@@ -0,0 +1,45 @@
+.. _ref-oidc-configuration:
+
+OIDC Configuration
+==================
+
+The following configuration options control OIDC authentication for a Testflinger
+server deployed via Juju charm.
```
ajzobro, line 7: Are these values still usable as env vars if not in a charm environment?

docs/reference/juju-oidc-config.rst
```diff
@@ -0,0 +1,45 @@
...
+
+.. note::
+
+    OIDC is only enabled when all required parameters are set.
+    If any is missing, the server will remain in ``BLOCKED`` status
```
ajzobro, line 12: what does a "BLOCKED" status look and feel like to the user?  it that just a charm thing?

I.e. if a user were to set up the server with environment variables and not use Juju?

So, if we are going to consider the server as a "charm" and nothing more, we should be clear what parts of the documentation are charm-specific and which are "server"-specific.

docs/reference/juju-oidc-config.rst
```diff
@@ -0,0 +1,45 @@
...
+   * - ``web_secret_key``
+     - Yes
+     - A secret key used by Flask to sign web session cookies. Must be a long
+       random string. Generate one with:
+       ``python -c 'import secrets; print(secrets.token_hex())'`` or with
+       any other secure random string generator.
+   * - ``oidc_client_secret``
+     - No
+     - The client secret issued by the OIDC provider. Required for confidential
+       clients. Omit for public clients, in which case Testflinger falls back to
```
ajzobro, line 39: Tell me more about "public clients"

Or tell me less about it -- but right now I have to look and say "what is a public client?"

docs/reference/juju-oidc-config.rst
ajzobro: `oidc-config.rst` -> `juju-config-oidc.rst` ?

docs/reference/juju-oidc-config.rst
rene-oromtz: Correct, if we choose to enforce a very long value, it must be done at Charm level with the Pydantic validation that is also used for other config options. @ajzobro I can quickly add this change in a spinoff PR if you want, should be very easy but perhaps doesn't need to be tied to this PR, what do you think?

docs/how-to/administer-server/maintain-testflinger-server.rst
rene-oromtz: Its no OIDC related but there is no reason why to set this parameter without auth enabled as otherwise there is no user session in the UI. 

That's why I choose to separate them as
A) you need to set the required OIDC parameters
B) But you also need this Flask variable because its needed to sign the session cookies. 

For that same reason I think they should be kept as-is, separate of the OIDC options but used together (server enforces this)

docs/reference/juju-oidc-config.rst
rene-oromtz: Yes, those can also be set as environment variables but I think this setup makes more sense for testing. For that reason, I'll take the other comment and also rename the file as its Juju related

docs/reference/juju-oidc-config.rst
rene-oromtz: Its a Charm thing, while the server can be launched without a charm deployment, this feels more testing-like environment instead of production which this documentation should target. Given this part of the documentation already specifies this are config options for Juju charm, do you think we should add anything else to this section?

docs/reference/juju-oidc-config.rst
rene-oromtz: Its a OIDC client that can't store the client secret, not advising to use them as those are less secure but its technically a possibility. Its the same config we are using on Dex for testing because for some reason, the client_secret_basic failed on dev environment

docs/explanation/oidc-auth.rst
rene-oromtz: I added a couple the clarification as "OIDC `client_id`" as the `oidc_client_id` feels more like a Charm term. The docs is also already differentiating between this and "Testflinger `client_id`"

docs/reference/juju-oidc-config.rst
ajzobro: I agree that it is a spin-off.
Do I want to enforce good security practices (in house? Yes) in general, yes and no: Just because we decide that a good secret key length is 30 or 120 doesn't mean that we should force everyone to use that minimum length -- especially since it is going to be a matter of one's individual security practice.

If I have to pick: err on the side of caution and add the checks, ask security for proper numbers, and set our service accordingly. I agree that this is a different scope, however.

I would stick this in a new Jira issue so that we can move forward and not lose track of this while we figure out exactly what we should do with it.

docs/reference/juju-oidc-config.rst
ajzobro: Can we/should we say more to that effect: that it's an anonymous client.

E.g. here are some words that you might be able to build from:
When the `oidc_client_secret` is not set, the connection to the OIDC provider will be anonymous if unauthenticated public clients are allowed. Such is the case in our development environment between `dex` and `testflinger`

docs/how-to/administer-server/maintain-testflinger-server.rst
ajzobro: LGTM

ajzobro (dismissed):
LGTM

docs/reference/juju-oidc-config.rst
tang-mm: Agreed that enforcing a min length is a separate concern, and I'll leave whether/how to add validation up to you. 
My original question was more about documentation accuracy; since it's unnecessary to state that "there's no enforced min length", I think keeping it as-is, in line with Flask API is good enough

tang-mm (dismissed):
Looks great!

docs/reference/juju-oidc-config.rst
rene-oromtz: As discussed, I added two references in [16ab948](https://github.com/canonical/testflinger/pull/1254/commits/16ab948d539b151f1387f6e0d84ac43ab8735c46) to point to OIDC resources which should provide more information on OIDC in general

ajzobro (approved):
Thanks for the additional reference links, Rene!

PR #1257, opened 2026-07-20
Feature/dev maas more detail

ajzobro:
## Description

Improves MAAS provisioning diagnostics and simplifies CLI OIDC auth. The MAAS connector now verifies the deployed node's IP matches the configured device_ip, captures and logs installation logs on deployment failures, and retries faster (10s backoff) with full error history. CLI auth removes the disk-polling monitor_for_refresh_token mechanism in favor of direct server polling and uses a single config file path.

## Resolved issues

- MAAS deployments with wrong device IP now fail with a clear error instead of timing out.
- Deployment failures now surface installation logs for diagnostics.
- Retry backoff reduced from 60s to 10s for faster failure detection.

## Documentation

Removed OIDC-related docs (oidc-auth.rst, juju-oidc-config.rst, and the "Enable OIDC" how-to section).

## Web service API changes

None.

## Tests

Updated MAAS tests to mock new node_addresses and get_current_installation_id calls. Updated CLI auth tests for simplified OIDC polling and config path; removed tests for deleted monitor_for_refresh_token.


device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
```diff
@@ -366,6 +371,12 @@ def deploy_node(
         ]
         # Do not use runcmd for this - we need the output, not the end user
         proc = self.run_maas_cmd_with_retry(cmd)
+
+        # Before starting, collect some important diagnostic information that
+        # might be useful later. Specifically, identify what the
+        # current-installation log id is so that we know when we get a new one
+        self.starting_installation_id = self.get_current_installation_id()
```
rene-oromtz: nitpick: I don't mind verbosity but in this case, I thing given the only "diagnostic" is the current installation ID, this can be simplified to something as:
"Before starting, identify what the current-installation log id is so that we know when we get a new one"

If you still prefer the verbosity, feel free to ignore and mark this as Resolved

device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
```diff
@@ -450,11 +472,33 @@ def check_test_image_booted(self):
                 cmd, stderr=subprocess.STDOUT, timeout=60, check=True
             )
         except subprocess.SubprocessError:
+            # Since ssh wasn't working (yet) let's check the installation log
+            # to make sure that the correct ip address is being used, as this
+            # has been a common problem with the MAAS setup.
+            txt_log = self.get_deployment_information()
+            if txt_log is None:
```
rene-oromtz: I personally don't think we should rely on the deployment information as it may not be stable or behavior change overtime
The deployment info is useful for including the logs from MAAS so it should only be called on failure. For IP, I perhaps would add a helper function eg. `get_node_ip(self)` with the same exact pattern `node_status(self)` but looking at the `ip_addresses` key. Its a list so we can check if the IP is in that list. 

This would also look clearer searching by key from the JSON output instead of searching string in a potential large output

rene-oromtz: Another alternative could be to extend `node_status` to return a tuple with status and ip as I could be sold that the IP is part of a "node status" and would reduce the amount of calls of MAAS CLI as we are already using this method on line 419. 

the only downside is match all other calls

device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
```diff
@@ -544,3 +588,67 @@ def get_maas_version(self) -> tuple[int, ...] | None:
...
+            "read",
+            self.node_id,
+            "type=installation",
+        ]
+
+        proc = self.run_maas_cmd_with_retry(cmd)
+        json_out = json.loads(proc.stdout.decode())
+        if not isinstance(json_out, list) or not json_out:
+            return None
+        json_out = [x for x in json_out if isinstance(x, dict)]
```
rene-oromtz: My only suggestion for this is make it more readable by including what "x" means in this context

device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
```diff
@@ -544,3 +588,67 @@ def get_maas_version(self) -> tuple[int, ...] | None:
...
+        if not json_out:
+            return None
+
+        # MAAS team suggested sorting by id and taking the largest
```
rene-oromtz, line 624: I can't seem able to add another comment on line 620 but same comment as above. Just replace "x" with a more readable value even if this is more verbose

rene-oromtz (changes requested):
Thanks for the changes and for also including the missing typehints! 
I added some comments, some of them are more suggestion-like but the one for getting the node_ip in a more stable way I think is worth to check

device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
```diff
@@ -287,19 +293,32 @@ def run_maas_cmd_with_retry(self, cmd, max_retries=5, backoff_start=60):
...
+            # track and share the history of failed attempts rather than only
+            # the last.
+            # The back-off timing was also drastically reduced from 60 seconds
+            # with 5 retries (total 31 minutes per maas command) to 10 seconds
+            # with 5 retries (total just over 5 minutes per command)
+            # The goal is to fail fast if we cannot make any head-way.
+            # Further improvements could be:
+            #  - compare the recurring maas responses to the same command to
+            #    determine if there is any change. If there is no change then
+            #    perhaps even the default of 5 retries is still too much.
```
rene-oromtz: I will probably just keep the first part. 
The backoff justification feels something worth to keep in the description of this PR as then we can easily blame and get this justification instead of justify here in code as this may later be easy to miss or can get outdated if we then chose to increase it back or even reduce this. 

As for the further improvements, this feels more or a Jira tracking, perhaps it will end up being in the backlog for a long time but should be easier to get back to this than refer to code:

```suggestion
            # Each time we error, we may or may not get a different error,
            # since this will be all the information that we know, we should
            # track and share the history of failed attempts rather than only
            # the last.
```

device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
ajzobro: 1) the deployment information may not be stable -- neither is our agent, if our dependencies change we need to also change.
2) the deployment info is useful for including in the logs -- yes
    a) it should only be called on failure -- no, it can be used to determine if the installation was successful and can do so potentially faster than the ssh route, I'm collecting the whole thing (whatever it may be) so that we can start with a search for the IP (useful, known today) and expand on that in the future once we have this information to start making further decisions with.

If we don't start collecting all of the useful information that we can, we will never know what information is there to build our solution on.

I have zero desire to collect *less* information given this opportunity to collect information could be beneficial in the future.

`value in string` is going to search the blob at least as fast as all the work it would take to reduce that blob to some other format and lose all the text that we want to keep and print on line 503 in the case where we DO have a problem.

Device IP as a short-circuit bail-out was a bonus.

Collecting ALL the data was the whole point from the start.

"number of maas cli calls" -- I don't understand why we want to reduce these?  We want:
 - fast failure
 - all the details on failure

I believe that we are already only calling maas on line 492 and that will only happen after each minute of no ssh.

device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
```diff
@@ -287,19 +293,32 @@ def run_maas_cmd_with_retry(self, cmd, max_retries=5, backoff_start=60):
...
             if retry_count > max_retries:
                 self._logger_error(
                     (
-                        f"maas error running: {' '.join(cmd)}"
+                        f"maas error running: {' '.join(cmd)}; "
                         "maximum retries reached"
                     )
                 )
-                raise ProvisioningError(proc.stdout.decode())
+                raise ProvisioningError(f"{errors}")
```
rene-oromtz, line 308: I guess my only concern with this is how often are we expect this to change the reason between retries. 
Something worth to test but if we are getting a bad response e.g. a 5xx error, I think its safe to say that we will get the same response over the next 5 min. 

Perhaps this is more justifiable with the longer retries but I can see this potentially be just a repetition of the same error 5 times

device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
rene-oromtz: In this case I'm not saying collect less info, I'm just pointing that that instead of checking for an IP in a log, we can just do it from `maas machine read` which also returns the IP in a nicer more parsable way. 
Line 503 can then be just to display the info that you want:
`self._logger_info(self.get_deployment_information())`

ajzobro: I'm struggling to see the benefit or the optimization.

`maas machine read` is faster, and I can add it before the ssh call so we can save even more time, possibly

device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
ajzobro: If we get an error from a server; and we try again... and again... and again...
It would be nice to be able to see the inefficiency so that we can make things better.  Presently we are blind.

ajzobro: I also would question if the 500 series errors in the proposed example would go away after 5, 15, or 60 minutes in this proposed scenario.  I don't know what normal situations we are expecting these retries to fix, but waiting up to 31 minutes for every maas command was a little crazy.  We can still wait up to 31 * base = 310 seconds for a single maas command to go through over a 5 minute window.

If the point was to survive long network issues, then I have broken that in favor of a fail-fast and provide as much detail mentality.

rene-oromtz: Yeah, fail fast makes sense, as if its noisy, guess we can figure it out eventually. For now, its okay to keep as-is

device-connectors/src/testflinger_device_connectors/devices/maas2/tests/test_maas.py
```diff
@@ -423,3 +435,255 @@ def test_non_ephemeral_deploy(
...
+        returncode=0,
+        stdout=json.dumps([{"id": 3}, {"id": 10}, {"id": 7}]).encode(),
+    )
+
+    assert maas2.get_current_installation_id() == 10
+    mock_run_cmd.assert_called_once_with(
+        [
+            "maas",
+            maas2.maas_user,
+            "node-script-results",
```
rene-oromtz:
```suggestion
            "node-script-result",
```

rene-oromtz (changes requested):
LGTM! There is only one test that is failing but I added a quick comment on that, I think that is all that's needed

device-connectors/src/testflinger_device_connectors/devices/maas2/tests/test_maas.py
ajzobro: Good catch, thanks!

rene-oromtz (approved):
Thanks for the changes, LGTM

PR #1271, opened 2026-07-30
Feat/device maas better recovery

ajzobro:
## Description

Incorporated feedback from the  MAAS team for how to deal with a stuck Release.

## Resolved issues

Recurrent stuck in Releasing issue

## Documentation

n/a

## Web service API changes

N/a

## Tests

yes, added tests for this and a few other functions.

### DIFF
--- a/device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
@@ -576,8 +576,8 @@ class Maas2:
         self._logger_info("Attempt to just power-off the machine")
         cmd = ["maas", self.maas_user, "machine", "power-off", self.node_id]
         proc = self.run_maas_cmd_with_retry(cmd)
+        output = proc.stdout.decode()
         if self.debug:
-            output = proc.stdout.decode()
             # Use logger_info to not modify agent config file
             self._logger_info(output)
 
--- a/device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
+++ b/device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
@@ -533,7 +533,7 @@ class Maas2:
             if status == "Ready":
                 return
         self._logger_error(
-            'Device {} still in "{}" state, attemtping recovery!'.format(
+            'Device {} still in "{}" state, attempting recovery!'.format(
                 self.agent_name, status
             )
         )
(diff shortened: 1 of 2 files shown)


device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
```diff
@@ -533,6 +532,64 @@ def node_release(self) -> None:
...
+
+    def _recover_node_control(self):
+        """Attempt to recover the node by any means necessary."""
+        # Gentle kick: recovery method one: mark-broken, mark-fixed.
+        self._logger_info("Attempt to mark-broken:")
+        cmd = ["maas", self.maas_user, "machine", "mark-broken", self.node_id]
+        proc = self.run_maas_cmd_with_retry(cmd)
+        if self.debug:
+            output = proc.stdout.decode()
+            # Use logger_info to not modify agent config file
```
rene-oromtz, line 550: Can you please clarify on this comment? Not sure the intention on using logger_info to not modify agent config file

device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
```diff
@@ -533,6 +532,64 @@ def node_release(self) -> None:
...
+            self._logger_info(output)
+        if not proc.returncode:
```
rene-oromtz, line 552: I guess this is fine given the previous command succeeded but I wonder if we should check the status as well before attempting to mark it as fixed e.g. check mark-broken succeeded AND check node_status == "Broken"

I assume if the command succeeded then the change will be reflected immediately but not sure if there will be a race condition between MAAS marking the device as Broken and attempting to mark it as fixed.

rene-oromtz: Worst case scenario I guess we will get to the brute force, but what do you think?

rene-oromtz (commented):
Thanks for the changes! LGTM, Copilot flagged a typo and I just added some questions. 
Those are not necessarily a request for changes as I think its good as-is but wanted to get your opinion first.

device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
ajzobro: I don't know where that came from, I assume it's copy/pasted -- but it's non-sensical

device-connectors/src/testflinger_device_connectors/devices/maas2/maas2.py
ajzobro: Right, I'm thinking straight to brute force.  I know that works, but if you think that we need to allow some time between commands (I don't think we ought to -- the server side should sort that out, I think) we could add that.

I want to waste as little time as possible since we already waited at least 5 minutes without doing anything earlier.

When I used the power command earlier it was near immediate.

rene-oromtz: Yeah, its fine, just wondering how long will it actually take for it to be marked as broken.

rene-oromtz (approved):
LGTM! Thanks for the changes

PR #1284, opened 2026-08-19
docs: add agent navigation guidance

rene-oromtz:
## Description
This adds the necessary files for allowing agents to read/write repository knowledge base allowing them to navigate more effectively in this monorepo. 

It aligns with the conventions defined in [canonical-agents](https://github.com/canonical/canonical-agents/tree/main)

- `.kb/` directory: contains the required `agents.md` copy along with relevant architecture and testing considerations
- top level `AGENTS.md` 
- individual `AGENTS.md` for relevant directories

All files are kept as terse as possible

## Resolved issues
Resolves [CERTTF-1131](https://warthogs.atlassian.net/browse/CERTTF-1131)

## Documentation
NA

## Web service API changes
NA

## Tests
NA

[CERTTF-1131]: https://warthogs.atlassian.net/browse/CERTTF-1131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.kb/agents.md
+++ b/.kb/agents.md
@@ -1,0 +1,44 @@
+# Preface
+
+This repository follows strict conventions for organizing agent-oriented knowledge documents, and this document is required reading for any agent that wants to read or write these.
+
+Read the top-level `.kb/agents.md` before continuing below.
+
+
+# Overview
+
+Every directory in this repository, including the root, may have its own `AGENTS.md` file and `.kb/` subdirectory. The AGENTS.md file provides a general view for the directory, while more specific information is found in `.kb/*.md` files with dashed lowercase names (e.g. `.kb/special-relativity.md`).
+
+The design of this structure has the following key goals:
+
+- **Mechanical** - Agents are the main actors reading and writing the knowledge base.
+- **Generic** - Benefits any agentic workflow, no matter the editor or platform.
+- **Distilled** - Avoids the use of verbose task or plan logs that pollute the context window.
+- **Hierarchical** - Avoids excessive information in a single place that also pollutes the context window.
+- **Human** - Information is readily available and reviewable in a friendly format.
+
+
+# Important
+
+- Read local `AGENTS.md` files upon navigating directories.
+- Keep the `.kb/*.md` files updated whenever there is something relevant to be documented or updated.
+- Follow the header conventions outlined below. Only the _Preface_ header is required, and the other headers should be omitted if empty or trivial.
+
+
+# Headers
+
+The following are the ONLY top-level headers allowed across the `.kb/*.md` files in this repository, to maintain semantic standardization across projects.
+Sub-headers are okay.
+
+- _Preface_ - A brief introduction outlining the scope and relevance of a specific `*.md` file. This section MUST be at the top of every `.kb/*.md` file so agents can easily grep for it, and the last line of this section MUST be "Read the top-level `.kb/agents.md` file before continuing below." so rules are followed.
+- _Overview_ - High-level summary of the directory, subsystem or knowledge base layout at large. Do NOT use this to list files or directories.
+- _Important_ - Essential directives for the agent outlining critical constraints, behaviors, or rules.
+- _Headers_ - Global registry of header definitions, uniquely hosted at the root `.kb/agents.md` (this file). Do NOT use this header in any other document.
+- _Architecture_ - Structural design details or boundary explanations for a given component. Only use this for software architecture concepts, NOT for directory outlining. Also avoid using this as a code reference (keep that in the code itself).
+- _Directory_ - Only in `AGENTS.md` files, it briefly outlines the contents and structure of the directory the `AGENTS.md` file is in, and potentially nested small directories that do not justify their own `AGENTS.md` file. For `.kb/*.md` files, use _Documents_ instead.
+- _Documents_ - Only in `AGENTS.md` files, it outlines the content of `.kb/*.md` files and also immediately nested `AGENTS.md` child files, to aid agent navigation. It MUST be the last section in the file. It's okay to also mention such filenames inline when they are relevant elsewhere in the text.
+
+For the _Directory_ and _Documents_ listings, format items as a dashed list starting with the file or directory name surrounded by backticks, a dash, and then a brief description:
+```
+- `filename` - Terse summary.
+```
--- a/.kb/testing.md
+++ b/.kb/testing.md
@@ -1,0 +1,14 @@
+# Preface
+
+Explains the cross-component developer workflow for building, linting, and testing the Python components (`agent/`, `cli/`, `common/`, `device-connectors/`, and `server/`). Relevant whenever running or adding checks in those components.
+Read the top-level `.kb/agents.md` file before continuing below.
+
+
+# Important
+
+- Each subproject (`agent/`, `cli/`, `common/`, `device-connectors/`, `server/`) manages its own dependencies with `uv`; run `uv sync` inside a component directory before working on it in isolation.
+- `just` is the task runner, with a modular justfile per component plus one at the repository root. Run `just` from any directory to list the recipes available there, or `just <component>::<recipe>` from the root to target a specific component.
+- The common recipes across components are `just format`, `just lint`, `just test`, and `just check` (runs format, lint, and test together; the root `just check` also lints GitHub workflows with `zizmor`).
+- Install git hooks with `just pre-commit` (uses `prek` to support the monorepo layout).
+- Prefer running the narrowest applicable `just check`/`just test` for the component you changed before considering work done.
+- Documentation uses separate `just` recipes that delegate to Make; read `docs/AGENTS.md` when changing it.
--- a/.kb/testing.md
+++ b/.kb/testing.md
@@ -9,6 +9,5 @@ Read the top-level `.kb/agents.md` file before continuing below.
 - Each subproject (`agent/`, `cli/`, `common/`, `device-connectors/`, `server/`) manages its own dependencies with `uv`; run `uv sync` inside a component directory before working on it in isolation.
 - `just` is the task runner, with a modular justfile per component plus one at the repository root. Run `just` from any directory to list the recipes available there, or `just <component>::<recipe>` from the root to target a specific component.
 - The common recipes across components are `just format`, `just lint`, `just test`, and `just check` (runs format, lint, and test together; the root `just check` also lints GitHub workflows with `zizmor`).
-- Install git hooks with `just pre-commit` (uses `prek` to support the monorepo layout).
 - Prefer running the narrowest applicable `just check`/`just test` for the component you changed before considering work done.
 - Documentation uses separate `just` recipes that delegate to Make; read `docs/AGENTS.md` when changing it.
--- a/.kb/testing.md
+++ b/.kb/testing.md
@@ -9,5 +9,5 @@ Read the top-level `.kb/agents.md` file before continuing below.
 - Each subproject (`agent/`, `cli/`, `common/`, `device-connectors/`, `server/`) manages its own dependencies with `uv`; run `uv sync` inside a component directory before working on it in isolation.
 - `just` is the task runner, with a modular justfile per component plus one at the repository root. Run `just` from any directory to list the recipes available there, or `just <component>::<recipe>` from the root to target a specific component.
 - The common recipes across components are `just format`, `just lint`, `just test`, and `just check` (runs format, lint, and test together; the root `just check` also lints GitHub workflows with `zizmor`).
-- Prefer running the narrowest applicable `just check`/`just test` for the component you changed before considering work done.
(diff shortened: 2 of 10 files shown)


agent/AGENTS.md
```diff
@@ -0,0 +1,14 @@
...
+
+
+# Directory
+
+- `src/testflinger_agent/` - Agent daemon source: job polling, config, job lifecycle, and test execution.
+- `tests/` - Unit tests.
+- `charms/testflinger-agent-host-charm/` - Juju charm for hosting multiple agents on one machine via Supervisor.
+- `terraform/` - Infrastructure-as-code for agent deployment.
+- `extra/` - Docker and test-environment helpers.
+- `justfile` - Task runner (format/lint/test/check).
```
ajzobro, line 14:
```suggestion
- `justfile` - Task runner (format/lint/test/check) -- use this for all tests/checks
```

.kb/testing.md
```diff
@@ -0,0 +1,14 @@
...
+Explains the cross-component developer workflow for building, linting, and testing the Python components (`agent/`, `cli/`, `common/`, `device-connectors/`, and `server/`). Relevant whenever running or adding checks in those components.
+Read the top-level `.kb/agents.md` file before continuing below.
+
+
+# Important
+
+- Each subproject (`agent/`, `cli/`, `common/`, `device-connectors/`, `server/`) manages its own dependencies with `uv`; run `uv sync` inside a component directory before working on it in isolation.
+- `just` is the task runner, with a modular justfile per component plus one at the repository root. Run `just` from any directory to list the recipes available there, or `just <component>::<recipe>` from the root to target a specific component.
+- The common recipes across components are `just format`, `just lint`, `just test`, and `just check` (runs format, lint, and test together; the root `just check` also lints GitHub workflows with `zizmor`).
+- Install git hooks with `just pre-commit` (uses `prek` to support the monorepo layout).
```
ajzobro: This is a human rule, I don't let the agent do any git commands, and since I run in a workshop, I provide the diffs to the agent via `git diff > gitdiff.txt` to give the agent context to review.

This is not ideal as it doesn't show the full set of changes for the branch, but it does allow for controlling the context under review.

.kb/testing.md
```diff
@@ -0,0 +1,14 @@
...
+- Prefer running the narrowest applicable `just check`/`just test` for the component you changed before considering work done.
```
ajzobro, line 12: I always run all tests.

.kb/testing.md
```diff
@@ -0,0 +1,14 @@
...
+- Documentation uses separate `just` recipes that delegate to Make; read `docs/AGENTS.md` when changing it.
```
ajzobro, line 13:
```suggestion
- Documentation uses separate `just` recipes that delegate to Make; read `docs/AGENTS.md` when changing it.
- Keep lines under 79 char, run `just check` to verify
```

AGENTS.md
```diff
@@ -0,0 +1,40 @@
...
+
+- `.kb/agents.md` - General rules for the knowledge base reading and writing.
+- `.kb/architecture.md` - How the server, agents, device connectors, CLI, and common library fit together.
+- `.kb/testing.md` - Cross-component developer workflow: `uv`, `just` recipes, and pre-commit setup.
+- `agent/AGENTS.md` - Agent daemon navigation.
+- `cli/AGENTS.md` - CLI navigation.
+- `common/AGENTS.md` - Shared-library navigation.
+- `device-connectors/AGENTS.md` - Device-connector navigation.
+- `docs/AGENTS.md` - Documentation-site navigation.
+- `server/AGENTS.md` - Server navigation.
```
ajzobro, line 40:
```suggestion
- `.kb/agents.md` - General rules for the knowledge base reading and writing
- `.kb/architecture.md` - How the server, agents, device connectors, CLI, and common library fit together
- `.kb/testing.md` - Cross-component developer workflow: `uv`, `just` recipes, and pre-commit setup
- `agent/AGENTS.md` - Agent daemon navigation
- `cli/AGENTS.md` - CLI navigation
- `common/AGENTS.md` - Shared-library navigation
- `device-connectors/AGENTS.md` - Device-connector navigation
- `docs/AGENTS.md` - Documentation-site navigation
- `server/AGENTS.md` - Server navigation
```

Why are we wasting tokens on punctuation?

ajzobro (commented):
Quick first reactions

ajzobro:
Overall, and generally, I believe that less is more.  I have been using agents without any project context and have only had to teach it a marginal amount. I think, overall, that there is a lot more being put into these files than absolutely necessary, including punctuation.

I was thinking about this after we (@rene-oromtz) talked about yesterday and I came up with effectively a 3-line AGENTS.md file that I would have proposed at the toplevel.

```
Keep lines under 79 char
use `just` commands to check syntax and run tests
Favor small diffs for easy review
```

When I run the agent in the workshop, i occasionally tell it things like this:
```
You are running along side docker in a container with access to the demo server/services, you have no access to git
```

I think all of this extra context is potentially unnecessary and borderline wasteful, overall, but I will experiment with these new files before I say more on the topic.

On the flip-side: If we broke down further and the first directive was to not load unnecessary things into context, maybe that could be good.

The only things that I have had to tell the agents to be successful (with VERY small token use) are the tidbits I shared above.

.kb/testing.md
rene-oromtz: Yep, agree, I can remove this rule.

.kb/testing.md
rene-oromtz: Yes, but I can also see a benefit of leaving as-is. Specially if during a session the agent already did any format + lint checks so the "narrowest" path would be to just run test instead of going through formatting and linting again.

.kb/testing.md
rene-oromtz: Not sure if we should add the line length here... I feel agents knowing they can use format or lint is enough, failures are clearly displayed there. 
That way this is kept simple and we could later update line length on the individual components without needing to update this file

AGENTS.md
rene-oromtz: In this case removing the punctuation only save us "one" token used 148 vs 147 for this bullet list (tested with [OpenAI tokenizer](https://platform.openai.com/tokenizer))

Given this is not a huge save and for consistency with the other files (including .kb/agents.md) I would prefer to leave it as-is

agent/AGENTS.md
rene-oromtz: I think this should be covered in the` testing.md ` (to use just for testing) and here is only displaying which are the available recipes in this directory. I would prefer to keep it simple here and if needed, perhaps add a modification in the `testing.md` file?

Otherwise I feel all other AGENTS.md file will need to add this line which is applicable to all components

AGENTS.md
ajzobro: one times all the places it's in times all the agents that are spawned.
We wrote these files like humans are reading them.

.kb/testing.md
```diff
@@ -0,0 +1,13 @@
+# Preface
+
+Explains the cross-component developer workflow for building, linting, and testing the Python components (`agent/`, `cli/`, `common/`, `device-connectors/`, and `server/`). Relevant whenever running or adding checks in those components.
+Read the top-level `.kb/agents.md` file before continuing below.
+
+
+# Important
+
+- Each subproject (`agent/`, `cli/`, `common/`, `device-connectors/`, `server/`) manages its own dependencies with `uv`; run `uv sync` inside a component directory before working on it in isolation.
+- `just` is the task runner, with a modular justfile per component plus one at the repository root. Run `just` from any directory to list the recipes available there, or `just <component>::<recipe>` from the root to target a specific component.
```
ajzobro, line 10:
```suggestion
- `just` is the task runner, with a modular justfile per component plus one at the repository root. Run `just` from any directory to list the recipes available there, or `just <component>::<recipe>` from the root to target a specific component.
   - before calling work complete, run `just check` at the top level and fix any formatting errors; note: when run within a workshop the agent will need to AVOID applying fixes for zizmor due to `gh` not being present in the workshop environment.
```

.kb/agents.md
```diff
@@ -0,0 +1,44 @@
+# Preface
+
+This repository follows strict conventions for organizing agent-oriented knowledge documents, and this document is required reading for any agent that wants to read or write these.
+
+Read the top-level `.kb/agents.md` before continuing below.
```
ajzobro, line 5: self-referencing to "top-level `.kb/agents.md`"

AGENTS.md
rene-oromtz: Removed punctuation marks as per your suggestions. I removed it from all Directory/Documents references

.kb/testing.md
rene-oromtz: Added this as suggested and replaced the other "before calling work complete"

.kb/testing.md
rene-oromtz: I ended up using the `just check` as suggested before calling work done

.kb/agents.md
rene-oromtz: This file is coming as-is from source: https://github.com/canonical/canonical-agents/blob/main/.kb/agents.md

Given this, my personal preference would be to keep it as-is but if you prefer, I can remove this line.

.kb/testing.md
ajzobro: This is the most common mistake introduced by the agents -- if we aren't going to add it here, then where?  It's a rule that needs to be followed, the most common things that `just check` finds ar line-length issues introduced by agents.

.kb/agents.md
ajzobro: I suppose that the agent ought to hit that line and say "that's what i'm trying to do now" and continue along.

.kb/testing.md
rene-oromtz: I could add something like:

> - After adding changes validate formatting with `just lint` 
This one time run is perhaps enough to instruct the agent about line length for future modification without hardcoding the value?

If you still prefer to include the line length I will include it but I think for this purpose `just lint` is the correct to validate 79 char without needed to run the whole format + lint + test

Something like:
> - Keep lines under 79 char, run `just lint` to verify

The `just check` is already included "before calling work done" as suggested in other comment

ajzobro: That will solve one problem.

IMO, I always have it "run `just check` and resolve any issues found"

If these bulky AGENT.md files aren't going to tell the agents what I always already have to tell them, then they aren't solving real problems.

The fact there is a 79 char rule is a concrete rule, enforced after-the-fact by the linting -- we should aspire to avoid the problem in the first place by being up-front with the actual rule so that the agents stay within the lines.

In this singular line-length case, the rule is that lines within python files need to be under X characters, with X being defined concretely somewhere.

If we can point the agent at all the real rules then it can make compliant code from the get-go.

General feedback:

1) there are too many files here for ME to keep track of and the information is spread out rather than consolidated into a singular place.

a) if the goals of the AGENTS.md files in the sub-directories is to describe the sub-components, then fine, keep them there and restricted to that specific scope.
b) at the top-level we have rules that need to be followed, I would like to know what specific file contains the rules. From my experience we need to have 3 rules defined to counter the common mistakes that the agents make.  The rest of this is educational "this directory is for X purpose"

"Keep lines under 79 char, run just lint to verify" yes, but is covered by `just check`;

Rules that I would give to the agents:
- keep deltas concise and limit unneccesary changes to improve on the ability to review the changes.
- lines are capped at 79 char
- use `just check` and `just test` to validate

If CONTRIBUTING.md is for Humans, then AGENTS.md should describe at least the same concepts in a reduced-token-count way for the agents.

Those are rules for the agent, where do rules go?  I guess _Important_ is the section that these would go in.

ajzobro: Oh! Another thing I've had to tell the agents: "We use Python 3.10 exclusively" -- as, for example datetime changed in 3.11 and so the handling of timezones and the use of 'Z' has changed.

rene-oromtz: Added ([8437475](https://github.com/canonical/testflinger/pull/1284/commits/84374753908cde9a8c0cad37000396cb516f70c8)) the following two rules in the _Important_ section for the top-level `AGENTS.md`:
```
- Keep deltas concise and limit unnecessary changes to improve the ability to review the changes.
- Keep lines under 79 characters.
```

The other rule should be covered by the `Before calling work complete, run`just check....` in the _Important_ section of `testing.md`

This separate concerns on how the agent needs to add changes vs how the agent needs to tests those changes (while still including all of your suggestions). Let me know what you think and if we can mark this as resolved

rene-oromtz: Also added ([f730a7a](https://github.com/canonical/testflinger/pull/1284/commits/f730a7aae77f8a6083d2c4aec5a2bdfefabed7ab)):
`- All code changes should be supported by Python 3.10.`

PR #1286, opened 2026-08-24
Update agent to use pathlib for paths, addresses syntax/type warnings

ajzobro:
## Description

There were type warning about the use of incorrect data types with respect to paths. This PR introduces the use the pathlib Paths to handle paths in the agent.

## Resolved issues

N/A

## Documentation

N/A

## Web service API changes

N/A

## Tests

Still pass, test update to use new data type.


agent/src/testflinger_agent/agent.py
```diff
@@ -463,8 +460,11 @@ def process_jobs(self):
                 # Other errors can happen too for things like connection
                 # problems
                 logger.exception(e)
-                results_basedir = self.client.config.get("results_basedir")
-                shutil.move(rundir, results_basedir)
+                results_basedir = Path(
+                    self.client.config.get("results_basedir")
+                )
+                if rundir is not None:
+                    shutil.move(str(rundir), results_basedir)
```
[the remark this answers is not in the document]
ajzobro: Noted: will add guard at 463

agent/src/testflinger_agent/job.py
```diff
@@ -122,9 +122,9 @@ def run_test_phase(self, phase, rundir):
                     "Skipping %s phase as requested in job data", phase
                 )
                 return 0, None, None
-        results_file = Path(rundir) / "testflinger-outcome.json"
-        output_log = Path(rundir) / f"{phase}.log"
-        serial_log = Path(rundir) / f"{phase}-serial.log"
+        results_file = rundir / "testflinger-outcome.json"
+        output_log = rundir / f"{phase}.log"
+        serial_log = rundir / f"{phase}-serial.log"
```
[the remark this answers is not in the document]
ajzobro: I think it is not optional to have a rundir for a test phase, the def is wrong.

agent/src/testflinger_agent/client.py
```diff
@@ -341,24 +341,24 @@ def transmit_job_outcome(self, rundir):
...
             return
 
         with tempfile.TemporaryDirectory() as tmpdir:
-            artifact_file = os.path.join(tmpdir, "artifacts")
+            artifact_file = str(Path(tmpdir) / "artifacts")
             shutil.make_archive(
                 artifact_file,
                 format="gztar",
-                root_dir=rundir,
+                root_dir=str(rundir),
```
rene-oromtz, line 361: Different than shutil.move, this is not documented as supported 🤔 I was looking at some GH [issues](https://github.com/python/cpython/issues/85809) and perhaps may now be supported in certain Python versions but given its not officially documented I won't add a suggestion comment for this, if any, perhaps a TODO to use path-like objects whenever officially supported but not a big deal

agent/src/testflinger_agent/agent.py
```diff
@@ -454,15 +450,18 @@ def process_jobs(self):
...
+                    self.client.transmit_job_outcome(rundir)
+                except Exception as e:
+                    # TFServerError will happen if we get other-than-good
+                    # status; other errors can happen too for things like
+                    # connection problems
+                    logger.exception(e)
+                    results_basedir = Path(
+                        self.client.config.get("results_basedir")
+                    )
+                    shutil.move(str(rundir), results_basedir)
```
rene-oromtz: nitpick, given we are on 3.10 path objects are supported by shutil.move (since 3.9) 
Ref: https://docs.python.org/3/library/shutil.html#shutil.move
```suggestion
                    shutil.move(rundir, results_basedir)
```

rene-oromtz (dismissed):
Thanks for the changes, just added couple nitpick comments in case you want to consider them. 
Not a big deal but for consistency of using Path-like objects whenever supported

agent/src/testflinger_agent/client.py
ajzobro: We CAN update to a later version of Python now that we don't have to support RPC calls.

rene-oromtz: Yep, I still probably won't add this until officially documented but we can definitively review on bumping the python version (and OS version for charms)

rene-oromtz (approved):
LGTM

PR #1288, opened 2026-08-25
Add client id (submitted by) to API endpoints.  Also adjusts the data…

ajzobro:
# Add `submitted_by` and `cancelled_by` to the Testflinger API

## Summary

Exposes job identity fields through the API that were previously only visible in the web interface.

### New fields

- **`GET /job/{job_id}`** — returns `submitted_by` (who submitted the job)
- **`GET /result/{job_id}`** — returns `cancelled_by` (who cancelled the job, `null` if not cancelled)
- **`GET /agents/data`** / **`GET /agents/data/{agent_name}`** — returns `job_submitted_by` (who submitted the job the agent is currently running)

### Internal rename

The job document field `client_id` has been renamed to `submitted_by` throughout — in MongoDB documents, templates, and all database queries. This is a clean rename with no migration required as no data exists on main yet.

`client_id` is retained where it refers to authenticated user identity (credentials, permissions, secrets, tokens) — those are unrelated concepts.

### Database

`get_agents` and `get_agent_info` now use a single MongoDB aggregation pipeline (`$lookup`) to join agent documents with their active job in one query, replacing the previous two-query approach.

`cancel_job` now accepts and stores the `client_id` of the cancelling user as `result_data.cancelled_by`.

`get_job` added as a shared function used by both the web views and API, replacing a direct `mongo.db.jobs.find_one` call in `views.py`.

### Docs

`docs/reference/api-roles.rst` updated to add the missing `GET /v1/` entry.

### DIFF
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -99,6 +99,7 @@ FIELDS_CHOICES = (
     "provision_type",
     "comment",
     "job_id",
+    "submitted_by",
     "queues",
 )
 
@@ -513,6 +514,20 @@ class TestflingerCli:
             default=None,
             help="Filter agents by comment (regex)",
         )
+        parser.add_argument(
+            "--filter-job-id",
+            dest="filter_job_id",
+            type=helpers.regex_arg,
+            default=None,
+            help="Filter agents by current job ID (regex)",
+        )
+        parser.add_argument(
+            "--filter-submitted-by",
+            dest="filter_submitted_by",
+            type=helpers.regex_arg,
+            default=None,
+            help="Filter agents by the submitter of their current job (regex)",
+        )
 
     def _add_queue_status_args(self, subparsers):
         """Command line arguments for queue status."""
@@ -749,6 +764,14 @@ class TestflingerCli:
                 return lambda a: regex.search(str(a.get(field, "")))
             return lambda a: True
 
+        def nested_job_re_filter(field: str, regex: object) -> callable:
+            """Filter on a field nested inside agent['job']."""
+            if regex:
+                return lambda a: regex.search(
+                    str((a.get("job") or {}).get(field, ""))
+                )
+            return lambda a: True
+
         # Filter agents by allowed states
         return [
             a
@@ -763,6 +786,10 @@ class TestflingerCli:
                         "provision_type", self.args.filter_provision_type
                     )(a),
                     re_filter("comment", self.args.filter_comment)(a),
+                    nested_job_re_filter("job_id", self.args.filter_job_id)(a),
+                    nested_job_re_filter(
+                        "submitted_by", self.args.filter_submitted_by
+                    )(a),
                 )
             )
         ]
@@ -827,6 +854,16 @@ class TestflingerCli:
         # Map 'status' to 'state' in agent dicts for backward compatibility
         field_map = {"status": "state"}
 
+        def get_field(agent: dict, field: str) -> str:
+            """Resolve a field value from an agent dict, including nested."""
+            if field in ("job_id", "submitted_by"):
+                return (agent.get("job") or {}).get(field, "-")
+            key = field_map.get(field, field)
+            val = agent.get(key, "-")
+            if isinstance(val, list):
+                val = ", ".join(str(v) for v in val)
+            return str(val)
+
         # Header names and valid fields
         header_map = {
             "name": "Name",
@@ -835,22 +872,19 @@ class TestflingerCli:
             "provision_type": "Provision Type",
             "comment": "Comment",
             "job_id": "Job ID",
+            "submitted_by": "Submitted By",
             "queues": "Queues",
         }
         headers = [header_map[f] for f in self.args.fields]
 
         # Calculate column widths
-        col_widths = []
-        for field, header in zip(self.args.fields, headers, strict=False):
-            key = field_map.get(field, field)
-            width = max(
+        col_widths = [
+            max(
                 len(header),
-                max(
-                    (len(str(agent.get(key, "-"))) for agent in agents),
-                    default=0,
-                ),
+                max((len(get_field(a, f)) for a in agents), default=0),
             )
-            col_widths.append(width)
+            for f, header in zip(self.args.fields, headers, strict=False)
+        ]
 
         # Print header
         print(
@@ -861,14 +895,10 @@ class TestflingerCli:
         )
         # Print rows
         for agent in agents:
-            row = []
-            for f, w in zip(self.args.fields, col_widths, strict=True):
-                key = field_map.get(f, f)
-                val = agent.get(key, "-")
-                # Convert list values to comma-separated string
-                if isinstance(val, list):
-                    val = ", ".join(str(v) for v in val)
-                row.append(f"{val:<{w}}")
+            row = [
+                f"{get_field(agent, f):<{w}}"
+                for f, w in zip(self.args.fields, col_widths, strict=True)
+            ]
             print("  ".join(row))
 
     def _print_agent_names(self, agents: list[dict]) -> None:
--- a/cli/testflinger_cli/__init__.py
+++ b/cli/testflinger_cli/__init__.py
@@ -515,13 +515,6 @@ class TestflingerCli:
             help="Filter agents by comment (regex)",
         )
         parser.add_argument(
-            "--filter-job-id",
-            dest="filter_job_id",
-            type=helpers.regex_arg,
-            default=None,
-            help="Filter agents by current job ID (regex)",
-        )
-        parser.add_argument(
             "--filter-submitted-by",
             dest="filter_submitted_by",
             type=helpers.regex_arg,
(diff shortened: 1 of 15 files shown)


server/src/testflinger/api/v1.py
```diff
@@ -621,7 +623,9 @@ def result_get(job_id: str):
 
     # Reconstruct result format with logs and phase statuses
     log_handler = MongoLogHandler(database.mongo)
-    return log_handler.format_logs_as_results(job_id, result_data)
+    result = log_handler.format_logs_as_results(job_id, result_data)
+    result["cancelled_by"] = result_data.get("cancelled_by")
```
ajzobro: no need to add cancelled_by, it's already there.

server/src/testflinger/database.py
```diff
@@ -502,10 +552,26 @@ def get_restricted_queues_owners() -> dict[str, list[str]]:
...
+    :param queue: If provided, filter agents to those listening on this
+        queue.
+    :returns: List of agent dicts, each with a ``job_submitted_by`` field.
+    """
+    match_stage = (
+        {"$match": {"queues": {"$in": [queue]}}} if queue else {"$match": {}}
+    )
+    pipeline = [
+        match_stage,
+        {"$project": {"_id": False, "log": False}},
```
ajzobro: does `log` still exist in agents document?

server/src/testflinger/api/schemas.py
```diff
@@ -71,6 +71,7 @@ class AgentOut(Schema):
     job_id = fields.String(required=False)
     comment = fields.String(required=False)
     restricted_to = fields.Dict(required=False)
+    job_submitted_by = fields.String(load_default=None)
```
rene-oromtz: I honestly would like to have something as:
`job = fields.Nested(JobInfo, required=False)`

JobInfo containing both of the id and the "submitted_by" (instead of prepending `job_*`) so we can have something as:
`job = {"job_id": "id", "submitted_by": "submitter"}` or `job = null` (instead of having two nullable flat fields)

This is a larger change but more semantically correct. This PR could:
1. add that nested structure (but its important that we keep the flat `job_id` for backward compatibility for now)
2. At least add a TODO so we can eventually check if it makes sense to introduce even when potentially breaking. 

My main justification for this change is so we can have consistency on the `submitted_by` field everywhere

server/src/testflinger/database.py
```diff
@@ -439,12 +444,57 @@ def check_queue_restricted(queue: str) -> bool:
     return queue_count != 0
 
 
-def get_agent_info(agent: str) -> dict:
-    """Return the information for a specified agent."""
-    agent_data = mongo.db.agents.find_one(
-        {"name": agent}, {"_id": False, "log": False}
-    )
-    return agent_data
+def _active_job_pipeline_stages() -> list[dict]:
```
rene-oromtz, line 447: From the previous [comment](https://github.com/canonical/testflinger/pull/1288/changes#r3864438436)

If approach 1 is preferred this pipeline will need to change so it returns a nested structure e.g job: {...}  otherwise I think there is a slight improvement that I'll comment next

server/src/testflinger/database.py
```diff
@@ -439,12 +444,57 @@ def check_queue_restricted(queue: str) -> bool:
...
+                "job_submitted_by": {
+                    "$let": {
+                        "vars": {"job": {"$arrayElemAt": ["$_job_docs", 0]}},
+                        "in": {
+                            "$cond": [
+                                {"$gt": [{"$size": "$_job_docs"}, 0]},
+                                "$$job.submitted_by",
+                                None,
+                            ]
+                        },
```
rene-oromtz, line 486: The conditional is redundant as arrayElemAt already return null if out of index. 
```suggestion
            "$addFields": {
                "job_submitted_by": {
                    "$let": {
                        "vars": {"job": {"$arrayElemAt": ["$_job_docs", 0]}},
                        "in": "$$job.submitted_by",
                    }
                }
            }
        },
```

server/src/testflinger/database.py
```diff
@@ -323,7 +328,7 @@ def get_queue_wait_times(queues: list[str] | None = None) -> list[dict]:
 
 
 def get_agents_on_queue(queue: str) -> list[dict]:
```
rene-oromtz: I noticed this was missing the pipeline as it was using the AgentOut schema but this is being used in two places `POST job` and GET `/queues/<queue_name>/agents`

POST should remain lightweight as it doesn't need the agregation but the second one does need to display the submitter otherwise it will always return null. I think this can be covered in a separate line, I'll add a comment about it

server/src/testflinger/api/v1.py
rene-oromtz: This is the line that can be modified so it also returns the submitter:

```suggestion
    agents = database.get_agents(queue=queue_name)
```

rene-oromtz (changes requested):
Thanks for the changes. I just added some suggestions and one comment that is not entirely necessary for this PR but still worth to consider

server/src/testflinger/api/schemas.py
ajzobro: Yes, I initially had this as a nested field, but didn't want to break the existing dependencies; will bring back the nested structure.

server/src/testflinger/database.py
```diff
@@ -439,12 +444,55 @@ def check_queue_restricted(queue: str) -> bool:
...
+                        "vars": {"job": {"$arrayElemAt": ["$_job_docs", 0]}},
+                        "in": "$$job.submitted_by",
+                    }
+                }
+            }
+        },
+                    }
+                }
+            }
+        },
```
rene-oromtz, line 490: My bad, I forgot to select this to be removed :( 
```suggestion
```

server/src/testflinger/api/schemas.py
```diff
@@ -68,9 +81,9 @@ class AgentOut(Schema):
     queues = fields.List(fields.String(), required=False)
     location = fields.String(required=False)
     provision_type = fields.String(required=False)
-    job_id = fields.String(required=False)
```
rene-oromtz, line 71: This is currently included in the agent_status --json response from the CLI. Given this is breaking for potential consumers, perhaps we should consider to keep it for backward compatibility with a TODO so we can later remove it

cli/testflinger_cli/__init__.py
```diff
@@ -513,6 +514,20 @@ def _add_list_agent_args(self, subparsers: object) -> None:
             default=None,
             help="Filter agents by comment (regex)",
         )
+        parser.add_argument(
+            "--filter-job-id",
+            dest="filter_job_id",
+            type=helpers.regex_arg,
+            default=None,
+            help="Filter agents by current job ID (regex)",
+        )
```
rene-oromtz: Do we really want a filter by job id? given its a random UUID I not see much value for a regex search. The only use case I see is for getting the agent for a known job but also feel that this could benefit from its own separate command to get even more information perhaps even "enriching" `testflinger-cli status <job_id>` to get the agent + device ip which is a common request

cli/testflinger_cli/__init__.py
rene-oromtz, line 1: I also just noticed that there are missing unit tests for this CLI code

cli/testflinger_cli/__init__.py
ajzobro: no, overzealous add: anything filterable is displable and vice-versa, wil undo

rene-oromtz (approved):

cli/testflinger_cli/__init__.py
ajzobro: tested locally:

```
$ TESTFLINGER_SERVER="http://testflinger-server:5000" tfcli list-agents --fields=name,status,location,submitted_by
uv run testflinger-cli list-agents --fields=name,status,location,submitted_by
Uninstalled 1 package in 1ms
Installed 1 package in 59ms
NAME    STATUS   LOCATION  SUBMITTED BY   
agent0  waiting  -         infra-agent-arm
agent1  waiting  -         -              
agent2  waiting  -         -              
agent3  waiting  -         -              
agent4  waiting  -         -              
agent5  waiting  -         -              
agent6  setup    -         ci-bot-kernel  
agent7  waiting  -         -              
agent8  waiting  -         -              
agent9  waiting  -         -              
$ TESTFLINGER_SERVER="http://testflinger-server:5000" tfcli list-agents --fi
elds=name,status,location,submitted_by --filter-submitted-by='infra'
uv run testflinger-cli list-agents --fields=name,status,location,submitted_by --filter-submitted-by=infra
NAME    STATUS   LOCATION  SUBMITTED BY   
agent0  waiting  -         infra-agent-arm
$ TESTFLINGER_SERVER="http://testflinger-server:5000" tfcli list-agents --fi
elds=name,status,location,submitted_by --filter-submitted-by='bot'
uv run testflinger-cli list-agents --fields=name,status,location,submitted_by --filter-submitted-by=bot
NAME    STATUS  LOCATION  SUBMITTED BY 
agent6  setup   -         ci-bot-kernel
```

PR #1291, opened 2026-08-27
Feat/agent restart keep state

ajzobro:
# CERTTF-1444: Preserve agent state on restart

## Problem

When an agent process restarted, it unconditionally posted `waiting` as its
new state during `__init__`. This meant that an agent which had been
deliberately taken `offline` or put into `maintenance` by an admin would
silently come back online after any restart (e.g. a system reboot, a
supervisor restart, or an OOM kill), overriding the admin's intent.

## Solution

Replace the unconditional `set_agent_state(WAITING)` call in
`TestflingerAgent.__init__` with a new `_set_startup_state()` method that
first asks the server what state it believes the agent should be in, then
decides accordingly:

| Server state | Action on startup |
|---|---|
| `offline` | Preserve — re-post `offline` with the existing comment |
| `maintenance` | Preserve — re-post `maintenance` with the existing comment |
| `restart` | Come up online (`waiting`) — this is the explicit restart signal |
| `waiting`, `unknown`, any stale job-phase state | Come up online (`waiting`) |
| No state / empty response / 404 / network error | Come up online (`waiting`) — first start or server unavailable |

The `restart` state is intentionally treated as "come back online": it exists
precisely to signal that the agent should restart and resume normal operation.

`_set_startup_state()` delegates to the existing `get_agent_state()` rather
than calling `get_agent_data()` directly, so the heartbeat handler update and
the `UNKNOWN` fallback logic remain in one place.

### `agent/src/testflinger_agent/agent.py`

- `__init__`: removed the unconditional `set_agent_state(WAITING)` call;
  `_post_initial_agent_data()` is now called before state selection so
  metadata is always current; replaced with a call to `_set_startup_state()`.
- `_set_startup_state()`: new method implementing the state-preservation logic
  described above.

### `agent/tests/test_agent.py`

Added `TestStartupState` test class with full parametrized coverage of every
meaningful server response the agent can receive at startup:

- **`test_startup_preserves_offline_and_maintenance`** — parametrized over
  `[OFFLINE, MAINTENANCE]`: confirms the state is re-posted unchanged.
- **`test_startup_comes_up_waiting_for_non_preserved_states`** — parametrized
  over `[RESTART, WAITING, UNKNOWN, setup, provision, test, allocate, reserve,
  cleanup]`: confirms all non-preserved states result in `waiting`.
- **`test_startup_comes_up_waiting_when_server_gives_no_state`** — parametrized
  over `[empty body, no state key, 404, ConnectionError]`: confirms that any
  failure to obtain a meaningful state from the server results in `waiting`
  (covering first-ever agent start and transient server unavailability).

### DIFF
--- a/agent/src/testflinger_agent/agent.py
+++ b/agent/src/testflinger_agent/agent.py
@@ -108,8 +108,8 @@ class TestflingerAgent:
             self.client, heartbeat_frequency=1
         )
         signal.signal(signal.SIGUSR1, self.restart_signal_handler)
-        self.set_agent_state(AgentState.WAITING)
         self._post_initial_agent_data()
+        self._set_startup_state()
         self.metrics_handler = PrometheusHandler(
             self.client.config.get("metrics_endpoint_port"), self.agent_id
         )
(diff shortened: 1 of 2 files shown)


agent/src/testflinger_agent/agent.py
```diff
@@ -134,6 +134,28 @@ def _post_initial_agent_data(self):
...
+        """
+        server_state, comment = self.get_agent_state()
+
+        if server_state == AgentState.OFFLINE:
+            logger.info(
+                "Agent was previously in state '%s' — preserving that state"
+                " on startup.",
+                server_state,
+            )
+            self.set_agent_state(server_state, comment)
```
rene-oromtz: IMO the logging is enough, since we (agent) are already offline, there is no need to send the same state + comment to the server. 
```suggestion
```

rene-oromtz (dismissed):
Just added one minor nitpick as the additional POST request is not needed. Besides that LGTM

rene-oromtz (approved):
LGTM

PR #1292, opened 2026-08-27
Refactor/use database module

ajzobro:
# CERTTF-1443: Route all direct MongoDB calls through the `database` module

## Summary

This is a pure refactor. No new functionality is introduced. The goal is to
enforce a clean separation of concerns: all MongoDB access lives in
`database.py`, and HTTP-layer code (`api/v1.py`, `views.py`) does not reach
into the database directly.

## Motivation

Before this change, `api/v1.py` and `views.py` both imported `mongo` and
called `mongo.db.*` directly alongside calls to `database.*` functions.
This made the database access boundary inconsistent and harder to reason
about, test, or swap out.

Two rules are enforced by this refactor:

1. All `mongo.db.*` calls must go through a function in `database.py`.
2. `database.py` must not reference HTTP status codes — that concern belongs
   to the HTTP layer.

### `src/testflinger/database.py`

Twenty new functions added to cover all previously uncovered call sites:

- **Jobs**: `get_job_data`, `search_jobs_by_pipeline`, `get_job_document`,
  `get_all_jobs_sorted`, `get_waiting_jobs_in_queue`, `cancel_job`
- **Queues**: `get_advertised_queues`, `upsert_queue`, `get_queue_images`,
  `set_queue_images`, `get_queue_document`,
  `get_advertised_queues_with_description`, `get_active_jobs_in_queue`
- **Agents**: `upsert_agent`, `get_all_agents`, `get_agent_document`,
  `get_all_agent_queue_names`, `update_agent_fields`,
  `get_agent_provision_streak`, `add_provision_log`

### `src/testflinger/api/v1.py`

All `database.mongo.db.*` call sites replaced with calls to the new
`database` functions. The `mongo` object is no longer accessed directly here.

### `src/testflinger/views.py`

All `mongo.db.*` call sites replaced with calls to the new `database`
functions. The `from testflinger.database import mongo` import removed.

### `tests/test_views.py`

Test patches updated from `testflinger.views.mongo` to
`testflinger.database.mongo`, reflecting where the database calls now live.

## What was not changed

- `logs.py`: `MongoLogHandler` receives `mongo` as a constructor argument
  and calls `mongo.db.logs` internally. It is the database-layer
  implementation for logs and is correctly instantiated via `database.mongo`
  at call sites. No change required.
- All existing `database.py` function signatures are preserved.
- No behaviour changes anywhere.


server/src/testflinger/database.py
```diff
@@ -765,6 +765,194 @@ def register_oidc_client(userinfo: dict) -> None:
     )
 
 
+def get_job_data(job_id: str) -> dict | None:
```
rene-oromtz: `get_job_data()` is a potential near-duplicate of `get_job()`

The only difference is the projection. This specifically only projects `job_data` and `submitted_by` but given `get_job()` projects this data already we could:

1. Drop this new function in favor of using `get_job()` (which btw is not used)
2. Drop `get_job()` in favor of this new function with this projection. If you chose this option, can you please also include `job_id` documented as a param in the docstring?

server/src/testflinger/database.py
```diff
@@ -765,6 +765,194 @@ def register_oidc_client(userinfo: dict) -> None:
...
+
+    :returns: Dict with job_data and submitted_by, or None if not found.
+    """
+    return mongo.db.jobs.find_one(
+        {"job_id": job_id},
+        projection={"job_data": True, "submitted_by": True, "_id": False},
+    )
+
+
+def search_jobs_by_pipeline(pipeline: list) -> list:
```
rene-oromtz: Nitpick on typehint. While `list` is also valid, there seems to be a lot of cases where we are documenting the type of the list so let's keep consistency
```suggestion
def search_jobs_by_pipeline(pipeline: list[dict]) -> list[dict]:
```

server/src/testflinger/database.py
```diff
@@ -765,6 +765,194 @@ def register_oidc_client(userinfo: dict) -> None:
...
+def get_advertised_queues() -> list[dict]:
+    """Return all advertised queues with name and description."""
+    return list(
+        mongo.db.queues.find(
+            {}, projection={"_id": False, "name": True, "description": True}
+        )
+    )
+
+
+def upsert_queue(name: str, description: str, timestamp) -> None:
```
rene-oromtz: nitpick on typehint:

```suggestion
def upsert_queue(name: str, description: str, timestamp: datetime) -> None:
```

Also, due to current design this is only for "advertised" queues so not sure if best to update the function to reflect this. I'm okay to leave it as-is but also want your opinion

server/src/testflinger/database.py
```diff
@@ -765,6 +765,194 @@ def register_oidc_client(userinfo: dict) -> None:
...
+    """
+    mongo.db.queues.update_one(
+        {"name": name},
+        {"$set": {"description": description, "updated_at": timestamp}},
+        upsert=True,
+    )
+
+
+def get_queue_images(queue: str) -> dict | None:
+    """Return the images dict for a given queue, or None if not found."""
```
rene-oromtz: This is missing the :param: and :returns: in the docstring, could you please add it for consistency with the other additions in this PR?

server/src/testflinger/database.py
```diff
@@ -765,6 +765,194 @@ def register_oidc_client(userinfo: dict) -> None:
...
+    :param image_data: Dict of image_name to provision_data mappings.
+    """
+    mongo.db.queues.update_one(
+        {"name": queue},
+        {"$set": {"images": image_data}},
+        upsert=True,
+    )
+
+
+def upsert_agent(agent_name: str, data: dict, log: list) -> None:
```
rene-oromtz: Nitpick on typehint. Same justification to keep consistency as least for `list`

```suggestion
def upsert_agent(agent_name: str, data: dict, log: list[str]) -> None:
```

Also, a rename can be made, perhaps to `upsert_agent_record` or `upsert_agent_document` to indicate more clearly we are upserting a document or for consistency with other functions but I'm also okay to keep as-is

server/src/testflinger/database.py
```diff
@@ -765,6 +765,194 @@ def register_oidc_client(userinfo: dict) -> None:
...
+    :param log: Log lines to push (kept to last 100).
+    """
+    mongo.db.agents.update_one(
+        {"name": agent_name},
+        {"$set": data, "$push": {"log": {"$each": log, "$slice": -100}}},
+        upsert=True,
+    )
+
+
+def add_provision_log(agent_name: str, data: dict) -> None:
```
rene-oromtz: Unused function, consider removing if not intended to be used in code

server/src/testflinger/database.py
```diff
@@ -765,6 +765,194 @@ def register_oidc_client(userinfo: dict) -> None:
...
+            "$set": data,
+            "$push": {
+                "provision_log": {"$each": [data], "$slice": -100},
+            },
+        },
+        upsert=True,
+    )
+
+
+def get_agent_provision_streak(agent_name: str) -> dict | None:
```
rene-oromtz: Unused function, consider removing if not intended to be used in code

server/src/testflinger/database.py
```diff
@@ -765,6 +765,194 @@ def register_oidc_client(userinfo: dict) -> None:
...
+    :returns: Dict with provision_streak_type and provision_streak_count,
+        or None if the agent does not exist.
+    """
+    return mongo.db.agents.find_one(
+        {"name": agent_name},
+        {"provision_streak_type": 1, "provision_streak_count": 1},
+    )
+
+
+def update_agent_fields(agent_name: str, fields: dict) -> None:
```
rene-oromtz: Unused function, consider removing if not intended to be used in code

server/src/testflinger/database.py
```diff
@@ -765,6 +765,194 @@ def register_oidc_client(userinfo: dict) -> None:
...
+    return list(
+        mongo.db.jobs.find(
+            {"job_data.job_queue": queue, "result_data.job_state": "waiting"},
+            {"job_id": 1},
+            sort=[("job_priority", -1)],
+        )
+    )
+
+
+def get_agent_document(agent_id: str) -> dict | None:
```
rene-oromtz: Unused function, consider removing if not intended to be used in code

server/src/testflinger/database.py
```diff
@@ -765,6 +765,194 @@ def register_oidc_client(userinfo: dict) -> None:
...
+
+    :param queue_name: Name of the queue.
+    :returns: Queue document or None if not found.
+    """
+    return mongo.db.queues.find_one({"name": queue_name})
+
+
+def get_all_jobs_sorted() -> list[dict]:
+    """Return all job documents sorted by created_at descending."""
+    return list(mongo.db.jobs.find(sort=[("created_at", -1)]))
```
rene-oromtz, line 868: Minor nitpick just to add the `:returns:` for consistency, the docstring is self documenting this return so if you prefer to keep as-is I could accept that for justification

server/src/testflinger/database.py
```diff
@@ -765,6 +765,194 @@ def register_oidc_client(userinfo: dict) -> None:
...
+
+
+def get_job_document(job_id: str) -> dict | None:
```
rene-oromtz, line 871: This is also potentially redundant given we have `get_job()` which does not project the `_id` (which should be preferred as we don't need it). 

We also have `get_job_data()` which does another projection on the same collection

rene-oromtz (commented):
There seems to be some near-duplicates and/or unused functions that could be removed or reworked. I also added some other changes for typehint/doctrings consistency

server/src/testflinger/database.py
ajzobro: - get_job_data() returns only job_data and submitted_by.
 - get_job() *not used* returns the complete job document, including result_data, but logs are stored separately in the logs collection and are not included.

Actions: delete `get_job`, update docstring for `get_job_data` to include `job_id` as a param.

server/src/testflinger/database.py
```diff
@@ -765,6 +765,194 @@ def register_oidc_client(userinfo: dict) -> None:
...
+    """Return the provision streak fields for a given agent.
+
+    :param agent_name: Name of the agent.
+    :returns: Dict with provision_streak_type and provision_streak_count,
+        or None if the agent does not exist.
+    """
+    return mongo.db.agents.find_one(
+        {"name": agent_name},
+        {"provision_streak_type": 1, "provision_streak_count": 1},
+    )
```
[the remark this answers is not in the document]
ajzobro: OBE

server/src/testflinger/database.py
ajzobro: After investigating, results data was determined not to have any logs (hence it's probably fairly small on average), I have removed get_job_data as a dupe and retained `get_job_document`

rene-oromtz (approved):
LGTM, thanks for addressing the comments

PR #1293, opened 2026-08-27
add capture of status change time

ajzobro:
# Track job state transition timestamps

## Summary

Adds a `job_state_changed_at` field to job result data that records the UTC timestamp whenever a job's state changes. This enables consumers of the API to determine exactly when a job transitioned between states (e.g. `waiting` → `running` → `cancelled`).

### API schema (`server/schemas/openapi.json`, `server/src/testflinger/api/schemas.py`)

- Added `job_state_changed_at` as an optional string field to the `ResultGet` schema and the OpenAPI spec.

### Job creation (`server/src/testflinger/api/v1.py`)

- `job_builder`: captures a single `now` timestamp and uses it for both `created_at` and the initial `job_state_changed_at` (set to `"waiting"`), avoiding a small time skew between the two fields.
- `cancel_job`: the `$set` update for cancelling a job now also writes `job_state_changed_at` with the current UTC time.

### Job dispatch (`server/src/testflinger/database.py`)

- `pop_job`: when an agent picks up a job and the state transitions to `"running"`, `job_state_changed_at` is recorded atomically in the same `find_one_and_update` call.
- `add_job_results`: when a result POST includes a `job_state` key, `job_state_changed_at` is injected automatically before the keys are prefixed with `result_data.`. If `job_state` is absent the timestamp is left unchanged.

## Tests (`server/tests/test_v1.py`)

Five new test cases cover the full lifecycle:

| Test | What it verifies |
|---|---|
| `test_initial_job_state_changed_at` | Field is present and a valid timezone-aware ISO 8601 string immediately after job creation (`waiting`). |
| `test_job_state_changed_at_on_result_post` | Field is updated (and monotonically non-decreasing) when a new `job_state` is POSTed to `/v1/result/{job_id}`. |
| `test_job_state_changed_at_not_updated_without_state` | Field is **not** changed when a result POST contains no `job_state`. |
| `test_job_state_changed_at_on_pop_job` | Field is set when an agent pops a job (transition to `running`). |
| `test_job_state_changed_at_on_cancel` | Field is set when a job is cancelled via the action endpoint. |

## Notes

- All timestamps are UTC ISO 8601 strings (produced via `datetime.now(timezone.utc).isoformat()`), consistent with the existing `created_at` field.
- The field is optional in the schema to remain backwards-compatible with existing result documents that pre-date this change.


server/src/testflinger/api/schemas.py
```diff
@@ -490,6 +490,7 @@ class ResultGet(ResultStatus):
 
     device_info = fields.Dict(required=False)
     job_state = fields.String(required=False)
+    job_state_changed_at = fields.String(required=False)
```
rene-oromtz: This should be a Marshmallow [Datetime](https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.DateTime) which defaults to ISO8601
```suggestion
    job_state_changed_at = fields.Datetime(required=False)
```

server/src/testflinger/api/v1.py
```diff
@@ -194,10 +194,12 @@ def has_attachments(data: dict) -> bool:
 
 def job_builder(data: dict) -> dict:
     """Build a job from a dictionary of data."""
+    now = datetime.now(timezone.utc)
     job = {
-        "created_at": datetime.now(timezone.utc),
+        "created_at": now,
         "result_data": {
             "job_state": "waiting",
+            "job_state_changed_at": now.isoformat(),
```
rene-oromtz: There is no need to convert to isoformat whenever storing in the database. MongoDB stores natively a datetime as a BSON datetime which should be prefer over a string. For API usage, the serialization needs to be covered by the [schema](https://github.com/canonical/testflinger/pull/1293/changes#r3898312325)

server/src/testflinger/database.py
```diff
@@ -229,11 +229,13 @@ def pop_job(queue_list: list[str], agent_name: str) -> dict | None:
             "job_data.exclude_agents": {"$nin": [agent_name]},
         }
 
+        running_at = datetime.now(timezone.utc)
         response = mongo.db.jobs.find_one_and_update(
             query_filter,
             {
                 "$set": {
                     "result_data.job_state": "running",
+                    "result_data.job_state_changed_at": running_at.isoformat(),
```
rene-oromtz: Same [comment](https://github.com/canonical/testflinger/pull/1293/changes#r3898344807) about iso format but given `running_at` is only used once, we can perhaps drop the variable and use it directly

rene-oromtz (changes requested):
Thanks for the changes. Aside from Copilot's comment, I added a note about storing in MongoDB as string (via `isoformat`) vs BSON Datetime (via `datetime` object). We should prefer the latter as storage-wise this is also 4 times less storage 32 bytes vs 8 bytes) and other [features](https://stackoverflow.com/a/29565489) that we may miss if we store as strings. (I did not add the comment to all date conversion but the same comment applies for all)

server/src/testflinger/database.py
```diff
@@ -784,16 +788,34 @@ def get_job(job_id: str) -> dict | None:
...
     """Add results to specified job id with "result_data" prepended."""
+    has_job_state = "job_state" in json_data
+    job_state = json_data.pop("job_state", None)
+    json_data.pop("job_state_changed_at", None)
     # First, we need to prepend "result_data" to each key in the result_data
     for key in list(json_data):
         json_data[f"result_data.{key}"] = json_data.pop(key)
 
     mongo.db.jobs.update_one({"job_id": job_id}, {"$set": json_data})
+    if has_job_state:
```
rene-oromtz: I would only suggest we update the document in case the job_state is also not None:

```suggestion
    if has_job_state and job_state is not None:
```

Perhaps just defensive but we are currently defaulting to None.

rene-oromtz (dismissed):
Just added a minor nitpick besides that LGTM

server/src/testflinger/database.py
ajzobro: In latest commit:
```
    # ResultPost validates job_state as a String (or absent), so popping
    # with a None default cleanly distinguishes "no transition" (None)
    # from "transition to <state>" (any string).
    job_state = json_data.pop("job_state", None)
...
    if job_state is not None:
```

server/src/testflinger/api/schemas.py
```diff
@@ -490,12 +490,24 @@ class ResultGet(ResultStatus):
...
+        """Configure marshmallow to reject unknown fields.
+
+        Explicit ``unknown=RAISE`` (matches marshmallow's default) locks the
+        contract so clients cannot smuggle server-managed fields like
+        ``job_state_changed_at``. See ``add_job_results`` for the
+        defense-in-depth strip.
+        """
+
+        unknown = RAISE
+
```
rene-oromtz, line 510: IMO, this explicitness is not needed as this is already the default behavior as indicated in the docstrings. The current ResultPost schema is already self documented and explicit on the dict values that are accepted (`status`, `agent_id`, `device_info`, `job_state`) so there is not a possibility to smuggle `job_state_changed_at`)

server/src/testflinger/database.py
```diff
@@ -784,16 +790,60 @@ def get_job(job_id: str) -> dict | None:
 
 def add_job_results(job_id: str, json_data: dict):
     """Add results to specified job id with "result_data" prepended."""
-    # First, we need to prepend "result_data" to each key in the result_data
-    for key in list(json_data):
-        json_data[f"result_data.{key}"] = json_data.pop(key)
+    # Never allow a client to set the change timestamp directly.
+    json_data.pop("job_state_changed_at", None)
```
rene-oromtz: The schema already rejects this, this check is redundant.

server/src/testflinger/database.py
```diff
@@ -784,16 +790,60 @@ def get_job(job_id: str) -> dict | None:
...
+    # ``ResultPost`` validates ``job_state`` as a String (or absent) at the
+    # API boundary, but ``add_job_results`` is also called directly from
+    # other server code paths. Re-validate here so a non-string ``job_state``
+    # can never reach Mongo, where it would poison the $ne comparison and
+    # every downstream reader of ``result_data.job_state``.
+    job_state = json_data.pop("job_state", None)
+    if job_state is not None and not isinstance(job_state, str):
+        raise TypeError(
+            f"job_state must be a string, got {type(job_state).__name__}"
+        )
```
rene-oromtz: note referring code validation:
The type check is redundant as it should be rejected at the schema level instead of checking it here. 
Additionally we are already checking is not None while attempting to write to the DB (line 812), consider to remove this section

server/tests/test_v1.py
```diff
@@ -1953,6 +1983,259 @@ def test_job_get_no_auth_headers(mongo_app):
...
+    # Timestamp must remain unchanged (no state transition).
+    assert (
+        updated_result["job_state_changed_at"]
+        == initial_result["job_state_changed_at"]
+    )
+    # But the sibling result_data field must have been written.
+    assert updated_result.get("device_info") == {"serial": "abc123"}
+
+
+def test_client_supplied_job_state_changed_at_is_ignored(
```
rene-oromtz: Consider reworking this test as this is handled at the schema level so potentially, its only up to this assertion:
`assert rejected.status_code == HTTPStatus.UNPROCESSABLE_ENTITY`

Additionally, the attempted spoofed value by directly calling `database.add_job_results` is never used, the write to the DB never uses `job_state_changed_at` value (even if you remove the `pop()` the test will continue to pass)

server/src/testflinger/database.py
rene-oromtz, line 794: The schema already rejects this, this check is redundant.

server/src/testflinger/database.py
rene-oromtz, line 805: The type check is redundant as it should be rejected at the schema level instead of checking it here. 
Additionally we are already checking is not None while attempting to write to the DB (line 812), consider removing this section

server/tests/test_v1.py
rene-oromtz, line 2116: Consider reworking this test as this is handled at the schema level so potentially, its only up to this assertion:
`assert rejected.status_code == HTTPStatus.UNPROCESSABLE_ENTITY`

Additionally, the attempted spoofed value by directly calling `database.add_job_results` is in any case overwritten by `atomic_set` so server value always win

server/src/testflinger/database.py
rene-oromtz: Additionally, its always overwritten by server side time before making the write:
```
atomic_set = {
            **set_fields,
            "result_data.job_state": job_state,
            "result_data.job_state_changed_at": datetime.now(timezone.utc),
        }
```
so its never updated client side (even if for some reason it skips the Schema check)

rene-oromtz (changes requested):
Added some comment that could overall simplify the proposed solution.

server/src/testflinger/api/schemas.py
ajzobro: The fact this is default is exactly why you're able to say that the current schema is explicit, no?

Relying on a default that we do not control means we have to be cognizant of any changes to the libraries defaults or we would become exposed to additional input fields being allowed in.

I am generally not a fan of relying on someone else's defaults, so "it's already the default" isn't going to sway me.

If you point to elsewhere in our code where this is setup, I think I would buy into it a little more. Similarly if we have a test that attempts to submit fields that aren't in the schema, thereby proving the expected behavior does in fact raise, I will stand-down.

As far as I can tell: There is no test asserting that an arbitrary unknown key (e.g. `{"totally_made_up_field": 1}`) is rejected.

server/src/testflinger/database.py
ajzobro: I've made updates that I'm comfortable with, but I don't think "we don't need to check" is the right answer, not all code paths go through the schema.

The schema level-validation is for the external schema; we still need to validate data which is being written to the database as it does not ALL come from external sources subject to schema validation.

server/tests/test_v1.py
ajzobro: I caught up.

ajzobro: Here's the thing; we CAN play the game of "look at where all the functions are used and only protect them if we need to" -- but then in the future we may get caught entirely off guard when those functions start getting used in a new way.

I am of the opinion that the `database.py` module should be wholly self-protected and not rely on an external schema (completely unassociated with an internal database function) as if it were protecting the data that we are writing.

If we are making a dedicated `database.py` it should robustly guarantee the data it writes.

server/src/testflinger/api/schemas.py
ajzobro: Thus we are completely exposed if the default that we don't control changes.

rene-oromtz: Only one I could find is `test_add_job_invalid_reserve_data()`
```
    # Invalid additional fields
    job_data = {
        "job_queue": "test",
        "reserve_data": {"invalid_field": "what am I doing here"},
    }
    app, _ = mongo_app
    output = app.post("/v1/job", json=job_data)
    assert 422 == output.status_code
```

I'm not saying we should not add a test to validate unknown fields (in fact you also have one already that checks the `UnprocessableEntity`) but perhaps the explicitness here is not needed due to the nature of a schema (which should by default reject any not know field regardless of the library used). 

If you prefer to keep the explicitness I can live with that but can you please update the docstrings given the defense-in-depth is no-op? (server side always wins anyway)

server/schemas/openapi.json
```diff
@@ -2753,7 +2761,7 @@
     },
     "/v1/result/{job_id}/status": {
       "get": {
-        "description": "This is a lightweight alternative to GET /result/<job_id> that omits\nlog data (output and serial).  Use this when only the job state or\nphase statuses are needed.\n\n:param job_id: UUID as a string for the job\n:raises HTTPError: If the job_id is not a valid UUID",
+        "description": "This is a lightweight alternative to GET /result/<job_id> that omits\nlog data (output and serial). Use this when only the job state,\nits last change time, or phase statuses are needed. The optional\njob_state_changed_at timestamp uses UTC by convention and is absent\nfor older jobs that have not recorded a state change.\n\n:param job_id: UUID as a string for the job\n:raises HTTPError: If the job_id is not a valid UUID",
```
ajzobro, line 2764:
```suggestion
        "description": "This is a lightweight alternative to GET /result/<job_id> that omits\nlog data (output and serial). Use this when only the job state,\nits last change time, or phase statuses are needed.\n\n:param job_id: UUID as a string for the job\n:raises HTTPError: If the job_id is not a valid UUID",
```

server/src/testflinger/api/schemas.py
```diff
@@ -496,6 +497,17 @@ class ResultGet(ResultStatus):
...
+        """Configure marshmallow to reject unknown fields.
+
+        Explicit ``unknown=RAISE`` (matches marshmallow's default) locks the
+        contract so clients cannot smuggle server-managed fields like
+        ``job_state_changed_at``. See ``update_job_results`` for the
+        defense-in-depth strip.
+        """
+
+        unknown = RAISE
+
```
ajzobro, line 510: I know this is the default.
I know that it's not wrong.
I know it would work without this code.
Adding this code is not detrimental and it locks in the behavior without relying on the library defaults for the behavior we are expecting in our product.

agent/src/testflinger_agent/client.py
```diff
@@ -349,11 +349,7 @@ def transmit_job_outcome(self, rundir: Path):
             logger.info("Submitting job outcome for job: %s", job_id)
             with outcome_file.open() as f:
                 data = json.load(f)
-                # Only include status in posted results
-                # TODO: Remove pop once backward compatibility is not needed
-                data.pop("output", None)
-                data.pop("serial", None)
-                data["job_state"] = "complete"
+                data["job_state"] = "completed"
```
rene-oromtz: Please note that this is breaking for the `cli` (specially during polling) which is only looking for `complete`. As per our discussion, this change can perhaps wait until we land all agent status changes and focus this PR on the `job_state_changed_at` field

server/src/testflinger/api/schemas.py
rene-oromtz: Okay, we can leave as-is if that is your preference

rene-oromtz (changes requested):
Just added one note with regards of the agent "completed" which I think its out of scope for this PR and its potentially breaking. Also consider that schema changes need to go through the `schema` recipe as currently the unit tests are failing. 

Besides that I'm marking all comments as resolved and we can merge as-is

agent/src/testflinger_agent/client.py
ajzobro: Will revert, CLI is messy.

rene-oromtz (approved):

PR #1294, opened 2026-08-28
fix(server): fix pyyaml multiline display

rene-oromtz:
## Description
Fixes a known issue with PyYAML where in case there is any trailing whitespace in a multiline string. 

This is using the exact same approach as the CLI:
https://github.com/canonical/testflinger/blob/2ac872bee342203c55b2981bed687336694dca85/cli/testflinger_cli/helpers.py#L167-L174

## Resolved issues
N/A

## Documentation
N/A

## Web service API changes
N/A

## Tests
Tested locally with a very simple test command:
```
test_data:
  test_cmds: |
    echo "hello" 
    echo "lsb_release -a" 
```

Without fix:
Screenshot from 2026-08-28 15-18-28

With fix:
Screenshot from 2026-08-28 15-18-40

### DIFF
--- a/server/src/testflinger/views.py
+++ b/server/src/testflinger/views.py
@@ -70,8 +70,14 @@ class _JobYamlDumper(yaml.SafeDumper):
 
 def _str_representer(dumper, data):
     """Represent multiline strings (e.g. test_cmds) as literal blocks."""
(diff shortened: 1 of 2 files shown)


ajzobro (approved):
LGTM

PR #1297, opened 2026-09-01
feat(server): add backend changes to store job events

rene-oromtz:
## Description
This is the foundation PR for adding job events to be retrievable by the API and visible in the UI. 

This adds an `events` module where we can:
1. format an event enum to a user friendly message containing additional context 
2. build an event with an generic return structure (that can also be leveraged by follow-up `AgentEvent`)
3. Detects an incomming POST request to determine which event should be logged

The changes are server side only as it determines the job/phase state based on the POST request received from an agent.
Each event was added in the relevant endpoint where we need to log the job event.

## Resolved issues
Merging the full stack resolves [CERTTF-1448](https://warthogs.atlassian.net/browse/CERTTF-1448)

## Documentation
N/A

## Web service API changes
No new endpoint in this PR

## Tests
Added unit tests for the code added in this PR. Follow-up end to end testing is made on subsequent PRs

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

[CERTTF-1448]: https://warthogs.atlassian.net/browse/CERTTF-1448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/src/testflinger/api/v1.py
+++ b/server/src/testflinger/api/v1.py
@@ -28,11 +28,11 @@ from flask import current_app, g, jsonify, request, send_file
 from marshmallow import ValidationError
 from prometheus_client import Counter
 from requests.adapters import HTTPAdapter
-from testflinger_common.enums import LogType, ServerRoles, TestPhase
+from testflinger_common.enums import JobEvent, LogType, ServerRoles, TestPhase
 from urllib3.util.retry import Retry
 from werkzeug.routing import BaseConverter
 
-from testflinger import database
+from testflinger import database, events
 from testflinger.api import auth, helpers, schemas
 from testflinger.api.auth import authenticate, require_role
 from testflinger.logs import LogFragment, MongoLogHandler
@@ -140,6 +140,16 @@ def job_post(json_data: dict) -> dict:
     # CAUTION! If you ever move this line, you may need to pass data as a copy
     # because it will get modified by submit_job and other things it calls
     database.add_job(job)
+
+    database.add_job_event(
+        job_id=job["job_id"],
+        event=events.build_event(
+            event_type=JobEvent.JOB_SUBMITTED,
+            timestamp=datetime.now(timezone.utc),
+            client_id=g.client_id,
+            queue_name=job["job_data"]["job_queue"],
+        ),
+    )
     return jsonify(job_id=job.get("job_id"))
 
 
@@ -261,6 +271,15 @@ def job_get():
     job = database.pop_job(queue_list=queue_list, agent_name=agent_name)
     if not job:
         return jsonify({}), HTTPStatus.NO_CONTENT
+
+    database.add_job_event(
+        job_id=job["job_id"],
+        event=events.build_event(
+            event_type=JobEvent.JOB_ASSIGNED,
+            timestamp=datetime.now(timezone.utc),
+            agent_name=agent_name,
+        ),
+    )
     if (secrets := retrieve_secrets(job)) is not None:
         job["test_data"]["secrets"] = secrets
     database.set_agent_job(agent_name, job["job_id"])
@@ -587,7 +606,14 @@ def result_post(job_id: str, json_data: dict) -> str:
     if content_length and content_length >= 16 * 1024 * 1024:
         abort(HTTPStatus.REQUEST_ENTITY_TOO_LARGE, message="Payload too large")
 
+    # We need to get new events before json_data is mutated by add_job_results
+    previous_results = database.get_job_results(job_id) or {}
+    previous_data = previous_results.get("result_data", {})
+    new_events = events.detect_new_result_events(previous_data, json_data)
+
     database.add_job_results(job_id, json_data)
+    for event in new_events:
+        database.add_job_event(job_id=job_id, event=event)
     return "OK"
 
 
@@ -686,6 +712,16 @@ def _cancel_job(job_id):
             "The job is already completed or cancelled",
             HTTPStatus.BAD_REQUEST,
         )
+
+    # Only add an event if the cancellation is actually made
+    database.add_job_event(
+        job_id=job_id,
+        event=events.build_event(
+            event_type=JobEvent.JOB_CANCELLED,
+            timestamp=datetime.now(timezone.utc),
+            client_id=g.client_id,
+        ),
+    )
     return "OK"
 
 
--- a/server/src/testflinger/api/v1.py
+++ b/server/src/testflinger/api/v1.py
@@ -145,7 +145,6 @@ def job_post(json_data: dict) -> dict:
         job_id=job["job_id"],
         event=events.build_event(
             event_type=JobEvent.JOB_SUBMITTED,
-            timestamp=datetime.now(timezone.utc),
             client_id=g.client_id,
             queue_name=job["job_data"]["job_queue"],
         ),
@@ -276,7 +275,6 @@ def job_get():
         job_id=job["job_id"],
         event=events.build_event(
             event_type=JobEvent.JOB_ASSIGNED,
-            timestamp=datetime.now(timezone.utc),
             agent_name=agent_name,
         ),
     )
@@ -607,8 +605,9 @@ def result_post(job_id: str, json_data: dict) -> str:
         abort(HTTPStatus.REQUEST_ENTITY_TOO_LARGE, message="Payload too large")
 
     # We need to get new events before json_data is mutated by add_job_results
-    previous_results = database.get_job_results(job_id) or {}
-    previous_data = previous_results.get("result_data", {})
+    previous_data = (database.get_job_results(job_id) or {}).get(
+        "result_data", {}
+    )
     new_events = events.detect_new_result_events(previous_data, json_data)
 
     database.add_job_results(job_id, json_data)
@@ -718,7 +717,6 @@ def _cancel_job(job_id):
         job_id=job_id,
         event=events.build_event(
             event_type=JobEvent.JOB_CANCELLED,
-            timestamp=datetime.now(timezone.utc),
             client_id=g.client_id,
         ),
     )
--- a/server/src/testflinger/api/v1.py
+++ b/server/src/testflinger/api/v1.py
@@ -595,7 +595,7 @@ def result_post(job_id: str, json_data: dict) -> str:
     :param job_id: UUID as a string for the job
     :raises HTTPError: If the job_id is not a valid UUID
     """
-    if not check_valid_uuid(job_id):
+    if not check_valid_uuid(job_id) or not database.job_exists(job_id):
         abort(HTTPStatus.BAD_REQUEST, message="Invalid job_id specified")
 
     # fail if input payload is larger than the BSON size limit
@@ -604,13 +604,11 @@ def result_post(job_id: str, json_data: dict) -> str:
     if content_length and content_length >= 16 * 1024 * 1024:
         abort(HTTPStatus.REQUEST_ENTITY_TOO_LARGE, message="Payload too large")
 
-    # We need to get new events before json_data is mutated by add_job_results
-    previous_data = (database.get_job_results(job_id) or {}).get(
-        "result_data", {}
-    )
+    # update_job_results atomically returns the previous result_data, so
+    # event detection is based on the exact state this update transitioned
+    # from
+    previous_data = database.update_job_results(job_id, json_data)
     new_events = events.detect_new_result_events(previous_data, json_data)
-
-    database.add_job_results(job_id, json_data)
     for event in new_events:
         database.add_job_event(job_id=job_id, event=event)
     return "OK"
--- a/server/src/testflinger/database.py
+++ b/server/src/testflinger/database.py
@@ -139,7 +139,7 @@ def create_indexes():
     mongo.db.agents.create_index("name", unique=True)
     mongo.db.client_permissions.create_index("client_id", unique=True)
     mongo.db.client_permissions.create_index("sub", sparse=True)
-    mongo.db.jobs.create_index("job_id")
+    mongo.db.jobs.create_index("job_id", unique=True)
     mongo.db.jobs_events.create_index("job_id", unique=True)
     mongo.db.jobs.create_index(["result_data.job_state", "job_data.job_queue"])
     mongo.db.agents.create_index("queues")
--- a/server/src/testflinger/database.py
+++ b/server/src/testflinger/database.py
@@ -128,6 +128,11 @@ def create_indexes():
         partialFilterExpression={"sub": {"$exists": True}},
     )
 
+    # Remove stale job events after defined expiration
+    mongo.db.jobs_events.create_index(
+        "updated_at", expireAfterSeconds=DEFAULT_EXPIRATION
+    )
+
     # Faster lookups for common queries
     mongo.db.refresh_tokens.create_index("refresh_token", unique=True)
     mongo.db.refresh_tokens.create_index("client_id")
@@ -135,6 +140,7 @@ def create_indexes():
     mongo.db.client_permissions.create_index("client_id", unique=True)
     mongo.db.client_permissions.create_index("sub", sparse=True)
     mongo.db.jobs.create_index("job_id")
+    mongo.db.jobs_events.create_index("job_id", unique=True)
     mongo.db.jobs.create_index(["result_data.job_state", "job_data.job_queue"])
     mongo.db.agents.create_index("queues")
 
@@ -1032,3 +1038,24 @@ def update_agent_provision_log(agent_name, json_data) -> bool:
         mongo.db.agents.update_one({"name": agent_name}, {"$set": agent})
         found = True
     return found
+
+
+def add_job_event(job_id: str, event: dict) -> None:
+    """Add an event to the job events collection.
+
+    :param job_id: The ID of the job.
+    :param event: The event data to add.
+    """
+    mongo.db.jobs_events.update_one(
+        {"job_id": job_id},
+        {
+            "$push": {
+                "events": {
+                    "$each": [event],
+                    "$sort": {"timestamp": -1},
+                }
+            },
+            "$set": {"updated_at": datetime.now(timezone.utc)},
+        },
+        upsert=True,
+    )
--- a/server/src/testflinger/database.py
+++ b/server/src/testflinger/database.py
@@ -22,6 +22,7 @@ from typing import Any
 
 from flask_pymongo import PyMongo
 from gridfs import GridFS, errors
+from pymongo import ReturnDocument
 from testflinger_common.enums import ServerRoles
 
 # Constants for TTL indexes
@@ -919,19 +920,38 @@ def get_job_results(job_id: str):
     )
 
 
-def add_job_results(job_id: str, json_data: dict):
-    """Add results to specified job id with "result_data" prepended."""
-    # First, we need to prepend "result_data" to each key in the result_data
-    for key in list(json_data):
-        json_data[f"result_data.{key}"] = json_data.pop(key)
+def update_job_results(job_id: str, json_data: dict) -> dict:
+    """Update results for a job and return the previous `result_data`.
 
-    mongo.db.jobs.update_one({"job_id": job_id}, {"$set": json_data})
+    The previous `result_data` is returned to reliable detect any
+    state transitions this document update introduced.
+
+    :param job_id: The job ID to update results for.
+    :param json_data: The result data to store (not modified).
+    :return: The previous `result_data` dict, or an empty dict
+    """
+    # Prepend "result_data" to each key in the result data
+    set_data = {
+        f"result_data.{key}": value for key, value in json_data.items()
+    }
+
+    # find_one_and_update guarantees atomicity as it locks the document
+    # for the duration of the update. Returning the previous document
+    # allows us to properly detect unique events
+    previous_doc = mongo.db.jobs.find_one_and_update(
+        {"job_id": job_id},
+        {"$set": set_data},
+        projection={"result_data": True, "_id": False},
+        return_document=ReturnDocument.BEFORE,
+    )
 
     # Additionally, because the job_data may reflect that the job is now done,
     # we need to disassociate the agent from the job if the job is done:
     terminal_states = {"complete", "completed", "cancelled"}
-    if json_data.get("result_data.job_state") in terminal_states:
+    if json_data.get("job_state") in terminal_states:
         clear_agent_job(job_id)
+
+    return (previous_doc or {}).get("result_data", {})
 
 
 def job_exists(job_id: str) -> bool:
--- a/server/src/testflinger/database.py
+++ b/server/src/testflinger/database.py
@@ -1043,18 +1043,15 @@ def update_agent_provision_log(agent_name, json_data) -> bool:
 def add_job_event(job_id: str, event: dict) -> None:
     """Add an event to the job events collection.
 
+    Events are appended in natural insertion order (oldest first).
+
     :param job_id: The ID of the job.
     :param event: The event data to add.
     """
     mongo.db.jobs_events.update_one(
         {"job_id": job_id},
         {
-            "$push": {
-                "events": {
-                    "$each": [event],
-                    "$sort": {"timestamp": -1},
-                }
-            },
+            "$push": {"events": event},
             "$set": {"updated_at": datetime.now(timezone.utc)},
         },
         upsert=True,
--- a/server/src/testflinger/database.py
+++ b/server/src/testflinger/database.py
@@ -139,7 +139,7 @@ def create_indexes():
     mongo.db.agents.create_index("name", unique=True)
     mongo.db.client_permissions.create_index("client_id", unique=True)
     mongo.db.client_permissions.create_index("sub", sparse=True)
-    mongo.db.jobs.create_index("job_id", unique=True)
+    mongo.db.jobs.create_index("job_id")
     mongo.db.jobs_events.create_index("job_id", unique=True)
     mongo.db.jobs.create_index(["result_data.job_state", "job_data.job_queue"])
     mongo.db.agents.create_index("queues")
--- a/server/src/testflinger/events.py
+++ b/server/src/testflinger/events.py
@@ -132,7 +132,6 @@ def _build_job_lifecycle_events(
 
     :param previous_data: The previous results data.
     :param new_data: The new results data.
-    :param timestamp: The timestamp of the event.
     :return: new job lifecycle events or empty list if no new events.
     """
     # Early return if there is no job_state reported yet
--- a/server/src/testflinger/events.py
+++ b/server/src/testflinger/events.py
@@ -1,0 +1,182 @@
+# Copyright (C) 2026 Canonical
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""Event handling functions for Testflinger server."""
+
+from datetime import datetime, timezone
+
+from testflinger_common.enums import (
+    JobEvent,
+    JobState,
+    TestflingerEvent,
+    TestPhase,
+)
+
+_MESSAGE_TEMPLATES = {
+    JobEvent.JOB_SUBMITTED: "Job submitted by user {client_id} for queue {queue_name}.",  # noqa: E501
+    JobEvent.JOB_ASSIGNED: "Job assigned to agent {agent_name}.",
+    JobEvent.JOB_STARTED: "Job started",
+    JobEvent.JOB_PHASE_STARTED: "Phase {phase} started.",
+    JobEvent.JOB_PHASE_COMPLETED: "Phase {phase} completed with exit code {status}",  # noqa: E501
+    JobEvent.JOB_COMPLETED: "Job completed.",
+    JobEvent.JOB_CANCELLED: "Job cancellation requested by user {client_id}.",
+}
+
+
+class _DefaultContext(dict):
+    """Render missing placeholders instead of raising KeyError."""
+
+    def __missing__(self, key):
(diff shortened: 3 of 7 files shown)


server/src/testflinger/api/v1.py
```diff
@@ -590,7 +609,14 @@ def result_post(job_id: str, json_data: dict) -> str:
     if content_length and content_length >= 16 * 1024 * 1024:
         abort(HTTPStatus.REQUEST_ENTITY_TOO_LARGE, message="Payload too large")
 
+    # We need to get new events before json_data is mutated by add_job_results
+    previous_results = database.get_job_results(job_id) or {}
+    previous_data = previous_results.get("result_data", {})
```
ajzobro: could be a one-liner
```suggestion
    previous_data = (database.get_job_results(job_id) or {}).get("result_data", {})
```

Might need a shorter variable name given the 79 char limit.

Would help with understanding "_results" (data containing result_data but not result_data) vs "_data" (result_data)

server/src/testflinger/api/v1.py
```diff
@@ -689,6 +715,16 @@ def _cancel_job(job_id):
             "The job is already completed or cancelled",
             HTTPStatus.BAD_REQUEST,
         )
+
+    # Only add an event if the cancellation is actually made
+    database.add_job_event(
+        job_id=job_id,
+        event=events.build_event(
+            event_type=JobEvent.JOB_CANCELLED,
+            timestamp=datetime.now(timezone.utc),
```
ajzobro: is there a reason not to just make this the default?

Are there cases where we intentionally don't use "now"?

server/src/testflinger/database.py
```diff
@@ -128,13 +128,19 @@ def create_indexes():
...
+    )
+
     # Faster lookups for common queries
     mongo.db.refresh_tokens.create_index("refresh_token", unique=True)
     mongo.db.refresh_tokens.create_index("client_id")
     mongo.db.agents.create_index("name", unique=True)
     mongo.db.client_permissions.create_index("client_id", unique=True)
     mongo.db.client_permissions.create_index("sub", sparse=True)
     mongo.db.jobs.create_index("job_id")
+    mongo.db.jobs_events.create_index("job_id", unique=True)
```
ajzobro, line 144: even the jobs document doesn't enforce a unique value -- probably should.

server/src/testflinger/database.py
```diff
@@ -901,3 +907,21 @@ def update_agent_provision_log(agent_name, json_data) -> bool:
...
+    Events are appended in natural insertion order (oldest first).
+
+    :param job_id: The ID of the job.
+    :param event: The event data to add.
+    """
+    mongo.db.jobs_events.update_one(
+        {"job_id": job_id},
+        {
+            "$push": {"events": event},
+            "$set": {"updated_at": datetime.now(timezone.utc)},
```
ajzobro, line 1075: It seems like we may not be storing the event time for each event (ability to measure between events)

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+from datetime import datetime, timezone
+
+from testflinger_common.enums import (
+    JobEvent,
+    JobState,
+    TestflingerEvent,
+    TestPhase,
+)
+
+_MESSAGE_TEMPLATES = {
```
ajzobro, line 26: Since this is a repeat of all event keys, I feel it should be maintained in the same place as all event keys are defined: enums.py

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+    JobEvent.JOB_SUBMITTED: "Job submitted by user {client_id} for queue {queue_name}.",  # noqa: E501
```
ajzobro: I don't know if this is better:
```suggestion
    JobEvent.JOB_SUBMITTED: "Job submitted by user {client_id} into queue {queue_name}.",  # noqa: E501
```

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+    JobEvent.JOB_ASSIGNED: "Job assigned to agent {agent_name}.",
+    JobEvent.JOB_STARTED: "Job started",
```
ajzobro: Job started ... means?  means that the agent is working on it?


```suggestion
    JobEvent.JOB_STARTED: "Agent {agent} started executing this job.",
```

This feels like a state that ought to start about 0.01 seconds after the JOB_ASSIGNED event.  (maybe 10 seconds) -- but anyway it feels like it's not a sufficient distinction as to warrant it's own event.

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+    JobEvent.JOB_PHASE_STARTED: "Phase {phase} started.",
```
ajzobro, line 29: I like this one, but I feel like the two above it will happen if not at exactly the same time, then very very close to the same time.

E.g. if the setup phase is registered as an event (I think it should be) then the JOB_PHASE_STARTED for the `setup` phase sounds to me like it would mean almost exactly the same thing as the JOB_STARTED event (making the JOB_STARTED event even more superflous)

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+    JobEvent.JOB_PHASE_COMPLETED: "Phase {phase} completed with exit code {status}",  # noqa: E501
+    JobEvent.JOB_COMPLETED: "Job completed.",
```
ajzobro, line 31: Could we include the overall status? It may not be worth it, need to see how these are used.
```suggestion
    JobEvent.JOB_COMPLETED: "Job completed with {overall_status}.",
```

server/src/testflinger/database.py
```diff
@@ -901,3 +907,21 @@ def update_agent_provision_log(agent_name, json_data) -> bool:
         mongo.db.agents.update_one({"name": agent_name}, {"$set": agent})
         found = True
     return found
+
+
+def add_job_event(job_id: str, event: dict) -> None:
```
ajzobro, line 1063: Missing complimentary `database.py` code for fetching this data.

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+    :param context: Additional context for formatting the message.
+    :return: Formatted event message based on template.
+    """
+    template = _MESSAGE_TEMPLATES[event_type]
+    return template.format_map(_DefaultContext(context))
+
+
+def build_event(
+    event_type: TestflingerEvent,
+    timestamp: datetime,
```
ajzobro:
```suggestion
    timestamp: Optional[datetime],
```

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+    :param event_type: The type of the event.
+    :param timestamp: The timestamp of the event.
+    :param detail: Additional details for the event.
+    :param context: Additional context for formatting the message.
+    :return: Dictionary representing the event.
+    """
+    message = format_event(event_type, **context)
+    return {
+        "event_name": event_type,
+        "timestamp": timestamp,
```
ajzobro:
```suggestion
        "timestamp": timestamp or datetime.now(timezone.utc),
```

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+    which events haven't been logged yet.
+
+    :param previous_data: The previous results data.
+    :param new_data: The new results data.
+    :return: List of detected new events.
+    """
+    timestamp = datetime.now(timezone.utc)
+    return [
+        *_build_phase_completed_events(previous_data, new_data, timestamp),
+        *_build_job_lifecycle_events(previous_data, new_data, timestamp),
```
ajzobro:
```suggestion
    """
    return [
        *_build_phase_completed_events(previous_data, new_data),
        *_build_job_lifecycle_events(previous_data, new_data),
```

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+    ]
+
+
+def _build_phase_completed_events(
+    previous_data: dict, new_data: dict, timestamp: datetime
```
ajzobro:
```suggestion
    previous_data: dict, new_data: dict
```

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+
+    :param previous_data: The previous results data.
+    :param new_data: The new results data.
+    :param timestamp: The timestamp of the event.
+    :return: new phase completed events or empty list if no new events.
+    """
+    return [
+        build_event(
+            event_type=JobEvent.JOB_PHASE_COMPLETED,
+            timestamp=timestamp,
```
ajzobro:
```suggestion
```

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+            phase=phase,
+            status=status,
+        )
+        for phase, status in new_data.get("status", {}).items()
+        if phase not in previous_data.get("status", {})
+    ]
+
+
+def _build_job_lifecycle_events(
+    previous_data: dict, new_data: dict, timestamp: datetime
```
ajzobro:
```suggestion
    previous_data: dict, new_data: dict
```

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+
+    previous_job_state = previous_data.get("job_state")
+    # Terminal job state: log a completion event once
+    if new_job_state in {"completed", "complete"}:
+        if previous_job_state in {"completed", "complete"}:
+            return []
+        return [
+            build_event(
+                event_type=JobEvent.JOB_COMPLETED,
+                timestamp=timestamp,
```
ajzobro:
```suggestion
```

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+    """Get a list of job lifecycle events.
+
+    This compares the previous and new results data to determine
+    which job lifecycle events have changed so we can later log an event.
+
+    A job lifecycle is defined as a change in a job's state.
+
+    :param previous_data: The previous results data.
+    :param new_data: The new results data.
+    :param timestamp: The timestamp of the event.
```
ajzobro:
```suggestion
```

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+) -> list[dict]:
+    """Get a list of phases events that are already completed.
+
+    This compares the previous and new results data to determine
+    which phases have a new `status` key reported so we can later
+    log an event.
+
+    :param previous_data: The previous results data.
+    :param new_data: The new results data.
+    :param timestamp: The timestamp of the event.
```
ajzobro:
```suggestion
```

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+    message = format_event(event_type, **context)
+    return {
+        "event_name": event_type,
+        "timestamp": timestamp,
+        "message": message,
+        "detail": detail,
+    }
+
+
+def detect_new_result_events(
```
ajzobro, line 79: The fact that this is called "*_new_*" has led me to remove the timestamps from all private methods to allow for the default to be used throughout.

This may result in minor differences between events that are added through these methods, as the resolution of the timestamp which is defaulted may be visible to the end-user, but it is no more or less accurate of a stance.

If it is truly desirable to log the SAME timestamp for all events detected by this "now" function, then ignore my suggestions related to the removal of the timestamp argument.

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+        if previous_job_state in {"completed", "complete"}:
+            return []
+        return [
+            build_event(
+                event_type=JobEvent.JOB_COMPLETED,
+                timestamp=timestamp,
+            )
+        ]
+
+    # Early return if the job state has not changed
```
ajzobro:
```suggestion
    # Early return if the job state has not changed or is not a known test phase
```

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+    :return: new job lifecycle events or empty list if no new events.
+    """
+    # Early return if there is no job_state reported yet
+    new_job_state = new_data.get("job_state")
+    if not new_job_state:
+        return []
+
+    previous_job_state = previous_data.get("job_state")
+    # Terminal job state: log a completion event once
+    if new_job_state in {"completed", "complete"}:
```
ajzobro, line 142: some day we need to resolve this "ed" vs "e" thing and should probably begin by introducing a "TERMINAL_PHASES" list.

Also, I want to revisit 'cancelled' as a terminal phase. "completed" nor "complete" are in `enums.TestPhase` any more than `cancelled` is -- all of them are in effect "terminal" in-so-far as they are not active phases.

server/src/testflinger/events.py
```diff
@@ -0,0 +1,184 @@
...
+    # Special handler for the first phase of a job
+    if previous_job_state in (None, JobState.WAITING):
+        new_events.append(
+            build_event(
+                event_type=JobEvent.JOB_STARTED,
+                timestamp=timestamp,
+                phase=new_job_state,
+            )
+        )
+    # Any other phase change should log a phase started event
```
ajzobro:
```suggestion
    # All phase changes should log a phase started event
```

ajzobro (changes requested):
First look, didn't focus on the tests, just the code for now.

With respect to this effort versus #1293 I think that they are both different, but could both be used to solve certain problems.

What I need to look closer at is if the "events" solution you present here assigns a timestamp for *every* job state that can exist, even if the agent rapidly progresses through several at once.  The current solution over in #1293 would not be able to log a timestamp except for a state that was published explicitly.

Edit: yes, it appears that all intermediate phases are accounted for, on line 122 of events.py https://github.com/canonical/testflinger/pull/1297/changes#diff-e5803e8d3dd52ad1f9fa87b23cb5cc6d00444764d522814b68cc9aa2ca197f4fR122

server/src/testflinger/events.py
ajzobro: But also, please add a comment as to why we kept the singular timestamp if that's the selection.

server/src/testflinger/api/v1.py
rene-oromtz: You are right, there is no need for other than default ("now"). Given the events are stored based on when they where recorded server-side, there is no need for the optional argument. I'll modify and simplify with a default timestamp

server/src/testflinger/database.py
rene-oromtz: I can add it, I think its safe given its not a TTL index. Only problem would be is if we had any duplicates but this is highly unlikely

server/src/testflinger/database.py
rene-oromtz: Yes, I think you already also had a quick look at #1298 where the fetch is being done. As this PR is just for adding it to the database, I split it in add (this PR) vs get(the followup PR)

server/src/testflinger/database.py
rene-oromtz: As per your confirmation, this is done by the event handler. This timestamp is for updating when the document was last modified.

server/src/testflinger/events.py
rene-oromtz: I don't feel this belongs in enum.py though (not really an enum), perhaps also doesn't belong here.

Given its only used by server I can perhaps move it to: `server/src/testflinger/event_templates.py` or alternatively to `common`  but this feels only if we were to use that template for other components (currently only `agent`)

Let me know your thoughts on the above

server/src/testflinger/events.py
rene-oromtz: Yes, the problem here is context, the `JOB_ASSIGNED` has the agent context because its triggered by `job_get` which contains the cookie. `result_post` doesn't have this in its context and given the job_assigned already contains this data I think its already clear on which agent is processing the job. 

Alternatively, I could remove JOB_STARTED if preferred

server/src/testflinger/events.py
rene-oromtz: No, you are right, there is no need for letting the event had its own defined timestamp as an argument. Given the timestamp is recorded server side on when the event happened, we should record this as default. Thanks for the feedback

server/src/testflinger/events.py
rene-oromtz: Yes, currently both setup and job started are logged at the same time. I though on JOB_STARTED as a "friendlier" way to indicate a  job start but its really just for that. I prefer to keep it for that reason but also I can remove if you prefer as it doesn't add much value. What do you prefer?

server/src/testflinger/events.py
ajzobro: 1) Scoping: if not used outside of server, then I would put it all into events.py rather than spreading it out.

2) The reason I am in favor of keeping the things together is that there are no guards against adding an enum and ensuring that it get's a new format string. Proximity in the code would help to ensure proper updates are made in future edits.

server/src/testflinger/events.py
rene-oromtz: To be fair, I think this is agent's fault for not using the enum. Perhaps you are right on the TERMINAL_PHASES  (using `completed` and `cancelled`) but we can easily remove the need of a "complete"

Given its a one line change in agent side, let me open a quick spinoff PR. 

As for the use here, it should then only use `JobState.COMPLETED` (cancelled is handled elsewhere)

server/src/testflinger/events.py
ajzobro: Ah, but each job has context now of which agent is working on it, so if we have job_id in context, we can also determine the agent's id.

It is `JOB_STARTED` which I am suggesting isn't needed, see later down where we also begin a phase within the job at the exact same time as we declare the `JOB_STARTED` event.

You are right (next comment, below), that it is a "friendlier" handle than having to know that "setup" is the first phase and means the same thing.

I do think that `JOB_ASSIGNED` and `JOB_STARTED` are effectively the same thing, so if we have `JOB_ASSIGNED` and we have "setup" then I don't know why we need `JOB_STARTED`.

`JOB_STARTED` only seems to differentiate between a job given and a job beginning "setup" -- humor me: what happens in between `JOB_ASSIGNED` and `JOB_PHASE_STARTED`:`setup`?
 - If the answer is "nothing" then: we don't need `JOB_STARTED`
 - If the answer is something else, why isn't that part of the job's "setup" phase?

server/src/testflinger/events.py
ajzobro: see above, resolving this comment thread in favor of the one above.

server/src/testflinger/events.py
rene-oromtz: The problem is context, there is currently no indicator on the overall job status and perhaps worth a separate discussion on how do we want this to be exposed

server/src/testflinger/events.py
rene-oromtz: If this is scoped to server and being kept here, I think a really simple way is to add a quick pytest using all enums values as a list just to make sure we are not skipping any in case we add more to the enum. 

I think this can solve both scoping and making sure every enum is included, what do you think?

server/src/testflinger/events.py
rene-oromtz: I'll remove JOB_STARTED as I agree it add no real value rather that being "friendly". The ASSIGNED and SETUP phase should be enough for indicating the job started and which agent is processing it

server/src/testflinger/events.py
rene-oromtz: For now I added a really simple unit test in [e4f8539](https://github.com/canonical/testflinger/pull/1297/commits/e4f853960c438bd9cf336290c45dfa1fdebf8110) that lives on server side to check all values in JobEvent are properly used. 

This will help in case we ever add something on the "enums" to make sure server side we have a template for that event. Let me know what you think

server/src/testflinger/events.py
rene-oromtz: As discussed, leaving as-is for now. We can check on the overall "complete" vs "completed" changes once all incoming changes are ready and we could play more with the enums

server/src/testflinger/events.py
rene-oromtz: As discussed, leaving as is for now. Given the results already includes information relevant to the overall status we are leaving simple this event

server/src/testflinger/database.py
ajzobro: AI helper says: "High: Existing deployments will fail during startup when creating the changed jobs.job_id index.  
server/src/testflinger/database.py:142 changes the pre-existing job_id_1 index from non-unique to unique using create_index("job_id", unique=True). MongoDB does not alter an existing index's options through create_index; it returns IndexOptionsConflict because the index name/key already exists with different uniqueness. Since create_indexes() runs during application setup, upgrades against any existing database will fail to start. Drop/rebuild the index through an explicit migration, or leave the existing index non-unique until such migration is in place."

ajzobro: I think we really need to consider how to do proper database migrations during server upgrades.

server/src/testflinger/api/v1.py
```diff
@@ -587,7 +604,15 @@ def result_post(job_id: str, json_data: dict) -> str:
     if content_length and content_length >= 16 * 1024 * 1024:
         abort(HTTPStatus.REQUEST_ENTITY_TOO_LARGE, message="Payload too large")
 
+    # We need to get new events before json_data is mutated by add_job_results
+    previous_data = (database.get_job_results(job_id) or {}).get(
```
ajzobro: None of this proves the job actually exists before posting the event.

server/src/testflinger/api/v1.py
```diff
@@ -587,7 +604,15 @@ def result_post(job_id: str, json_data: dict) -> str:
...
+        "result_data", {}
+    )
+    new_events = events.detect_new_result_events(previous_data, json_data)
+
     database.add_job_results(job_id, json_data)
+    for event in new_events:
+        database.add_job_event(job_id=job_id, event=event)
```
ajzobro, line 613: AI helper says:
" Medium: Concurrent result posts can emit duplicate lifecycle and phase-completion events.  
server/src/testflinger/api/v1.py:610-618 reads the old result document, decides which events are new, writes the result, then appends events in separate operations. Two simultaneous retries can both read the same previous state before either write occurs, and each will append the same event. The sequential retry test does not cover this race. Make event creation idempotent atomically, such as by recording a unique event key per job/phase/lifecycle transition or using a transaction/conditional update."

server/src/testflinger/api/v1.py
ajzobro: I recognize this is not a new problem introduced by this PR, but we should check the job exists before writing job-related data.

ajzobro (changes requested):

server/src/testflinger/database.py
```diff
@@ -1032,3 +1038,21 @@ def update_agent_provision_log(agent_name, json_data) -> bool:
...
+
+    Events are appended in natural insertion order (oldest first).
+
+    :param job_id: The ID of the job.
+    :param event: The event data to add.
+    """
+    mongo.db.jobs_events.update_one(
+        {"job_id": job_id},
+        {
+            "$push": {"events": event},
```
ajzobro, line 1074: Because we are using `push` consecutive retries could result in the same event being posted more than once. Do we want to only push if the last known event is a different event than the currently detected?

May not matter in the case of receiving a batch of states, as if we post 1,2 retry, post 1,2 checking that 1 != 2 still wouldn't solve the retry problem.

Thoughts?

server/tests/test_events.py
```diff
@@ -0,0 +1,125 @@
...
+from testflinger.events import _MESSAGE_TEMPLATES, detect_new_result_events
+
+
+def _event_names(events: list[dict]) -> list[str]:
+    """Extract event_name values from a list of event dicts."""
+    return [event["event_name"] for event in events]
+
+
+def test_subsequent_phase_transition_emits_only_phase_started():
+    """Test a phase-to-phase transition only emits JOB_PHASE_STARTED."""
```
ajzobro: Interesting... why are we not producing a phase complete for each phase that is started?
If we only sometimes log when a phase is completed then we don't have a consistent way to digest that phase transition data later, e.g. in the case that we wanted to measure and report phase durations -- we will have to basically make an algorithm that works in the absence of phase_complete events and thus might as well not ever have them.,

Did I miss something?

server/tests/test_events.py
```diff
@@ -0,0 +1,125 @@
...
+    assert all(
+        event["event_name"] == JobEvent.JOB_PHASE_COMPLETED for event in events
+    )
+    messages = [event["message"] for event in events]
+    assert any(JobState.SETUP.value in message for message in messages)
+    assert any(JobState.PROVISION.value in message for message in messages)
+
+
+@pytest.mark.parametrize("terminal_state", (JobState.COMPLETED, "completed"))
+def test_terminal_job_state_completed_emits_job_completed(terminal_state):
```
ajzobro: I still view "cancelled" as a terminal state.

server/tests/test_results.py
```diff
@@ -321,13 +326,22 @@ def test_result_status_exit_codes(mongo_app, agent_auth_header, exit_code):
     assert response.status_code == HTTPStatus.OK
     assert response.json.get("test_status") == exit_code
 
+    # The JOB_PHASE_COMPLETED event message should surface the exit code.
```
ajzobro, line 332: If this is because the job phase in question is the test phase, please update the comment.

If other phases are also supposed to return an exit code, we should test other phases as well as `test`

server/src/testflinger/events.py
```diff
@@ -0,0 +1,164 @@
...
+            return []
+        return [
+            build_event(
+                event_type=JobEvent.JOB_COMPLETED,
+            )
+        ]
+
+    # Early return if the job state has not changed or is not a known TestPhase
+    if (
+        new_job_state not in {phase.value for phase in TestPhase}
```
ajzobro, line 153: Won't the following also work?
```suggestion
        new_job_state not in TestPhase
```

server/src/testflinger/events.py
```diff
@@ -0,0 +1,164 @@
...
+    :param new_data: The new results data.
+    :return: List of detected new events.
+    """
+    return [
+        *_build_phase_completed_events(previous_data, new_data),
+        *_build_job_lifecycle_events(previous_data, new_data),
+    ]
+
+
+def _build_phase_completed_events(
```
ajzobro, line 97: wouldn't it be possible for the list of events to contain multiple job phase starts and only one job phase exit, even if that same phase were seen to start again?
Thought · 4ms
Yes. The current model allows that mismatch.

ajzobro (changes requested):

server/src/testflinger/api/v1.py
rene-oromtz: Good catch! I can do that as we are already here and its easy to add

server/src/testflinger/api/v1.py
rene-oromtz: Yeah... I think the risk may be low but I can see that the read/write being separate can lead to duplicates, let me check on the best approach and I'll come back with a proposal on next push

server/src/testflinger/database.py
rene-oromtz: same comment as above, I think I can come up with a solution that guarantees the event to be posted only once. I'll include on next push

server/src/testflinger/database.py
rene-oromtz: Yeah... I sadly noticed this before the demo as changing `unique=true` broke my local server so I had to revert the change :( for now I removed it given its unlikely we get the same UUID (not impossible) but we definitively need to revisit this later so we can be able to modify the indices whenever needed in a safe way

server/tests/test_events.py
rene-oromtz: The phase complete event is only present f there is a status transmitted for that phase. In this test we are only transmitting "job_state" without any result so the phase completed is not triggered.

server/tests/test_events.py
rene-oromtz: I can rename this to "test_completed" for clarity. I do agree that "cancelled" is a terminal state but the "complete" event is not triggered by it due to the current design given the agent will still continue to process the job and eventually, send a "complete" status. This is one of the things we discussed as the limitations on the agent still "processing" the job even when it was requested a cancellation. For now, cancellation is handled separately

server/tests/test_results.py
rene-oromtz: I can extend it to other phases, all sends their results

server/src/testflinger/events.py
rene-oromtz: Unfortunately, not in Python 3.10... This breaks as follows:
`TypeError: unsupported operand type(s) for 'in': 'str' and 'EnumMeta'`

rene-oromtz: Only available until Python 3.12
https://docs.python.org/3/library/enum.html#enum.EnumType.__contains__

> Changed in version 3.12: Before Python 3.12, a TypeError is raised if a non-Enum-member is used in a containment check.

server/tests/test_events.py
rene-oromtz: I updated the docstrings and the test name to make it clearer that this is only for testing "plain" job_state changes without results

server/tests/test_events.py
rene-oromtz: Renamed to indicate this is for "complete/d" only

server/tests/test_results.py
rene-oromtz: Extended to include phase as a parametrized options

server/src/testflinger/api/v1.py
rene-oromtz: Done, aborting early in case the job does not exists

server/src/testflinger/api/v1.py
rene-oromtz: Okay, should be now addressed by both `ReturnDocument.BEFORE` and `find_one_and_update()` as this guarantees atomicity and a clean `previous_data` before actually making the update

server/src/testflinger/database.py
rene-oromtz: This should also be addressed by this [comment](https://github.com/canonical/testflinger/pull/1297#discussion_r3983633887) as only "new" events per request are made so the push should be safe now

server/src/testflinger/events.py
rene-oromtz: Yes but we don't allow for a phase to be started again as everything is sequential in the agent. Perhaps something for a future discussion if we were to allow a phase to be retried?

ajzobro (approved):

PR #1299, opened 2026-09-01
feat(server): add tabular visualization for job events in the UI

rene-oromtz:
## Description
This PR adds the job events into the job_detail view.

This:
1. Get all available events or a generic message in case none available
2. Adds all available events by the time of page load
3. Also add icons in case the `status` is available for an event, otherwise, use an information icon

I added the section as an "Activity" as I wanted to keep it generic and have this same section for the agent detail page which can contain events + agent logs in future PRs

## Resolved issues
Merging the full stack resolves [CERTTF-1448](https://warthogs.atlassian.net/browse/CERTTF-1448)

## Documentation
N/A

## Web service API changes
N/A

## Tests
Added unit tests for view pages. Additionally, tested locally:
Screenshot from 2026-09-01 15-32-03

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

[CERTTF-1448]: https://warthogs.atlassian.net/browse/CERTTF-1448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/tests/test_views.py
+++ b/server/tests/test_views.py
@@ -325,6 +325,7 @@ def test_job_detail_shows_activity_events(testapp):
                     ),
                     "message": "Job started.",
                     "detail": "",
+                    "status": 0,
                 },
                 {
                     "event_name": "job_submitted",
@@ -346,6 +347,8 @@ def test_job_detail_shows_activity_events(testapp):
     assert "Job started." in html
     assert "Job submitted." in html
     assert "2026-01-01 12:00:00" in html
+    assert '<i class="p-icon--success" aria-label="Success"></i>' in html
+    assert '<i class="p-icon--information" aria-label="Info"></i>' in html
 
 
 def test_job_detail_no_events_shows_placeholder(testapp):
--- a/server/tests/test_views.py
+++ b/server/tests/test_views.py
@@ -302,6 +302,72 @@ def test_job_results_mongo_logs(testapp):
     assert "Exit Status:</span> 1" in html
 
 
+def test_job_detail_shows_activity_events(testapp):
+    """Test that job_detail renders stored events in the Activity section."""
+    mongo = mongomock.MongoClient()
+    job_id = str(uuid.uuid4())
+    mongo.db.jobs.insert_one(
+        {
+            "job_id": job_id,
+            "created_at": datetime.now(timezone.utc),
+            "job_data": {"job_queue": "queue1"},
+            "result_data": {"job_state": "complete"},
+        }
+    )
+    mongo.db.jobs_events.insert_one(
+        {
+            "job_id": job_id,
+            "events": [
+                {
+                    "event_name": "job_started",
+                    "timestamp": datetime(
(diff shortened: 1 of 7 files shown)


server/tests/test_views.py
```diff
@@ -302,6 +302,75 @@ def test_job_results_mongo_logs(testapp):
...
+    with patch("testflinger.database.mongo", mongo):
+        with testapp.test_request_context():
+            response = job_detail(job_id)
+
+    html = str(response)
+    assert "Activity" in html
+    assert "Job started." in html
+    assert "Job submitted." in html
+    assert "2026-01-01 12:00:00" in html
+    assert '<i class="p-icon--success" aria-label="Success"></i>' in html
```
ajzobro, line 377: Should we also verify a failure for completeness?

ajzobro (dismissed):

server/tests/test_views.py
rene-oromtz: updated for completeness: [6b559e1](https://github.com/canonical/testflinger/pull/1299/commits/6b559e149b789d09883f6a9d1b8f128b00c26563)

server/tests/test_views.py
```diff
@@ -302,6 +302,103 @@ def test_job_results_mongo_logs(testapp):
...
+def test_job_detail_shows_activity_events(testapp):
+    """Test that job_detail renders stored events in the Activity section."""
+    mongo = mongomock.MongoClient()
+    job_id = str(uuid.uuid4())
+    mongo.db.jobs.insert_one(
+        {
+            "job_id": job_id,
+            "created_at": datetime.now(timezone.utc),
+            "job_data": {"job_queue": "queue1"},
+            "result_data": {"job_state": "complete"},
```
ajzobro, line 314: I think `completed` is the proper status, no?

rene-oromtz: It's actually the other way :( the agent always sends "complete". In fact I think the enum COMPLETED is not even used other than tests and server

rene-oromtz: I had a branch that I'm holding off (due to incoming changes) because I wanted to use the enum COMPLETED all over the place but this was more "breaking" (cli + server + agent) than just renaming the enum as COMPLETE. But that is a discussion for later on how we want to deal with that inconsistency

ajzobro: The agent is wrong.  If you scan the code, it should confirm that the enum vale is "completed" and that the agent has a place where it sends the wrong value.  It's a quick fix to update them as we find them.

ajzobro (dismissed):

ajzobro (approved):
+1

ajzobro (approved):

PR #1304, opened 2026-09-04
fix(agent): allow agent to came back online after restart

rene-oromtz:
## Description
Minor bug fix, during restart, the status can't be set as offline otherwise, the offline will be preserved.

## Resolved issues
N/A

## Tests
Existing tests passed.


ajzobro (approved):
Great catch!

PR #1305, opened 2026-09-07
docs(server): generate the API role-permission table from OpenAPI spec

tang-mm:
## Description

This PR generates the API endpoint role table in `docs/reference/api-roles.rst` from the server's
OpenAPI spec to avoid drift in the manually maintained table. The roles are already annotated in each operation with `x-permission-roles`, sourced from the `@require_role` decorator. 

- Adds a new script `server/scripts/generate_api_roles.py` to generate the API role-permission matrix in a separate RST-formatted file (`docs/reference/api-roles-table.txt`). 
  - The generated matrix is imported to the main API roles page `api-roles.rst` via an RST directive  `.. include::` 
  - The output file uses `.txt` extension to avoid being directly processed as a new doc file by Sphinx.
  - The order preserves the original hand-curated grouping: rows cluster by role set, ordered agent-first then by human-privilege tier (a mirrored order to keep the visual hierarchy consistent),
- Updated `just` `tox` `pre-commit` and CI to include the new role-permission table generation and checks.

Notes: The table of no-authentication endpoints remains hand-maintained in the RST, because the list is small, stable, and mostly verified by the exceptions in its test case.

## Resolved issues

CERTTF-1292, follow-up of the discussions in https://github.com/canonical/testflinger/pull/1172#issuecomment-4899976184

## Documentation

- `docs/reference/api-roles.rst` now includes the generated matrix fragment.
- `CONTRIBUTING.md` documents the regenerate/check workflow and that the no-auth table is hand-maintained.

## Web service API changes

N/A

## Tests

- `just server::roles-docs` regenerates `docs/reference/api-roles-table.txt`.
- `just server::check-roles-docs` fails if the committed fragment is stale;
  this runs in `server-tox` CI and as a pre-commit hook.
- Verified `make html` builds the docs page (both tables render).


server/scripts/generate_api_roles.py
```diff
@@ -0,0 +1,239 @@
...
+        type=Path,
+        help="Compare generated matrix with the specified file for validation",
+    )
+
+    args = parser.parse_args()
+
+    if args.diff:
+        if not diff_roles_matrix(args.diff):
+            sys.exit(1)
+        print(" API roles matrix is up to date")
```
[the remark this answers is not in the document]
tang-mm: Won't change - The leading space is intentional to distinguish from the main command output.

server/scripts/generate_api_roles.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+def _role_group_key(method: str, path: str, roles: list) -> tuple:
+    """Sort key that clusters rows by their (validated) role set.
+
+    The order of role set is described by two dimensions:
+    - whether AGENT is allowed,
+    - whether the lowest (widest-audience) human role is allowed. 
+    
+    This groups identical role sets together and keeps a semantic ordering.
```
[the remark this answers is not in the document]
tang-mm: updated

docs/reference/api-roles.rst
```diff
@@ -33,6 +33,9 @@ that are permitted to call it.  The four roles are:
 ``ADMIN``
     Full access to all endpoints, including user/credential administration.
 
+Endpoints that require no authentication 
+------------------------------------------
```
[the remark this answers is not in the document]
tang-mm: updated

tang-mm:
Hi @ajzobro , it looks like the [zizmor validation check](https://github.com/canonical/testflinger/actions/runs/34110526805/job/101705439546?pr=1305) is failing on several existing files under `.github/workflows/` and `.github/actions/`. Could you take a quick look when you have a moment? Thanks!

ajzobro:
> Hi @ajzobro , it looks like the [zizmor validation check](https://github.com/canonical/testflinger/actions/runs/34110526805/job/101705439546?pr=1305) is failing on several existing files under `.github/workflows/` and `.github/actions/`. Could you take a quick look when you have a moment? Thanks!

Looks like these were resolved by @rene-oromtz in PRs #1306 and #1307

server/justfile
```diff
@@ -20,6 +20,17 @@ schema:
 check-schema:
     @uvx tox run -e check-schema
 
+[doc('Generate the API roles reference page.')]
+[group('project')]
+roles-docs:
```
ajzobro: Since this needs to be run after the schema is updated, should we add `check-schema` as a predecessor?

```suggestion
roles-docs: check-schema
```

ajzobro: And for that matter, should we add `schema` as the pre-hook?

For that matter, should `schema` include `roles-docs` as a post-execution-hook?

docs/reference/api-roles-table.txt
```diff
@@ -0,0 +1,241 @@
+.. Generated by `python scripts/generate_api_roles.py`; do not edit by hand.
```
rene-oromtz: Perhaps this is more than enough but alternatively, if we were to enforce this, can you please consider adding a check to either:

- [server-tox.yml](https://github.com/canonical/testflinger/blob/main/.github/workflows/server-tox.yml): include `docs/reference/api-roles-table.txt` in the path filter so manual changes are also flagged by CI

or
- [documentation_check.yml](https://github.com/canonical/testflinger/blob/main/.github/workflows/documentation_checks.yml): add a check for the API roles on every docs changes

Personally, I would go with the second option given this is more docs scoped but I can also see it on server workflow as its tightly coupled via the server recipe. I think with either option the failure will be more clearly visible if a future PR attempts to manually modify this file i.e. currently, any PR editing only `docs/` will not trigger the check as its only enforced on server changes

server/scripts/generate_api_roles.py
```diff
@@ -0,0 +1,239 @@
...
+
+def _validate_role_hierarchy(method: str, path: str, allowed: set) -> None:
+    """Fail if human-role access for an endpoint is not upward-closed.
+
+    Allowing a lower human role while denying a higher one contradicts the
+    role hierarchy (a MANAGER should have at least a CONTRIBUTOR's access),
+    so it almost certainly signals a mistaken `@require_role` decorator
+    rather than an intended permission set.
+    """
+    for lower, higher in zip(HUMAN_ROLE_HIERARCHY, HUMAN_ROLE_HIERARCHY[1:]):
```
rene-oromtz: nitpick:
```suggestion
    for lower, higher in zip(HUMAN_ROLE_HIERARCHY, HUMAN_ROLE_HIERARCHY[1:], strict=False):
```

Only reason why I suggest adding this now is because perhaps we later would want to enforce all linting/formatting rules to the `scripts/` directory and this will immediately be flagged for not using `strict=`. This will make explicit that the tuples are not of the same length and will save us to attemp to add `strict=True` by accident.

docs/reference/api-roles.rst
```diff
@@ -33,6 +33,9 @@ that are permitted to call it.  The four roles are:
 ``ADMIN``
     Full access to all endpoints, including user/credential administration.
 
+Endpoints that require no authentication 
+----------------------------------------
```
rene-oromtz: nitpick: trailing whitespace

server/scripts/generate_api_roles.py
```diff
@@ -0,0 +1,239 @@
...
+        if lower in allowed and higher not in allowed:
+            hierarchy = " < ".join(r.value for r in HUMAN_ROLE_HIERARCHY)
+            raise ValueError(
+                f"{method.upper()} {path} allows '{lower.value}' but not "
+                f"'{higher.value}'. Human-role access must be upward-closed "
+                f"({hierarchy}); a more-privileged role should never have "
+                "less access. Fix the @require_role decorator."
+            )
```
rene-oromtz, line 132: Given we are raising here, could you please add a try/except in `generate_roles_matrix` or `main()` to prevent the full traceback? I think the message is clear so we can benefit of getting rid of the traceback in case of any failure.

server/scripts/generate_api_roles.py
```diff
@@ -0,0 +1,239 @@
...
+def diff_roles_matrix(local_path: Path) -> bool:
+    """
+    Generate the API roles matrix from code and compare with the committed
+    include fragment.
+
+    Args:
+        local_path: Path to the expected RST include fragment
+
+    Returns:
+        True if the fragment is up to date, False otherwise
```
rene-oromtz: Can you consider using sphinx style docstrings instead of Google Style docstrings for consistency with the repo? Not a big deal because there is no enforce mechanism but using `:param <param>:`, `:returns:` and `:raises <exception>: ` (when applicable) should be more than enough if this could be a possibility. 

Also I noticed this is the only function that is documenting params/returns so perhaps worth to add it to the other functions as well for clarity and consistency if you have the chance.

docs/reference/api-roles-table.txt
rene-oromtz, line 1: Perhaps this is more than enough but alternatively, if we were to enforce this, the following are two options worth considering

- [server-tox.yml](https://github.com/canonical/testflinger/blob/main/.github/workflows/server-tox.yml): include `docs/reference/api-roles-table.txt` in the path filter so manual changes are also flagged by CI

or

- [documentation_check.yml](https://github.com/canonical/testflinger/blob/main/.github/workflows/documentation_checks.yml): add a check for the API roles on every docs changes

Personally, I would go with the second option given this is more docs scoped but I can also see it on server workflow as its tightly coupled via the server recipe. I think with either option the failure will be more clearly visible if a future PR attempts to manually modify this file i.e. currently, any PR editing only docs/ will not trigger the check as its only enforced on server changes

rene-oromtz (changes requested):
Thanks for taking the time to add these changes! This looks really good, I added minor suggestions/nitpicks but overall LGTM and happy to see this land. 

Not now, but given we now have two scripts, I think it will be worth to eventually add `scripts/` path to the server linter so we can also check format on both of these files but this can be done in a follow-up PR

docs/reference/api-roles-table.txt
tang-mm: Good point. I've only considered changes to the server code rather than doc edits. I opted for the second one because it's less costly to run a single check if the text file is an unwanted change ([d077624](https://github.com/canonical/testflinger/pull/1305/commits/d0776240fdaaaa18fb4c0385bf9e4b8e2e56a26c))

server/scripts/generate_api_roles.py
tang-mm: Done. I also updated the docstrings for the other methods. There is another usage of the Google style in the sibling `generate_openapi_schema` script, and it's also updated

server/scripts/generate_api_roles.py
tang-mm: Done. I added it to `main` so the exit is clearer

server/scripts/generate_api_roles.py
tang-mm: Thanks for pointing it out! I updated it in [fe330e2](https://github.com/canonical/testflinger/pull/1305/commits/fe330e2b7f3fcc179bcbb2c3c43f8d651237109d)

tang-mm:
Thanks for the review @rene-oromtz !  I've addressed the comments

> Not now, but given we now have two scripts, I think it will be worth to eventually add `scripts/` path to the server linter so we can also check format on both of these files but this can be done in a follow-up PR

Since I added more docstrings there, I went ahead and added the `scripts/` path to the `[tool.ruff]` include filter, so the scripts are linted/formatted in CI and pre-commit. Both pass cleanly now.

rene-oromtz (approved):
Thank you for addressing the changes and for even including the lint/format checks! 
This LGTM

PR #1308, opened 2026-09-08
fix(workshop): make LXD tunnel endpoints asymmetric so external access works

ajzobro:
## Summary

The LXD `docker-ce` **slot** endpoints and the LXD `system` **plug** endpoints
were both set to `localhost:PORT`. That looks symmetric, but the two sides of
an LXD tunnel proxy are semantically different:

- **Slot endpoint** = *connect* target dialled inside the container's netns.
  With `userland-proxy: false` (see PR #1309) dockerd binds the published ports
  on container loopback, so this side must be `127.0.0.1:PORT`.
- **Plug endpoint** = *listen* address on the host. Binding it to `localhost`
  silently makes the host bind loopback-only, so external clients (e.g. a
  neighbour hitting `<external iface>:PORT`) cannot reach it — the failure is
  silent because `localhost` accepts writes and just drops packets from
  non-loopback peers.

This PR:

- pins the slot endpoints to `127.0.0.1:PORT` (explicit, matching the connect
  semantics) and the plug endpoints to `0.0.0.0:PORT` (explicit listen-all);
- adds a header comment on `.workshop/dev.yaml` documenting the asymmetry
  so nobody "unifies" them in a follow-up refactor.

No behaviour change for local-only development; external access previously
failed silently and now works.

## Order / independence

Cut from `main`. Independent of every other split branch. Sharing
`.workshop/dev.yaml` with `chore/workshop-docker-gc-sdk` means one of
the two will need a trivial `git pull --rebase` after the other merges;
neither ordering blocks the other.

## Risk

Low. Workshop files and related documentation only. No production code path touched.


rene-oromtz (approved):
LGTM

rene-oromtz (approved):
+1

PR #1310, opened 2026-09-08
chore: repo-wide `clean` recipes + ruff config for common

ajzobro:
## Summary

Two independent housekeeping items grouped because both are
lint/dev-workflow chores and neither is worth its own PR:

**A. `clean` just recipes**

Every Python subproject gains a `clean:` recipe under the `venv` group that
removes `.tox` to force a full reinstall on the next `just check`. A root
aggregator `just clean` walks the seven subprojects and additionally
`uv cache clean`s the two path-installed local packages. Motivation:
`.tox` environments cache stale installs of the sibling `-common`/
`-device-connectors` packages when their source changes without a version
bump; a targeted `just clean` is faster than manually purging.

**B. Ruff config for `common/`**

Enables a documented set of ruff rules (A, B, C4, D, DTZ, E, F, G, I,
LOG, N, PLE, PLW, S, W) with pep257 docstring convention on
`common/`. This adds two `# noqa: LOG004` markers on existing
`owasp_logger.exception` calls in `server/src/testflinger/application.py`
so `common/`'s rule set doesn't spuriously break server CI, and a
one-line import cleanup in `common/tests/test_duration.py` that ruff
now insists on.

The `noqa` markers belong to this PR because **this PR is what enables
the rule that flags them**. The follow-up PRs (dev auto sign-in, etc.)
add only `.warning(...)` calls which LOG004 does not affect.

## Files

Whole-file takes (see `files.txt`):
- `common/justfile`
- `common/pyproject.toml`
- `common/tests/test_duration.py`
- `cli/justfile`
- `device-connectors/justfile`
- `agent/charms/testflinger-agent-host-charm/justfile`
- `server/charm/justfile`
- `justfile` (root)

Files also touched by other branches (disjoint regions; whichever
lands second needs a trivial rebase):
- `agent/justfile` — this branch adds only the `clean:` recipe. The
  fake-agents recipes on the kitchen-sink branch are held back until
  that feature is complete.
- `server/justfile` — this branch adds only the `clean:` recipe.
  `chore/server-expose-compose-override` adds `serve-exposed` in the
  same file.
- `server/src/testflinger/application.py` — this branch adds two
  `# noqa: LOG004` markers on pre-existing `owasp_logger.exception`
  lines (needed because this is the branch that turns on the LOG rule
  set). `feat/server-dev-auto-signin` also edits this file elsewhere
  (imports, context processor, blueprint registration).

## Order / independence

Cut from `main`. Independent of every other split branch. The two
`noqa: LOG004` markers exist because *this* branch is the one that
turns on the LOG rule set (in `common/pyproject.toml`); they don't
create a dependency on any other branch.

## Risk

Low. Chore-only. New CI rules apply only to `common/` and are opt-in via
its own `pyproject.toml`.


ajzobro:
Code coverage of the two `exception` lines should be acheived in our owasp logging test...
This PR only changed the comments at the end of the line to avoid warnings about https://docs.astral.sh/ruff/rules/log-exception-outside-except-handler/

In both cases we are producing a 500 Server Error, akin to an exception, with an exception `exc` passed along.

rene-oromtz (commented):
Due to the motivation and before proceeding with the review, can you please consider using [editable dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#editable-dependencies) as well?

This could be on top of the `clean` recipes (not necessarily a replacement) which could save time because editable dependencies does not require to rebuild the environment. With this perhaps the `clean` can remain for other potential use cases that could need a full cleanup rather than an actual need to cleanup the environment while making changes to `common`

The server is already doing this for `common` and this is why the server doesn't need a version bump to get the latest changes:
https://github.com/canonical/testflinger/blob/148934207a9e22b8ddbf67f0464642abba0c7f87/server/pyproject.toml#L97-L98

Agent is currently not using editable install so changes are not immediately picked up:
https://github.com/canonical/testflinger/blob/148934207a9e22b8ddbf67f0464642abba0c7f87/agent/pyproject.toml#L89-L91

Additionally, linting for `common` is failing with some (7) `D105`, `D101`, `D400` and `D200` errors. Given its potentially a one line fix for each, up to you if you decide to include the "fix" or add to the ignore list for now.

rene-oromtz (approved):
Thanks for adding the suggestions. LGTM

PR #1312, opened 2026-09-08
feat(server): agent mode/state API with v1 backward compatibility

ajzobro:
## Summary

Adds server-side support for the `mode` + `state` wire contract. Old
clients that send flat `state=` are still accepted; old clients that
read `state` on GET still see the same values they used to. The change
is safe to deploy ahead of any client update.

### On ingress (POST /v1/agents/data/<name>)

`_normalise_agent_mode_state` accepts either shape:

- **legacy flat** — `{"state": "offline"}` → derived as
  `mode=offline, state=None`; `{"state": "provision"}` →
  `mode=online, state=provision`; `state=maintenance` without an
  explicit `mode` is rejected (never a valid legacy value).
- **new mode+state** — `{"mode": "online", "state": "waiting"}` and
  friends; validated for consistency (offline/restart carry no
  sub-state; online/maintenance require one).

### On egress (GET /v1/agents/data)

`_synthesise_legacy_state` derives a `state` value from `mode` so v1
readers continue to work. This is additive — the new `mode` field is
also returned.

### Schema and OpenAPI

`AgentIn` gains an optional `mode` field. `AgentOut` gains `mode`,
`mode_changed_at`, `mode_changed_by`, `state_changed_at`,
`state_changed_by`. `openapi.json` is regenerated to match.

## Files

- `server/src/testflinger/api/schemas.py`
- `server/src/testflinger/api/v1.py`
- `server/schemas/openapi.json`
- `server/tests/test_v1.py`

## Order

Cut from `main`. Depends on `chore/common-agent-mode-enum` for
`AgentMode`.

## Deferred cleanup

The words "v1" and "v2" appear throughout the docstrings and comments
in `_normalise_agent_mode_state`, `_synthesise_legacy_state`,
`AgentIn`/`AgentOut`, and the `openapi.json` description. There is no
formal API versioning behind this; the terms leaked in during the
maintenance-mode refactor. A follow-up cleanup will rephrase these as
"legacy flat state" vs "mode+state" (or similar) without changing
behaviour. Tracked in `prs/04-breakdown/NOTES.md`.


rene-oromtz (commented):
Before reviewing, can you please merge the changes from main? (or `rebase` given the stack)

Also, I wanted to get your opinion on `v2`. I see this extensively being used without a "real" v2 path. That got me questioning, should we do a `/v2/agents/...`? Just thought on mentioning because when we did the logging endpoint it ended up complicating the deployment because how tight the integration was between `agent` <-> `server` <-> `cli` when that change perhaps deserved its own `v2` for logging. 

I'm thinking having a different version (or alternative a separate endpoint e.g. `/v1/agents/data/<agent_name>/status`) can help us let the agents run "old" code vs "new" with separation of paths (rather than handling it with helper functions.)

It really depends on the size of the effort of course but wanted to get your opinion before moving through the stack.

PR #1316, opened 2026-09-08
feat(server): dev auto sign-in blueprint (TF_DEV_AUTO_SIGNIN)

ajzobro:
## Summary

Used by PR #1317 

Adds a **development-only** `/auth/dev-signin` POST endpoint that bypasses
Dex and creates a session as one of four well-known dev identities (admin,
manager, two contributors). Gated by the `TF_DEV_AUTO_SIGNIN=1` env var
**and** OIDC being configured; disabled by default.

Why POST, not GET: this route mutates server state (session +
`client_permissions` row). GET makes it reachable by prefetchers, link
crawlers, `` shenanigans, or a stray browser history revisit,
and it also lets a reverse proxy mangle the `@` in the email
query parameter (which is how the first draft failed).

The base dev docker-compose sets `TF_DEV_AUTO_SIGNIN=1` so
`just server::serve` gives you the picker in the header dropdown
immediately. The compose-override for exposed testing
(`chore/server-expose-compose-override`) relies
on this being wired.

### Sample-data credential source of truth

`sample_users.py` now exports `TESTFLINGER_ADMIN_ID` /
`TESTFLINGER_ADMIN_SECRET`; `create_sample_users.py` and
`create_sample_data.py` both import them so there is one source of
truth for the seeded admin credential.

## Files

Whole-file takes (see `files.txt`):
- `server/src/testflinger/dev_signin.py` (new)
- `server/src/testflinger/oidc/views.py` (adds
  `session["user_email"] = …`)
- `server/src/testflinger/templates/_header.html` (dev-signin dropdown)
- `server/devel/sample_users.py`
- `server/devel/create_sample_users.py`
- `server/tests/test_dev_signin.py` (new)

Files also touched by other branches (disjoint regions; whichever
lands second needs a trivial rebase):
- `server/src/testflinger/application.py` — this branch adds the
  dev-signin imports, context-processor entries, and blueprint
  registration. `chore/repo-clean-recipes-and-ruff` adds two
  `# noqa: LOG004` markers elsewhere in the same file.
- `server/src/testflinger/static/assets/css/testflinger.css` — this
  branch adds the `.p-navigation__dropdown-item-form` reset for the
  dev-signin picker. `feat/agent-device-maintenance-mode-clean` adds
  the agent-state and job-state chip styles elsewhere in the same file.
- `server/devel/create_sample_data.py` — this branch adds the
  `sample_users` credential import and updates the client-id/secret
  defaults. `feat/agent-device-maintenance-mode-clean` adds the
  offline/maintenance seed elsewhere in the same file.
- `server/docker-compose.yml` — this branch sets
  `TF_DEV_AUTO_SIGNIN: "1"` so dev auto sign-in is active in the
  default dev compose.

## Order / independence

Cut from `main`. Independent of every other split branch. Shared files
above are all disjoint regions; merge in any order.

## Not in this branch

- Compose override for reverse-proxy exposed testing
  (`chore/server-expose-compose-override`). PR #1317

## Risk

Low in production (env var default is unset → blueprint never
registered). Medium in dev if `TF_DEV_AUTO_SIGNIN` accidentally leaks to
a shared environment: the log warns loudly on every request.


server/devel/create_sample_data.py
```diff
@@ -28,7 +28,11 @@
 from typing import Iterator, Optional, Tuple
 
 import requests
-from sample_users import SAMPLE_CLIENTS
+from sample_users import (
+    SAMPLE_CLIENTS,
+    TESTFLINGER_ADMIN_ID,
+    TESTFLINGER_ADMIN_SECRET,
+)
```
rene-oromtz: nitpick that you are free to ignore if prefer as-is but for consistency with create sample user (and given `TESTFLINGER_ADMIN_ID`/`TESTFLINGER_ADMIN_SECRET` are not used elsewhere) perhaps we could use:

```suggestion
from sample_users import (
    SAMPLE_CLIENTS,
    TESTFLINGER_ADMIN
)
```

Feel free to mark as resolved if you prefer the current usage.

server/devel/generate_sample_users.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+    lines.extend(["]", ""])
+    path = ROOT / "src/testflinger/dev_signin_identities.py"
+    path.write_text("\n".join(lines))
+
+
+def main() -> None:
+    """Generate all local account consumer inputs."""
+    users = _development_users()
+    _write_dex_config(users)
```
rene-oromtz: Same comment as above, the users is not actively used and we are genereting all identities from SAMPLE_CLIENTS. 


```suggestion
    _write_dex_config()
```

Alternatively:

```suggestion
    _write_dex_config(SAMPLE_CLIENTS)
```

server/devel/generate_sample_users.py
```diff
@@ -0,0 +1,127 @@
...
+    missing_roles = REQUIRED_ROLES - roles
+    if missing_roles:
+        missing = ", ".join(sorted(missing_roles))
+        raise ValueError(f"Missing development sign-in roles: {missing}")
+    return users
+
+
+def _write_dex_config(users: list[dict]) -> None:
+    """Write Dex configuration for all sample clients with an OIDC identity."""
+    dex_users = [user for user in SAMPLE_CLIENTS if "dex_user_id" in user]
```
rene-oromtz: using SAMPLE_CLIENTS directly instead of `users` param given this seems to create **all** identities and not only dev_auto_signing  I think we can drop the users:


```suggestion
def _write_dex_config() -> None:
    """Write Dex configuration for all sample clients with an OIDC identity."""
    dex_users = [user for user in SAMPLE_CLIENTS if "dex_user_id" in user]
```

Alternative if passing SAMPLE_CLIENTS to the function:

```suggestion
def _write_dex_config(users: list[dict]) -> None:
    """Write Dex configuration for all sample clients with an OIDC identity."""
    dex_users = [user for user in users if "dex_user_id" in user]
```

server/devel/sample_users.py
```diff
@@ -59,9 +75,18 @@
...
+]
+
+# Existing callers use these defaults to create sample data.
+TESTFLINGER_ADMIN = next(
+    client
+    for client in SAMPLE_CLIENTS
+    if client["client_id"] == "testflinger-admin"
+)
+TESTFLINGER_ADMIN_ID = TESTFLINGER_ADMIN["client_id"]
+TESTFLINGER_ADMIN_SECRET = TESTFLINGER_ADMIN["secret_key"]
```
rene-oromtz: nitpick related to create sample data [comment](https://github.com/canonical/testflinger/pull/1316/changes#r4049528569)

We could drop this and use TESTFLINGER_ADMIN instead but I don't mind if you keep as-is so feel free to mark this as resolved if preferred.

server/docker-compose.yml
```diff
@@ -29,6 +29,7 @@ services:
       OIDC_CLIENT_SECRET: ""
       OIDC_PROVIDER_ISSUER: http://dex:5556
       TEMPLATES_AUTO_RELOAD: "true"
+      TF_DEV_AUTO_SIGNIN: "1"
```
rene-oromtz: any particular reason on keeping "1" over a more explicit "true" or "enabled"?

PR #1321, opened 2026-09-10
feat(testflinger_agent): add recovery mechanism

fhost:
## Description

Add a configuration option (`recovery_enabled`) that makes it runs the `$REBOOT_SCRIPT` command on output timeout. When that command is run, it enters recovery mode for `recovery_timeout` (default: 5 min). If a new output is detected before the timeout, returns to normal execution, otherwise it exits.

This should allow to recover and mark as crashed a checkbox test that crashes/stucks a DUT.

## Documentation

I will do the documentation if this mechanism has a chance to be merged.

## Tests

I did not test it yet as I'm not sure how to be able to test it.

### DIFF
--- a/agent/src/testflinger_agent/job.py
+++ b/agent/src/testflinger_agent/job.py
@@ -146,6 +146,10 @@ class TestflingerJob:
             output_timeout_checker = OutputTimeoutChecker(
                 self.get_output_timeout()
             )
+            runner.output_timeout_checker = output_timeout_checker
+            runner.output_timeout = self.get_output_timeout()
+            runner.recovery_enabled = self.get_recovery_enabled()
+            runner.recovery_timeout = self.get_recovery_timeout()
             runner.register_stop_condition_checker(output_timeout_checker)
             runner.subscribe_event(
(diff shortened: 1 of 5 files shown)


agent/src/testflinger_agent/job.py
```diff
@@ -146,6 +146,10 @@ def run_test_phase(self, phase, rundir: Path):
             output_timeout_checker = OutputTimeoutChecker(
                 self.get_output_timeout()
             )
+            runner.output_timeout_checker = output_timeout_checker
+            runner.output_timeout = self.get_output_timeout()
+            runner.recovery_enaled = self.get_recovery_enabled()
```
ajzobro: spelling error. "enaled"

There's a lot that I don't understand here, mostly because this is a fix applied to the *agent* but it seems related to a specific type of *device*

We have some larger changes landing in the agent code soon, and it might be good to talk about what issues you were facing that this solution is expected to handle.

I admittedly didn't review this in depth because of workload, but let's talk!

ajzobro (changes requested):

PR #1326, opened 2026-09-17
Add s390x Support for the Testflinger-agent-host charm

mreed8855:
## Description
Add support for the s390x in the testflinger-agent-host charm

## Resolved issues
I have been using the testflinger-agent-host charm on s390x z lpars but have been building and deploying it locally, this should eliminate the need for that.


rene-oromtz (changes requested):
Hi Michael! 
This will also require to add a matrix in this workflows: 
https://github.com/canonical/testflinger/blob/main/.github/workflows/agent-host-charm-release.yml

The action does a ` charmcraft pack --quiet --destructive-mode` that builds the charm on the host arch so as of now, it will only release amd64

rene-oromtz:
@mreed8855 I just created #1333 in case you want to take some inspiration. That adds noble to the mix for amd64 but you could theoretically follow the same approach for s390x.

PR #1333, opened 2026-09-18
feat(agent-host-charm): add support for amd64 with 24.04 base

rene-oromtz:
## Description
This adds support for `noble` base charm while also keeping `jammy`. 

Proposal:
1. Add multiple platforms in charmcraft.yaml
2. Update integration tests to test each platform
3. Update release charm to release both charm bases.
4. This uses `pack-host` which on CI uses disruptive mode along a targeted platform. Bare `charmcraft pack` requires LXD as it builds both bases on AMD64. 

Non disruptive as releasing a new base does not require an immediate upgrade to the next base. Our current environments will remain pulling from its current base (`jammy`) until we are ready to upgrade to next base. 

## Resolved issues
First step towards multi-arch

## Tests
Tested releasing to edge: https://github.com/canonical/testflinger/actions/runs/35404272842

Noble:
Screenshot from 2026-09-18 17-07-26

Jammy:
Screenshot from 2026-09-18 17-14-43

Please note that charmcraft uses a revision per submission so that's why the revision differ.


pedro-avalos (dismissed):
LGTM

.github/workflows/agent-host-charm-release.yml
```diff
@@ -26,7 +26,24 @@ concurrency:
...
+      max-parallel: 1
+      fail-fast: false
+      matrix:
+        include:
+          - runner: ubuntu-22.04
+            arch: amd64
+            distro: jammy
+          - runner: ubuntu-24.04
+            arch: amd64
+            distro: noble
```
ajzobro:
```suggestion
          - runner: ubuntu-24.04
            arch: amd64
            distro: noble
          - runner: ubuntu-24.04
            arch: arm64
            distro: noble
          - runner: ubuntu-24.04
            arch: s390x
            distro: noble
```
Are we saving the additional arch's for another PR?
Are we going to effectively hard-code the arch per-charm so that they grab the right docker image (or is that already going to key off the running arch?)

.github/workflows/agent-host-charm-release.yml
```diff
@@ -51,6 +80,16 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - base: ubuntu@22.04
+            arch: amd64
+            distro: jammy
+          - base: ubuntu@24.04
+            arch: amd64
+            distro: noble
```
ajzobro, line 80: Same question as above re additional arch's

.github/workflows/agent-host-charm-release.yml
ajzobro: Actually, this `matrix` is defined, but we only seem to use `.runner`
Then we define another on line 85, but then appear to use all the components of that definition.
i.e. it seems we are not using `arch` and `distro` from this `matrix`

ajzobro (changes requested):

.github/workflows/agent-host-charm-release.yml
rene-oromtz: Yep, only runner is needed but having the other are useful as "labels" in the matrix.

Screenshot from 2026-09-21 10-36-24

Also I wasn't adding s390x as it can quickly be added in #1326 with Michael's change. I could do it in single PR but didn't wanted to close Michael PR

rene-oromtz:
@pedro-avalos @ajzobro releasing with non destructive mode is cleaner but a bit slower as it uses LXD. 
The release workflow worked as intended: https://github.com/canonical/testflinger/actions/runs/35626965443

Let me know what you guys think. Please note that for release, the "matrix" is just a single item but I added comments so @mreed8855 can add s390x in #1326

.github/workflows/agent-host-charm-release.yml
rene-oromtz: Also, regarding the docker image, no need to hardcode. The charm will pull the right docker image base on the running arch.

ajzobro: OK, thanks for the clarifications!

ajzobro (approved):
LGTM

pedro-avalos (approved):

